PR #310, opened 2014-12-18
Use native Ruby implementations in Yast::URL and Yast::IP

mvidner:
This is a continuation of #305, using a branch in the shared repo so that more people can contribute commits to the PR.


mvidner:
In this PR we have found that replacing the URL implementation is hard. If we don't have bugs/features to address then let's not continue in the refactoring.

PR #311, opened 2015-01-06
Add new PowerPC - PowerNV platform

k0da:
Threat PowerNV as CHRP for now. The major difference is PowerNV doesn't
have bootloader, firmware parses grub.cfg and doing kexec. But having
bootlaoder (PReP partition) is harmless.

### DIFF
--- a/library/general/src/modules/Arch.rb
+++ b/library/general/src/modules/Arch.rb
@@ -221,6 +221,11 @@ module Yast
         # Cell and Maple based boards have no CHRP in /proc/cpuinfo
         # Pegasos and Cell do have CHRP in /proc/cpuinfo, but Pegasos2 should no be handled as CHRP
         # Efika is handled like Pegasos for the time being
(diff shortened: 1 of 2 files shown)


library/general/src/modules/Arch.rb
```diff
@@ -221,6 +221,9 @@ def board_compatible
         # Cell and Maple based boards have no CHRP in /proc/cpuinfo
         # Pegasos and Cell do have CHRP in /proc/cpuinfo, but Pegasos2 should no be handled as CHRP
         # Efika is handled like Pegasos for the time being
+        # Treat PowerNV as CHRP. It is harmless for now. Patch for hwinfo is sent but it is better to be safe
+        if @_board_compatible == "PowerNV"
+          @_board_compatible = "CHRP"
```
jreidinger: here is missing `end` or elsif for next if. This way it do not pass travis build ( hard requirement to merge request ).

jreidinger:
few parts are not pure ruby, but it is consistent with rest. So please fix travis build and it will be OK

jreidinger:
LGTM

PR #312, opened 2015-01-07
keep kernel cmdline options during live installation (bsc#793065)

jsrain:


library/system/src/modules/Kernel.rb
```diff
@@ -248,6 +253,10 @@ def ParseInstallationKernelCmdline
       @cmdline_parsed = true
       return if !(Stage.initial || Stage.cont)
       tmp = Convert.to_string(SCR.Read(path(".etc.install_inf.Cmdline")))
+      # live installation does not create /etc/install.inf (bsc#793065)
+      if Mode.live_installation
+        tmp = Convert.to_string(SCR.Read(path(".target.string"), "/proc/cmdline"))
```
ancorgs: We have an specific ```.proc.cmdline``` agent. Shouldn't we use it instead of ```target.string```? I even have a forever-pending bug about it https://bugzilla.suse.com/show_bug.cgi?id=889497

On the other hand, I think that you should use if-else
```ruby
if Mode.live_installation
  # read from /proc
else
  # read from install.inf
end
```
And last but now least :-) I'm not sure if calling ```to_string``` is actually needed.

jsrain: I know about the agent. The reason not to use it is to process cmdline from install.inf the same as from /proc/cmdline during live install - to get a single string in both cases, parsing them via the same parser. Otherwise, one would be parsed via agent and another one via ruby code.
If-else - agreed, will adjust.

jreidinger:
@jsrain your explanation is good, but here will be forget. Please add reason why you do not use proc agent to code directly. So we know in future reason and maybe if we separate file from parser, then we can just have different file and same parser.

jreidinger:
LGTM

PR #313, opened 2015-01-07
Add new PowerPC - PowerNV platform

k0da:
Treat PowerNV as CHRP for now. The major difference is PowerNV doesn't
have bootloader, firmware parses grub.cfg and doing kexec. But having
bootlaoder (PReP partition) is harmless.


jreidinger:
LGTM

PR #314, opened 2015-01-08
Add event dispatcher module

jreidinger:


ancorgs:
I'd need some more example because I currently find the expected return value of ```xxx_handler``` to be confusing. It's always ignored unless it's a Hash with a :_exit_dialog key.

If we are going to ignore it in most cases, I'd not enforce such an arbitrary returning value. I'd rather have a method to set the return value of the dialog. If it's called from a handler, the dialog will close and return that value. The returning value of the handler would always be ignored.

So a method not causing the dialog to close would be.
```ruby
def add_handler
  @whatever += 1
  log.debug "Summed"
end
```

And a similar method causing the dialog to close would be
```ruby
def add_and_return_handler
  @whatever += 1
  exit_dialog(@whatever)
  log.debug("Summed and set as return value")
end
```

Maybe we could change the name of ```exit_dialog``` to empathise that it will not close immediately or we can actually ensure that it closes the dialog immediately stopping further execution. Anyway, I find calling a method less confusing that returning such a strange value.

jreidinger:
@ancorgs fair enough. I can add instance variable only for this purpose and set it. and then check it. any idea for better name then exit_dialog?

library/general/src/lib/ui/event_dispatcher.rb
```diff
@@ -0,0 +1,65 @@
...
+  #     end
+  #
+  #     def ok_handler
+  #       finish_dialog(:ok)
+  #       log.info "OK button pressed"
+  #     end
+  #   end
+  module EventDispatcher
+    # @internal constant to continue with dispatching
+    CONTINUE_WITH_DISPATCHING = :_dispatcher_continue
```
ancorgs: Why not using the symbol directly? I don't find much added value in defining a constant.

library/general/src/lib/ui/event_dispatcher.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    # Does UI event dispatching.
+    # @return value from exit_dialog method.
+    def event_loop
+      Yast.import "UI"
+      @_finish_dialog_dispatcher = CONTINUE_WITH_DISPATCHING
+
```
ancorgs: Wouldn't it be more clear to use two instance variables? One boolean to decide when to break (let's say ```@_return_from_loop``` or ```@_continue_loop```) and another one to store the value to return (let's say ```@_return_value```). Not only the code would be cleaner, it would also save us from defining esoteric reserved symbols like ```:_dispatcher_continue```.

library/general/src/lib/ui/event_dispatcher.rb
jreidinger: to indicate that it is special value, it is also used twice, so I want simple visual that it is special value.

library/general/src/lib/ui/event_dispatcher.rb
jreidinger: yep, also possible. good idea :+1:

jreidinger:
@ancorgs updated

ancorgs:
LGTM

PR #317, opened 2015-01-13
Use native Ruby implementations in Yast::IP (only)

mvidner:
This is a restriction of #310 to only the IP class (excluding URL which still has unresolved issues).


mvidner:
LGTM, but I've been looking at this for too long, somebody else should check.

library/types/src/modules/IP.rb
```diff
@@ -272,32 +241,23 @@ def CheckNetworkShared(network)
...
       generic_check = CheckNetworkShared(network)
       if generic_check != nil
-        return generic_check 
+        return generic_check
 
-        # 192.168.0.1, 0.8.55.999
-      elsif Check4(network)
-        return true 
+      # 192.168.0.0/20, 0.8.55/158
+      elsif network =~ Regexp.new("^[" + @ValidChars4 + "]+/[0-9]+$")
```
jreidinger: it always make me crazy why ValidChars are not constant, then also this regexp can be constant and many others, making code easier to read

jreidinger:
other then small idea to improve ValidChars it looks good for me

library/types/src/modules/IP.rb
```diff
@@ -151,34 +141,25 @@ def ValidNetwork
...
+      IPAddr.new(ip, Socket::AF_INET).to_s
     end
 
     # Converts IPv4 address from string to hex format
     # @param [String] ip IPv4 address as string in "ipv4" format
     # @return [String] representing IP in Hex
-    # @example IP::ToHex("192.168.1.1") -> "0xC0A80101"
-    # @example IP::ToHex("10.10.0.1") -> "0x0A0A0001"
+    # @example IP::ToHex("192.168.1.1") -> "C0A80101"
+    # @example IP::ToHex("10.10.0.1") -> "0A0A0001"
```
ancorgs: Does this mean that the behavior of the class actually changed?

mvidner: It hasn't, as you can see in the unchanged tests: https://github.com/yast/yast-yast2/blob/573553179d769ac5dcc08d37c05baeabfbc921c9/library/types/test/ip_test.rb#L146
Just the docs was wrong.

library/types/src/modules/IP.rb
mvidner: Code in other modules uses the variables, so we must keep that API. I will leave it as it is now.

jreidinger: but only read-only if I remember correctly, so we can make accessor to constant. It is confusing to have editable ValidChars

PR #321, opened 2015-01-20
Robust tests for SLE12 branch

ancorgs:
Let's start fixing it in the SLE12 branch:

 * First, sync the tests with the ones in master (i.e. adding new tests for agents)
 * Then add the needed fixes. I didn't wanted to go too far in refactoring, just the minimum subset of fixes that makes SLE12 and master branch (and 13.2 I assume) work again.

Once merged, I'll create PR to merge the fixes to the other branches.

### DIFF
--- a/library/general/test/test_helper.rb
+++ b/library/general/test/test_helper.rb
@@ -3,4 +3,4 @@ inc_dirs = Dir.glob("#{top_srcdir}/library/*/src")
 ENV["Y2DIR"] = inc_dirs.join(":")
 
 require "yast"
-require_relative "SCRStub"
+require_relative "scr_stub"
--- a/library/general/test/test_helper.rb
+++ b/library/general/test/test_helper.rb
@@ -4,7 +4,3 @@ ENV["Y2DIR"] = inc_dirs.join(":")
 
 require "yast"
 require_relative "SCRStub"
-
-RSpec.configure do |c|
-  c.include SCRStub
-end
--- a/library/general/test/test_helper.rb
+++ b/library/general/test/test_helper.rb
@@ -1,4 +1,13 @@
-
 top_srcdir = File.expand_path("../../../..", __FILE__)
 inc_dirs = Dir.glob("#{top_srcdir}/library/*/src")
 ENV["Y2DIR"] = inc_dirs.join(":")
+
+def set_root_path(directory)
+  check_version = false
+  handle = Yast::WFM.SCROpen("chroot=#{directory}:scr", check_version)
+  Yast::WFM.SCRSetDefault(handle)
+end
+
+def reset_root_path
+  Yast::WFM.SCRClose(Yast::WFM.SCRGetDefault)
+end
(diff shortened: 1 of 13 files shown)


library/general/test/test_helper.rb
```diff
@@ -1,4 +1,10 @@
...
 top_srcdir = File.expand_path("../../../..", __FILE__)
 inc_dirs = Dir.glob("#{top_srcdir}/library/*/src")
 ENV["Y2DIR"] = inc_dirs.join(":")
+
+require "yast"
+require_relative "SCRStub"
+
+RSpec.configure do |c|
+  c.include SCRStub
+end
```
jreidinger: I think this part should be in `SCRStub` so when you require it, it is automatically assigned to RSpec. It will tight SCRStub to rspec, but it on other hand allow one place where we maintain connection and we do not need to fix all test_helper when rspec changes.

library/runlevel/test/service_test.rb
```diff
@@ -35,7 +35,7 @@ def stub_service_with(method, result)
       end
 
       it "returns the result of the original result of the command call" do
-        expect(Service.call("status", "sshd")).to be_kind_of(String)
+        expect(Service.call('status', 'sshd')).to be_kind_of(::String)
```
jreidinger: why this? our style guide say something different :)

jreidinger:
travis failed, probably due to rubocop, please check

library/system/test/kernel_test.rb
```diff
@@ -1,24 +1,21 @@
...
 Yast.import "Kernel"
 Yast.import "FileUtils"
-Yast.import "SCR"
-
-DEFAULT_DATA_DIR = File.join(File.expand_path(File.dirname(__FILE__)), "data/modules.d")
 
 describe "Kernel" do
-  before do
+  let(:stubbed_modules_dir) {
+    File.join(File.expand_path(File.dirname(__FILE__)), "data", "modules.d")
```
jreidinger: expand_path is useless here dirname of __FILE__ is already absolute.

library/runlevel/test/service_test.rb
ancorgs: Two days following the openQA's style guide for Perl confused me, I guess.

jreidinger: thats why we have mighty rubocop :)

jreidinger:
looks really good, just please improve a bit documentation and then LGTM and maybe SCRStub can be part of ruby bindings. Just please follow ruby convertion for naming files, so better to name it `scr_stub.rb`

ancorgs:
Updated, although I'm not 100% that scr_stub fits the convention, since there is actually not camelcasing between "SCR" and "Stub" in "SCRStub" (if you follow me). Other than that, I agree that it's a better file name, I was polluted by openQA and Perl again :-)

jreidinger:
@ancorgs LGTM . As long as we do not use rails auto require feature we are on safe side and I usually split it this way, so it is definitively better then before ;)

PR #322, opened 2015-01-20
add base class for dialogs

jreidinger:


library/general/src/lib/ui/dialog.rb
```diff
@@ -0,0 +1,84 @@
+require "yast"
+require "ui/event_dispatcher"
+
+module UI
+  # Base class for dialogs in Yast. Include usefull modules and provides
```
ancorgs: "useful". We all hate English sometimes. :-)

library/general/test/dialog_test.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      expect { subject.run }.to raise_error
+    end
+
+    it "ensure dialog is closed even if exception is raised in event loop" do
+      mock_ui_events(:ok)
+      expect(Yast::UI).to receive(:CloseDialog)
+
+      begin
+        subject.run
```
ancorgs: I don't see why subject.run is supposed to raise an exception here.

jreidinger: yep, exactly. Do you read my mind? :+1:

ancorgs:
LGTM, I will use it this week :smile:

jreidinger:
great, let me know when you do it, it will be interesting to see it in action. Maybe I get idea how to improve it even more.
One think that cannot be done is generic value obtainer. I ask @mvidner if libyui support introspection of widget and it is not possible, so it is not possible to have generic query for widget value unless I track content somewhere else, but it will be fragile due to ReplaceWidget call.

PR #325, opened 2015-01-22
Remove locked proposals

jreidinger:


library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -758,21 +758,18 @@ def MediaChange(error_code, error, url, product, current, current_label, wanted,
...
       )
 
-      url_tokens = URL.Parse(url)
-      url_scheme = Ops.get_string(url_tokens, "scheme", "")
-      url_scheme = Builtins.tolower(url_scheme)
+      url_scheme = Ops.get_string(URL.Parse(url), "scheme", "").downcase
 
       # true if it makes sense to offer an eject button (for cd/dvd only ...)
-      offer_eject_button = url_scheme == "cd" || url_scheme == "dvd"
+      is_disc = ["cd", "dvd"].include?(url_scheme)
```
kobliha: In fact, this means: is_removable[_media] instead of is_disc

jreidinger: Funny part is that below everything that is not this is called medium and this is called disc

jreidinger: see https://github.com/yast/yast-yast2/pull/325/files#diff-a9862329d78632ec55051c5a35eed347R799

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -1047,7 +1027,7 @@ def MediaChange(error_code, error, url, product, current, current_label, wanted,
               r = :retry
 
               # decrease attempt counter
```
kobliha: Oops, increase

jreidinger: yep, inverse logic, reach limit instead of descresing remaining attempts

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -1110,7 +1086,10 @@ def MediaChange(error_code, error, url, product, current, current_label, wanted,
         @_provide_popup = false
       end
 
-      if r == :cancel
+      case r
+      when :ignore then "I"
```
kobliha: All these look like some constants

library/types/src/modules/String.rb
```diff
@@ -499,24 +499,17 @@ def ParseOptions(options, parameters)
...
-
-              if Builtins.haskey(backslash_seq, nextcharacter) == true
-                character = Ops.get_string(
-                  backslash_seq,
-                  nextcharacter,
-                  "DUMMY"
-                )
+                "\\" => "\\", # backslash
+                # backslash will be removed later,
+                # double quote and escaped double quote have to different yet
```
kobliha: `have to different yet`?

library/packages/src/modules/PackageCallbacks.rb
jreidinger: yep, described in method description for possible values to return

library/types/src/modules/String.rb
jreidinger: I just copy it, to be honest I have no idea what it mean and hope there is good enough testsuite for this parser :)

library/packages/src/modules/PackageCallbacks.rb
kobliha: I know, I've seen it. I don't have a strong opinion on this. From my POV a variable just should describe which information it contains.

library/packages/src/modules/PackageCallbacks.rb
kobliha: See it there, sure.

library/types/src/modules/String.rb
kobliha: Then I'd rather delete it, it doesn't tell anything if nobody understands it :)

.rubocop.yml
```diff
@@ -2,41 +2,29 @@
...
 Metrics/ParameterLists:
   Max: 11
 
-# Offense count: 179
 Metrics/PerceivedComplexity:
-  Max: 76
+  Max: 70
 
 # Configuration parameters: EnforcedStyle, SupportedStyles.
 # JR: I personally think that having this variable on class level is much more readable
```
kobliha: BTW, just a side note: and do you think such comment belongs to code?

kobliha:
No big deal, a few different expectations and POVs, but the code is much better. Well done!
LGTM

library/types/src/modules/String.rb
jreidinger: I hope @mvidner understand :)

.rubocop.yml
jreidinger: @kobliha it is not in shared yast style guide, so I explain here why I use it here. I think we should agreed on that. Here I select one option to be consistent.

jreidinger:
I fixed two comment issues, so lets merge it.

mvidner:
@jreidinger Another problem with this PR is that it removes `getLockedProposals`. Its caller has been removed too, in https://github.com/yast/yast-installation/pull/247 , BUT that PR did not bump the package version. Therefore this failure <https://openqa.opensuse.org/tests/44408/modules/installation_overview/steps/2>.

jreidinger:
@mvidner ah, you are right. I think I forgot something and it as always appear :( . Thanks for fixing it.

PR #326, opened 2015-01-22
added handling of new created configuration files (bsc#860856)

jsrain:


library/system/src/modules/FileChanges.rb
```diff
@@ -221,9 +221,61 @@ def CheckFiles(files)
...
+        if new_files.size > 1
+          # Continue/Cancel question, %s is a comma separated list of file names
+          msg = _(
+            "Files %s have been created manually.\nYaST might lose these files."
+          )
+        end
+        msg = msg % new_files.join(", ")
+        popup_file = "/filechecks_non_verbose"
+        popup_file_path = File.join(Directory.vardir, popup_file)
+        if !FileUtils.Exists(popup_file_path)
```
jreidinger: so if I get it is persistent option. So if I in bootloader said do not show me warning again it is never again shown for any module, right?

jreidinger:
code-wise it is ok. Just it would be if there is test case for this method.

library/system/src/modules/FileChanges.rb
jsrain: This is how old (or known from before) files are handled (see CheckFiles method). This way it is consistent and there were no complaints about the behavior yet (at least AFAIK).

jreidinger:
LGTM

PR #327, opened 2015-01-22
Package callbacks cleaning

jreidinger:
recommended to review per commit.
still WIP, do not merge yet ( it is here for early comments )
you can see improvement in metrics -


lslezak:
Syntax check reports several errors like `library/packages/src/modules/PackageCallbacks.rb:427: syntax error, unexpected ')', expecting keyword_end`

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -3762,6 +3612,110 @@ def PackageCallbacks
...
+          ReplacePoint(Id(:info), Empty()),
+          HSpacing(0.1)
+        ),
+        HBox(HSpacing(0.1), button_box, HSpacing(0.1)),
+        VSpacing(0.2)
+      )
+    end
+
+    def textmode
+      if Mode.commandline
```
lslezak: I think it could be simplified to
```ruby
Mode.commandline || UI.TextMode
```

jreidinger: good idea

jreidinger:
@lslezak that is last minute fixes, will fix it. I am sure it will help when I write test if it works as expected

lslezak:
LGTM, thanks!

PR #329, opened 2015-01-26
Fix killing 12

jreidinger:


library/types/src/modules/String.rb
```diff
@@ -492,11 +492,11 @@ def ParseOptions(options, parameters)
...
-                "b"  => "", # backspace
+                "b"  => "\b", # backspace
                 "e"  => "e", # escape
-                "f"  => "", # FF
+                "f"  => "\f", # FF
                 "n"  => "\n", # NL
-                "r"  => " ", # CR
+                "r"  => "\b", # CR
                 "t"  => "\t", # tab
                 "v"  => "v", # vertical tab
```
lslezak: Interestingly, the backslash was lost before `v`. It looks like a different YCP-killer bug.

See the original code before the conversion: https://github.com/yast/yast-yast2/blob/good-bye-ycp/library/types/src/String.ycp#L518

jreidinger: thanks for catching, Thankfully it is only place here

lslezak:
LGTM, thanks! :+1:

mvidner:
@jreidinger @lslezak it seems you have replaced carriage return with a
backspace.
Dne 2015 1 27 10:53 "Josef Reidinger" <notifications@github.com> napsal(a):

jreidinger:
@mvidner where?

PR #330, opened 2015-01-26
Removed more icons from other dialogs (bnc#875201)

cwh42:


aschnell:
REQ: Changelog is missing. Esp. since features are removed.

library/wizard/src/modules/Confirm.rb
```diff
@@ -55,7 +54,7 @@ def main
 
     # Confirm hardware detection (only in manual installation)
     # @param [String] class hardware class (network cards)
-    # @param [String] icon_name name of the icon. If nil, generic hardware icon will be used
+    # @param [String] obsolete - just for compatibility
     # @return true on continue
     def Detection(class_, icon_name)
```
jreidinger: you use it somewhere or rubocop should complain, that unused parameter should start with _

cwh42:
So there are two public methods whose API contain an icon-parameters. I kept those parameters to not change the API. Maybe an deprecation warning would be the right thing here but I am not sure if we do such things as deprecation warnings.
Changelog follows.

library/wizard/src/modules/Progress.rb
```diff
@@ -677,33 +586,9 @@ def SubprogressTitle(title)
...
-      tits = deep_copy(tits)
-      icons_definition = deep_copy(icons_definition)
-      @global_visible_icons_definition = Ops.get(icons_definition, 0, [])
-      @global_invisible_icons_definition = Ops.get(icons_definition, 1, [])
-
-      @use_icons_in_progress = true
       New(window_title, progress_title, length, stg, tits, help_textmap)
-      @use_icons_in_progress = false
-
       nil
```
jreidinger: New already return nil, so this is not needed

jreidinger:
travis failed on rubocop rules and I am right, you do not mark unused parameters according to convention

lslezak:
@cwh42 to the deprecation:

- You can use `@deprecated` yard tag (see http://www.rubydoc.info/gems/yard/file/docs/Tags.md#deprecated)
- You can log the deprecation warning (e.g. via `log.warn`) with the caller location using `caller(0)` (or the old way via `Builtins.y2warning(-1, message)` to easily find where it was called from

cwh42:
Added deprecation warnings and updated documentation.
Changelog added.

jreidinger:
LGTM, from my side

lslezak:
LGTM

PR #331, opened 2015-01-27
Fix killing

jreidinger:


library/general/test/test_helper.rb
```diff
@@ -3,13 +3,8 @@
...
 
-def assign_root_path(directory)
-  check_version = false
-  handle = Yast::WFM.SCROpen("chroot=#{directory}:scr", check_version)
-  Yast::WFM.SCRSetDefault(handle)
-end
-
-def reset_root_path
-  Yast::WFM.SCRClose(Yast::WFM.SCRGetDefault)
+RSpec.configure do |c|
```
lslezak: Is this block needed? It is already present in `scr_stub.rb` file which is required here.

library/runlevel/test/service_test.rb
```diff
@@ -35,7 +35,7 @@ def stub_service_with(method, result)
       end
 
       it "returns the result of the original result of the command call" do
-        expect(Service.call("status", "sshd")).to be_kind_of(String)
+        expect(Service.call("status", "sshd")).to be_kind_of(::String)
```
lslezak: Ouch, we should get rid of the `String` module, the conflict with the standard Ruby class makes me always crazy...

library/general/test/scr_stub.rb
```diff
@@ -0,0 +1,81 @@
...
+      raise "Error closing the chrooted SCR instance, it's not the current default one"
+    end
+
+    Yast::WFM.SCRClose(default_handle)
+    Yast::WFM.SCRSetDefault(@scr_handles.last)
+  end
+end
+
+RSpec.configure do |c|
+  c.include SCRStub
```
lslezak: When we are adding this config section we should enable only the new RSpec syntax (like here: https://github.com/yast/yast-registration/blob/master/test/spec_helper.rb#L19), that would break on `should_receive` etc. when used.

Some places are fixed in this PR, that would ensure we do not introduce them later.

library/general/test/test_helper.rb
lslezak: Or rather fix it the other way round, remove it from `scr_stub.rb` and leave it here.

library/general/test/scr_stub.rb
jreidinger: @ancorgs plan in very near future to get rid of this file and use shared one in ruby bindings. I just merge SLE-12 branch into master to not loose any fixes. so this part probably will change early.

lslezak:
LGTM

PR #332, opened 2015-01-27
Be less verbose at Travis to have shorter log

lslezak:
- removed `rake check:syntax`, not needed anymore, the syntax is checked by rubocop
- use `-s` (silent) make option

The log size has been decreased from [~4600 lines](https://travis-ci.org/yast/yast-yast2/builds/48464147) to [~2500 lines](https://travis-ci.org/yast/yast-yast2/builds/48468523), which is a huge difference (almost half of the original size!).


jreidinger:
LGTM

PR #333, opened 2015-01-28
Tests fixes

ancorgs:
* Fix some broken tests
 * Rely on RSpec extensions from ruby-bindings instead of custom SCRStub
 * Add BuildRequires on modern ruby-bindings


jreidinger:
LGTM, nice improvement

PR #334, opened 2015-01-28
Merging SLE12-GA into 13.2 to get the new robust tests...

ancorgs:
...and many other things, according to the diff. Did we actually changed all that code in SLE-12-GA?


jreidinger:
yep, as it looks contain rubocop changes, which was quite big. LGTM.

jreidinger:
I just hope we do not release maintenance update, otherwise I worry, that maintenance guys will throw it on our head :)

kobliha:
@jreidinger  That's easy. It's open-source, so "take it or leave it"... :)

PR #336, opened 2015-01-29
Coveralls

jreidinger:


lslezak:
NTH: please add the coveralls badge to the README.md (after the build status badges so we have unified order across repos).

lslezak:
Good, I'd just prefer the SVG icon (instead of PNG), the already used badges are in SVG format.

lslezak:
LGTM, thanks! :+1:

PR #338, opened 2015-01-30
String module testsuite conversion and small cleanup

jreidinger:


ancorgs:
Apart from the yardoc comment, shouldn't we log somehow the use of deprecated methods.

jreidinger:
@ancorgs well, in general I prefer to use grepping over source before logging, because log is only from runtime system and need not find all usage, but grep will find all usage of old method. So for me it is better method to use. Also deprecation warning can break old testsuite if you make it too serious ( anything above milestone ).

library/types/src/modules/String.rb
```diff
@@ -369,14 +358,15 @@ def Repeat(text, number)
     # @param [Symbol] alignment alignment to use, either `left or `right
     # @return padded text
     def SuperPad(text, length, padding, alignment)
-      text = "" if text.nil?
+      text ||= ""
+      return text if length.nil? || text.size >= length || padding.nil?
 
-      pad = Repeat(padding, Ops.subtract(length, Builtins.size(text)))
+      pad = padding * (length - text.size)
```
ancorgs: Maybe is a good idea to check that ```padding.size == 1``` or even to cut it down otherwise.

jreidinger: well, then it won't be backward compatible. But I agree that it looks like buggy ( old code also do repetition ).

library/types/test/string_test.rb
```diff
@@ -0,0 +1,176 @@
...
+      expect(subject.UnQuote("a'\\'''\\''b'\\''c")).to eq "a''b'c"
+    end
+  end
+
+  describe ".FormatSize" do
+    it "returns empty string if nil passed" do
+      expect(subject.FormatSize(nil)).to eq ""
+    end
+
+    FORMAT_SIZE_DATA = {
```
ancorgs: Why not use ```let``` or simply a variable instead of this constant?

jreidinger: maybe old habit to use constants for constant data :) No strong opinion for any solution.

library/types/test/string_test.rb
```diff
@@ -0,0 +1,176 @@
...
+      end
+    end
+  end
+
+  describe ".FormatSizeWithPrecision" do
+    it "returns empty string if nil passed" do
+      expect(subject.FormatSizeWithPrecision(nil, nil, nil)).to eq ""
+    end
+
+    it "returns bytes in proper unit with passed precision forcing trailing zeroes if omit_zeroes not passed" do
```
ancorgs: This description is quite hard to read and understand. Maybe breaking it in two or even using contexts (to group omit_zeroes true/false, for example) would help.

library/types/src/modules/String.rb
ancorgs: Let's keep it fully backward compatible, but I'd add a testcase with a bigger string documenting this "buggy" behavior, so we don't forget about it.

library/types/test/string_test.rb
ancorgs: Actually it's mainly a matter of taste, but I find those constants "floating" in the describe to be kind of weird and ```let``` to be more RSpec-like.

library/types/src/modules/String.rb
jreidinger: good idea, done

library/types/test/string_test.rb
jreidinger: well I prefer single sentence over single use context as it add additional nesting, which is harder to follow for me, but matter of taste

ancorgs:
Apart from some small irrelevant typos in some RSpec descriptions, it LGTM

jreidinger:
OK, so lets merge and enjoy nicer String module :)

PR #339, opened 2015-02-02
fix typo in method name

jreidinger:


lslezak:
LGTM, thanks for the quick fix! :+1:

jreidinger:
I broke it, so I will fix it. It is nice that openQA catch it so quickly.

PR #341, opened 2015-02-03
Merge of fix for bnc#912169 into 13.2

ancorgs:


jreidinger:
LGTM, even if test case would be nice, but I know it is hard to do in client.

PR #318, opened 2015-01-13
Rubocop sle12 attempt2

jreidinger:
ready to review

PR #343, opened 2015-02-12
Fix error introduced while trying to please rubocop

ancorgs:


jreidinger:
good catch, LGTM

ancorgs:
It was not me who found it. The error was distributed as a maint. update for both SLE12 and openSUSE 13.2 some days ago. The users found it :-/

kobliha:
I thought we have a rule: test-cases, don't we?

ancorgs:
@kobliha If you mean this PR not including tests. It's actually an urgent and obvious fix to mitigate the effect of the broken maintenance update. If you mean the error being introduced in the first place, it was as part of pull request #318, which in one hand looks a little bit too risky for an stable branch but on the other hand allowed to introduce rubocop in master without introducing a huge divergence between branches.

kobliha:
@ancorgs Yes, I'm talking about the #318 because `how much we haven't found`? I know that judging today is easier than earlier, but we should at least learn from it.

PR #345, opened 2015-02-12
Merge of #344 into master

ancorgs:


jreidinger:
nothing to review :)

ancorgs:
the affected code is not longer in master, so I'm only merging to sync branches

jreidinger:
yep, I understand and I hope my code cleaned code do not contain it :) LGTM

ancorgs:
Don't worry, I checked it https://github.com/yast/yast-yast2/blob/master/library/packages/src/modules/PackageCallbacks.rb#L563

PR #347, opened 2015-03-02
update to match style with rubocop 0.29.1

jreidinger:


library/types/src/modules/URLRecode.rb
```diff
@@ -47,7 +47,7 @@ def escape(input, safe_chars)
       return nil if input.nil?
       input.gsub(/[^#{safe_chars}]/) do |unicode_char|
         escaped = ""
-        unicode_char.each_byte { |b| escaped << sprintf("%%%02x", b) }
+        unicode_char.each_byte { |b| escaped << format("%%%02x", b) }
```
mvidner: So this is the only non-whitespace change in the code... and it uses a different alias for the same method. OK!

mvidner:
This looks safer than the last time we applied Rubocop to this repo :)
Please add a .changes entry.

> - Use Rubocop 0.29.1, updated the style to match it.

library/types/src/modules/URLRecode.rb
jreidinger: yep, new version of rubocop prefer format over sprintf. I think it can be switched to prefer sprintf over format. I think goal is to use just one alias to avoid confusion.

jreidinger:
yep, now changes are not much intrusive. Only think that I dislike is `if` indentation on multiline. But in general it looks stable and not so buggy, so good candidate for unified version.

jreidinger:
@mvidner: I do not agree that it deserve entry in changelog because it have no user impact. Anyone using YaST2 should not see it, it is just internal change. I think we should more focuse on user oriented changes like removing deprecated methods as it is sometimes challenging to find what happens with method from SLE11 ( hint OES modules ;)

mvidner:
Fair enough, because the refactoring looks safe.
LGTM.

jreidinger:
thanks

PR #348, opened 2015-04-08
use suse bugzilla url (bnc#924427)

jreidinger:
I decide to use the simpliest approach as there is only another instance in yast2-registration, so do not spend too much time in it.


jreidinger:
thanks

PR #350, opened 2015-05-13
iSCSI firewall proposal for SLE 12 (L3/Maintenance Update)

kobliha:
- Propose SuSEfirewal2 to fully initialize (e.g. open ports) already in the init phase to allow using iSCSI (bsc#916376)

### DIFF
--- a/library/network/src/modules/SuSEFirewall.rb
+++ b/library/network/src/modules/SuSEFirewall.rb
@@ -202,7 +202,7 @@ module Yast
         "FW_CONFIGURATIONS_INT",
         "FW_CONFIGURATIONS_DMZ",
         # bsc#916376: Ports need to be open already during boot
-        "FW_BOOT_FULL_INIT",
+        "FW_BOOT_FULL_INIT"
       ]
 
       @one_line_per_record = [
--- a/library/network/src/modules/SuSEFirewall.rb
+++ b/library/network/src/modules/SuSEFirewall.rb
@@ -38,6 +38,9 @@ module Yast
 
     include Yast::Logger
 
+    YES = "yes"
+    NO  = "no"
+
     def main
       textdomain "base"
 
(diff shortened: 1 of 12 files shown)


library/network/src/modules/SuSEFirewall.rb
```diff
@@ -38,6 +38,9 @@ class SuSEFirewallClass < Module
 
     include Yast::Logger
 
+    YES = "yes"
+    NO  = "no"
```
jreidinger: better way is to create mapping between boolean like
```
INIT_MAPPING = {
  true  => "yes",
  false => "no"
}
```

and below you can have `@SETTINGS["FW_BOOT_FULL_INIT"] = INIT_MAPPING[new_state]`

library/network/test/susefirewall_proposal_test.rb
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env rspec
+
+top_srcdir = File.expand_path("../../../..", __FILE__)
+inc_dirs = Dir.glob("#{top_srcdir}/library/*/src")
+ENV["Y2DIR"] = inc_dirs.join(":")
```
jreidinger: please, use helper to share such stuff

library/network/src/modules/SuSEFirewall.rb
kobliha: and also `@SETTINGS["FW_BOOT_FULL_INIT"] == INIT_MAPPING[true]` ?
that sounds strange

library/network/test/susefirewall_proposal_test.rb
kobliha: OK, I can, this was just copied from another test

library/network/src/modules/SuSEFirewall.rb
jreidinger: it is not strange, it say what it should. You set true to given settings. Of course in perfect OOP world, it will be object and set it with enable/disable

jreidinger:
LGTM

PR #351, opened 2015-05-14
iSCSI firewall proposal for master / SLE 12

kobliha:
- merged from #350 to master


jreidinger:
LGTM, if travis passed ( see no obvious reason why it should failed )

PR #352, opened 2015-05-15
Feature/add fs snapshots

imobach:


library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+    #
+    # @return [true,false] true if it's configured; false otherwise.
+    def self.configured?
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), FIND_CONFIG_CMD)
+      out["exit"] == 0
+    end
+
+    # Configures snapper
+    #
+    # @return [true,false] true if it's configured; false otherwise.
```
jreidinger: if I get it correctly it mean that it return false only if it failed to configure? Is it common case? If not then exception will be better

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+    def self.configure
+      if configured?
+        true
```
jreidinger: I think early exit is better style like
```
return true if configured?

out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), CREATE_CONFIG_CMD)
...
```

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    # Creates a new snapshot
+    #
+    # It raises and exception if Snapper is not configured.
+    #
+    # @param description [String] Snapshot's description.
+    # @return [FsSnapshot,nil] The created snapshot if the operation was
+    #                          successful. Otherwise, it returns nil.
+    def self.create(description)
+      raise SnapperNotConfigured unless configured?
```
jreidinger: just my preference but I prefer have empty line between validation of state and real commands.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    # Returns all snapshots
+    #
+    # It raises and exception if Snapper is not configured.
+    #
+    # @return [Array<FsSnapshot>] All snapshots that exist in the system.
+    def self.all
+      raise SnapperNotConfigured unless configured?
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), LIST_SNAPSHOTS_CMD)
+      lines = out["stdout"].split(/\n/)[2..-1] # relevant lines from output.
```
jreidinger: quite fragile if output of snapped changed, better using something like `lines = out["stdout"].lines.grep(VALID_LINE_REGEXP)`

imobach:
I'll take a look at rubocop issues.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+      lines.map do |line|
+        data = line.split(/\s*\|\s*/)
```
jreidinger: I do not like this faking strip in split. Better be more explicit so it is easier to read like `data = line.split("|").map(&:strip)`

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+        timestamp = data[3] == "" ? nil : DateTime.parse(data[3])
+        new(data[1].to_i, data[0].to_sym, data[2].to_i, timestamp, data[4],
```
jreidinger: so `previous` is snapshot number and not snapshot itself? Quite unintuitive

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+      all.find { |s| s.number == number }
+    end
+
+    def initialize(number, snapshot_type, previous, timestamp, user, cleanup, description)
+      @number = number
+      @snapshot_type = snapshot_type
+      @previous = previous
+      @timestamp = timestamp
+      @user = user
+      @cleanup = cleanup
```
jreidinger: what is cleanup? I am not sure if it is good name. If I see in code `snapshot.cleanup` I except something different

library/system/test/fs_snapshot_test.rb
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env rspec
+
+require_relative "test_helper"
+#require_relative "../src/lib/yast2/fs_snapshot"
```
jreidinger: dead code, please remove

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+      @description = description
+    end
+  end
+end
```
jreidinger: One more idea. We need more logging, ideally for all bash call to see what is called and what is result. It really helps with debugging on customer side when we do not know, what can go wrong. If we have logs we can clearly say - snapper return wrong value or it return correct data and problem is in our code.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+        out["exit"] == 0
+      end
+    end
+
+    # Creates a new snapshot
+    #
+    # It raises and exception if Snapper is not configured.
+    #
+    # @param description [String] Snapshot's description.
+    # @return [FsSnapshot,nil] The created snapshot if the operation was
```
mchf: NP: I don't see into snapper at all, but it if I understand it correctly it returns a sort of snapshot id. Snapshot and snapshot id has slightly different meaning to me (in case of snapshot I'd expect something else than number).

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+    # It raises and exception if Snapper is not configured.
+    #
+    # @return [Array<FsSnapshot>] All snapshots that exist in the system.
+    def self.all
+      raise SnapperNotConfigured unless configured?
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), LIST_SNAPSHOTS_CMD)
+      lines = out["stdout"].split(/\n/)[2..-1] # relevant lines from output.
+      lines.map do |line|
+        data = line.split(/\s*\|\s*/)
+        timestamp = data[3] == "" ? nil : DateTime.parse(data[3])
```
mchf: if format of data will change in the future this dies with exception

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,125 @@
...
+        nil
+      end
+    end
+
+    # Returns all snapshots
+    #
+    # It raises and exception if Snapper is not configured.
+    #
+    # @return [Array<FsSnapshot>] All snapshots that exist in the system.
+    def self.all
```
mchf: I'd prefer more verbose name like ```all_snapshots```

library/system/src/lib/yast2/fs_snapshot.rb
imobach: Yes, I've renamed to `previous_number` and I've added a `previous` method that retrieve that previous snapshot.

library/system/src/lib/yast2/fs_snapshot.rb
imobach: Yes, it's the "cleanup algorithm". Renamed to `cleanup_algo`.

library/system/src/lib/yast2/fs_snapshot.rb
imobach: I'm a REGEX fan :) Anyway, I changed it because you're right and it could be more readable.

library/system/src/lib/yast2/fs_snapshot.rb
imobach: Fixed.

library/system/src/lib/yast2/fs_snapshot.rb
imobach: It shouldn't be the common case.

library/system/src/lib/yast2/fs_snapshot.rb
imobach: Yes, you're right, they're different things. Snapper uses a `number` to identify each snapshot. Anyway, this method returns the whole snapshot (an `FsSnapshot` instance).

library/system/src/lib/yast2/fs_snapshot.rb
jreidinger: you call it with class always, so it looks like `FsSnapshot.all` so snapshots will be duplicite here

library/system/src/lib/yast2/fs_snapshot.rb
jreidinger: another nice thing is that it also strip whitespaces before and after first, respective last element

library/system/src/lib/yast2/fs_snapshot.rb
jreidinger: exception is fine from my POV. We catch it early, so no problem and we can adapt. I see no way how to handle changed format in better way. Any ideas?

library/system/src/lib/yast2/fs_snapshot.rb
jreidinger: question is what is use case for previous_number reader. Who will need it? maybe previous itself will be enough? I like minimal API, so no exposed call in API without use case for it.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,152 @@
...
+  class SnapperNotConfigured < StandardError
+    def initialize
+      super "Snapper is not configured."
+    end
+  end
+
+  # Snapper could not be configured.
+  class SnapperConfigurationFailed < StandardError
+    def initialize
+      super "Snapper could not be configured."
```
jreidinger: Quite useless message, I think it make sense to add reason to message.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,152 @@
...
+# Authors:
+#	Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+require "yast"
+
+module Yast2
+  # Represents the fact that Snapper is not configured for "/" (root).
+  class SnapperNotConfigured < StandardError
+    def initialize
+      super "Snapper is not configured."
```
jreidinger: if user see it, it do not help. Please add advice what to do if user see it ( I expect it is programmer error, so advise to report bug that there is wrong work-flow )

library/system/src/lib/yast2/fs_snapshot.rb
jreidinger: OK, great that you add expection. So I think return value is now useless. so it can look like:
```
return if configured?

out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), CREATE_CONFIG_CMD)
if out["exit"] != 0
  log.error("Snapper configuration failed: #{CREATE_CONFIG_CMD} returned: #{out}")
  raise SnapperConfigurationFailed
end
...
```

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,152 @@
...
+    #
+    # It raises and exception if Snapper is not configured.
+    #
+    # @return [Array<FsSnapshot>] All snapshots that exist in the system.
+    def self.all
+      raise SnapperNotConfigured unless configured?
+
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), LIST_SNAPSHOTS_CMD)
+      lines = out["stdout"].lines.grep(VALID_LINE_REGEX) # relevant lines from output.
+      log.info("Retrieving snapshots list: #{LIST_SNAPSHOTS_CMD}. Found #{lines.size} snapshot(s).")
```
jreidinger: ok, consider that there something changed in output and you get 0 snapshots, this line won't help you much. Usually it is good idea to capture input or output on "borders" which is in this case CLI. So I think it is better to log output itself.

library/system/src/lib/yast2/fs_snapshot.rb
mchf: well, not always. ```all.find``` looks weird at line 135. but its NP

library/system/src/lib/yast2/fs_snapshot.rb
imobach: Yes, you're right. Anyway, the list of snapshots is not expected to be that long.

library/system/src/lib/yast2/fs_snapshot.rb
imobach: I expect the user of `FsSnapshot` to be a programmer. Also these messages are not translated. But you're right: I could make the programmer live easier if I point him in the right direction.

library/system/src/lib/yast2/fs_snapshot.rb
imobach: Yes. I've added the command which failed.

jreidinger:
LGTM, maybe @aschnell have some notes regarding needs for snapshot API

imobach:
Ok. We wait until monday for @aschnell. Thank you!

aschnell:
I don't see why you need the command to configure snapper here since that is done already in yast2-storage. And the command used here is unfortunately not appropriate for installation into a snapshot.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,154 @@
...
+    #
+    # It raises and exception if Snapper is not configured.
+    #
+    # @param description [String] Snapshot's description.
+    # @return [FsSnapshot,nil] The created snapshot if the operation was
+    #                          successful. Otherwise, it returns nil.
+    def self.create(description)
+      raise SnapperNotConfigured unless configured?
+
+      cmd = CREATE_SNAPSHOT_CMD % description
```
aschnell: Here options for the snapshot type and pre-num are missing.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,154 @@
...
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), cmd)
+      if out["exit"] == 0
+        find(out["stdout"].to_i) # The CREATE_SNAPSHOT_CMD returns the number of the new snapshot.
+      else
+        log.error "Snapshot could not be created: #{CREATE_SNAPSHOT_CMD} returned: #{out}"
```
aschnell: Here you assume that log.error return nil. Might be correct but does not seem robust.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,154 @@
...
+      true
+    end
+
+    # Creates a new snapshot
+    #
+    # It raises and exception if Snapper is not configured.
+    #
+    # @param description [String] Snapshot's description.
+    # @return [FsSnapshot,nil] The created snapshot if the operation was
+    #                          successful. Otherwise, it returns nil.
```
aschnell: The error handling seems inconsistent: If snapper is not configured -> exeception, if snapshot creation fails -> nil.
That will require users of the function to check for both.

aschnell:
Changelog is missing.

library/system/src/lib/yast2/fs_snapshot.rb
imobach: I didn't though about giving support until today :) I'm working on that now.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+          timestamp = DateTime.parse(data[3])
+        rescue ArgumentError
+          timestamp = nil
+        end
+        new(data[1].to_i, data[0].to_sym, data[2].to_i, timestamp, data[4],
+          data[5].to_s.to_sym, data[6])
+      end
+    end
+
+    # Finds an snapshot by its number
```
aschnell: "an" only before vowels, https://www.englishclub.com/pronunciation/a-an.htm.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+    #
+    # It raises and exception if Snapper is not configured.
```
aschnell: "and" -> "an"

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+    #
+    # @param description     [String] Snapshot's description.
+    # @param previous_number [Fixnum] Number of the previous snapshot
+    # @return [FsSnapshot] The created snapshot.
+    #
+    # @see FsSnapshot.create
+    def self.create_post(description, previous_number = nil)
+      previous =
+        if previous_number.nil?
+          last = all.reverse.find { |s| s.snapshot_type == :pre || s.snapshot_type == :post }
```
jreidinger: I really do not recommend to use `reverse.find` as there is chance, that list of all snapshot can be quite big and then you create new array in reverse order which can be time consuming. Same effect can be done with
```
all.reverse_each.find
```
which is lazy. It is lazy reverse without creating new array. I can provide also performance proof, if you are really interested.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+    # 'pre' snapshot is not given, it will try to figure out that number.
+    #
+    # @param description     [String] Snapshot's description.
+    # @param previous_number [Fixnum] Number of the previous snapshot
+    # @return [FsSnapshot] The created snapshot.
+    #
+    # @see FsSnapshot.create
+    def self.create_post(description, previous_number = nil)
+      previous =
+        if previous_number.nil?
```
jreidinger: previous number is number, so explicit nil check is not needed. so it can look like
```
  previous = 
     if previous_number
        find(previous_number)
     else
        ...
```

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+    # @param snapshot_type [Symbol]    Snapshot's type: :pre, :post or :single.
+    # @param description   [String]    Snapshot's description.
+    # @param previous      [FsSnashot] Previous snapshot.
+    # @return [FsSnapshot] The created snapshot if the operation was
+    #                      successful.
+    private_class_method def self.create(snapshot_type, description, previous = nil)
+      raise SnapperNotConfigured unless configured?
+
+      cmd = CREATE_SNAPSHOT_CMD % [snapshot_type, description]
+      cmd << " --pre-num %s" % previous.number unless previous.nil?
```
jreidinger: previous is FsSnapshot, so not boolean, so explicit nil check is not needed so if can be `if previous`

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), cmd)
+      if out["exit"] == 0
+        find(out["stdout"].to_i) # The CREATE_SNAPSHOT_CMD returns the number of the new snapshot.
```
jreidinger: I think it make sense to log content of stdout, so we know number of created snapshot, because there is a risk if output slightly change, that `to_i` return `0` which can be quite hard to debug.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), LIST_SNAPSHOTS_CMD)
+      lines = out["stdout"].lines.grep(VALID_LINE_REGEX) # relevant lines from output.
+      log.info("Retrieving snapshots list: #{LIST_SNAPSHOTS_CMD} returned: #{out}")
+      lines.map do |line|
+        data = line.split("|").map(&:strip)
+        begin
+          timestamp = DateTime.parse(data[3])
+        rescue ArgumentError
+          timestamp = nil
```
jreidinger: please at least log data that is not valid DateTime, otherwise we have later trouble to recognize that parsing date failed. ( I know we log complete list, but specific date entry can help with debugging.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+        end
+        new(data[1].to_i, data[0].to_sym, data[2].to_i, timestamp, data[4],
+          data[5].to_s.to_sym, data[6])
```
jreidinger: why to_s is needed? It should be string by default, not?

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -0,0 +1,198 @@
...
+    # @param description     [String] Snapshot's description.
+    # @param previous_number [Fixnum] Number of the previous snapshot
+    # @return [FsSnapshot] The created snapshot.
+    #
+    # @see FsSnapshot.create
+    def self.create_post(description, previous_number = nil)
+      previous =
+        if previous_number.nil?
+          last = all.reverse.find { |s| s.snapshot_type == :pre || s.snapshot_type == :post }
+          (last && last.snapshot_type == :pre) ? last : nil
```
aschnell: In general detecting the pre-num is fragile since there can be several pre/post pairs even betweem pre/post pairs. For installation it's likely OK since no other task runs at that time.

imobach: I guess that it could break also during upgrades/updates because zypper is ran, isn't it?

aschnell: On Tue, May 19, 2015 at 07:22:39AM -0700, Imobach González Sosa wrote:
> > +
> > +    # Creates a new 'post' snapshot
> > +    #
> > +    # Each 'post' snapshot corresponds with a 'pre' one. If the number of the
> > +    # 'pre' snapshot is not given, it will try to figure out that number.
> > +    #
> > +    # @param description     [String] Snapshot's description.
> > +    # @param previous_number [Fixnum] Number of the previous snapshot
> > +    # @return [FsSnapshot] The created snapshot.
> > +    #
> > +    # @see FsSnapshot.create
> > +    def self.create_post(description, previous_number = nil)
> > +      previous =
> > +        if previous_number.nil?
> > +          last = all.reverse.find { |s| s.snapshot_type == :pre || s.snapshot_type == :post }
> > +          (last && last.snapshot_type == :pre) ? last : nil
> 
> I guess that it could break also during upgrades/updates because zypper is ran, isn't it?

If snapper-zypp-plugin is installed then yes.

imobach: I guess that a piece of code like that should work (using a 'stack'):

```ruby
private_class_method def self.find_unpaired_pre_snapshot
      unpaired = []
      all.each do |s|
        if s.snapshot_type == :post
          unpaired.pop
        else s.snapshot_type == :pre
          unpaired.push(s)
        end
      end
      unpaired.pop
    end
```

imobach: I read your comment on IRC and it seems that it won't work either. So I'll store the number then. Thanks a lot!

jreidinger:
@imobach thanks for changes, also please ensure it is mergable. So remaining thing is to solve searching previous pre, which is not valid if I get discussion

imobach:
@jreidinger I'll have to follow another approach (maybe at client level) and store the pre-num.

jreidinger:
LGTM

PR #353, opened 2015-05-15
Fix Travis locale

lslezak:
Fixed sending coveralls report at Travis - it failed when the commit author name contained an UTF-8 (non-ASCII) character.

I also moved the locale setting just to the tests to not potentially influence the other commands (to be compatible with the other Travis scripts).

(BTW locally it works without the `LC_ALL=` setting but failed without it at Travis. It took me some time to figure this out...)


jreidinger:
LGTM

PR #354, opened 2015-05-18
Make iscsi_proposal public

kobliha:
- We have found out that there are two firewall proposals: one used in second stage, but the other one used in first stage. I'm sharing the functionality by using SuSEFirewallProposal
- The first stage proposal lives at yast/yast2-network

### DIFF
--- a/library/network/src/modules/SuSEFirewallProposal.rb
+++ b/library/network/src/modules/SuSEFirewallProposal.rb
@@ -771,8 +771,6 @@ module Yast
       { "output" => output, "warning" => warning }
     end
 
-  private
-
     # Proposes firewall settings for iSCSI
     def propose_iscsi
       log.info "iSCSI has been used during installation, opening #{@iscsi_target_service} service"
@@ -781,6 +779,8 @@ module Yast
 
       # bsc#916376: ports need to be open already during boot
       SuSEFirewall.full_init_on_boot(true)
+
+      nil
     end
 
     publish function: :OpenServiceOnNonDialUpInterfaces, type: "void (string, list <string>)"
@@ -791,6 +791,7 @@ module Yast
     publish function: :Reset, type: "void ()"
     publish function: :Propose, type: "void ()"
     publish function: :ProposalSummary, type: "map <string, string> ()"
+    publish function: :propose_iscsi, type: "void ()"
   end
 
   SuSEFirewallProposal = SuSEFirewallProposalClass.new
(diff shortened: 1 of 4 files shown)


library/network/src/modules/SuSEFirewallProposal.rb
```diff
@@ -781,6 +779,8 @@ def propose_iscsi
 
       # bsc#916376: ports need to be open already during boot
       SuSEFirewall.full_init_on_boot(true)
+
```
mvidner: What does the `true` argument mean?

library/network/test/susefirewall_proposal_test.rb
```diff
@@ -16,15 +36,11 @@
...
 
-    context "when iscsi is used" do
-      it "proposes opening iscsi-target firewall service and full firewall initialization on boot" do
-        allow(Yast::Linuxrc).to receive(:useiscsi).and_return("initial")
-
-        expect(Yast::SuSEFirewall).to receive(:full_init_on_boot).and_return(true)
-        expect(Yast::SuSEFirewall).to receive(:SetServicesForZones).with(["service:iscsitarget"], ["EXT"], true).and_return(true)
+    it "proposes opening iscsi-target firewall service and full firewall initialization on boot" do
+      expect(Yast::SuSEFirewall).to receive(:full_init_on_boot).and_return(true)
+      expect(Yast::SuSEFirewall).to receive(:SetServicesForZones).with(["service:iscsitarget"], ["EXT"], true).and_return(true)
```
mvidner: Here we are actually testing `OpenServiceOnNonDialUpInterfaces`. Maybe out of scope for this PR, but the tests are a bit mixed up.

kobliha: Yes and no, it tests what is called, not what is does with it. Besides, I'm just moving the current test.

mvidner:
Minor issues, LGTM.

library/network/src/modules/SuSEFirewallProposal.rb
kobliha: That's described here: https://github.com/yast/yast-yast2/blob/SLE-12-GA/library/network/src/modules/SuSEFirewall.rb#L3735-L3739

library/network/test/susefirewall_proposal_test.rb
kobliha: The code, which is called, is a bit more complex than it should be. Cleanup would be nice, but that's more than expected for an L3.

kobliha:
@mvidner thanks

PR #358, opened 2015-05-25
Fixed firewall proposal in installation (iSCSI-related)

kobliha:
- It's not needed to open any ports in firewall, only enable early initialization (bsc#916376)
- As a side effect, I've fixed functionality to open fallback ports when firewall service does not exist
- If network interfaces are not assigned to any zone (e.g., in initial stage of installation) fallback ports are open in all zones (worked the same way for firewall services)
- Including extensive test case

### DIFF
--- a/library/network/src/modules/SuSEFirewallProposal.rb
+++ b/library/network/src/modules/SuSEFirewallProposal.rb
@@ -252,6 +252,7 @@ module Yast
 
     # Function opens service for network interfaces given as the third parameter.
     # Fallback ports are used if the given service is uknown.
+    # If interfaces are not assigned to any firewall zone, all zones will be used.
     #
     # @see OpenServiceOnNonDialUpInterfaces for more info.
     #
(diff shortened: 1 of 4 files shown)


library/network/src/modules/SuSEFirewallProposal.rb
```diff
@@ -234,16 +230,13 @@ def ServiceEnabled(service, zones)
...
     def EnableFallbackPorts(fallback_ports, zones)
-      fallback_ports = deep_copy(fallback_ports)
-      zones = deep_copy(zones)
-      Builtins.y2warning(
-        "Enabling fallback ports: %1 in zones: %2",
-        fallback_ports,
-        zones
-      )
+      known_zones = SuSEFirewall.GetKnownFirewallZones()
+      unknown_zones = zones.reject{|zone| known_zones.include?(zone)}
```
jreidinger: `unknown_zones = known_zones - zones`

jreidinger:
just one improvement, otherwise LGTM

library/network/src/modules/SuSEFirewallProposal.rb
kobliha: Hmm, good idea :) thanks

jreidinger:
LGTM

PR #359, opened 2015-05-25
Fixed firewall proposal in installation (iSCSI-related) [master]

kobliha:
Port of #358 to master branch


jreidinger:
LGTM if travis passed ( and no it do not pass, due to rubocop ;)

PR #360, opened 2015-05-27
initial modification to support pre update snapshot

jreidinger:


library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -139,6 +152,36 @@ def self.create(snapshot_type, description, previous = nil)
     end
     private_class_method :create
 
+    # detects if module runs in initial stage before scr is switched to target system
```
imobach: np: It would be nice adding yard-style method documentation (`@return` and that kind of things).

jreidinger: I cannot found link, but I read somewhere that it is recommended to use it only when it make sense, so having param or return that just copy general description method is quite useless for me. Is there reason for it? E.g. in Store class I use it where it make sense ( like raising exception, or non-obvious return type )

imobach: @jreidinger It's fine with me. I try adding it always just for coherence.

imobach:
Just as a side note: regarding `FsSnapshotStore` class, I'm not a big fan of using class methods as the only API (I prefer creating a new instance). However, it's fine with me because it's a common pattern (and not only in YaST).

So, if Travis is happy, LGTM.

jreidinger:
@imobach well, now it is functional API, for OO one we need slightly different one with something like FsSnapshotStore.new(purpose) and load/save there, but I see no advancement for it. I think we can be a bit relaxed with strict OO or strict functional and should use what fit the best given usecase. Usually for smaller util classes functional API make sense and when it start growing it make sense to change it to OO API.

imobach:
@jreidinger Yes, it's fine and I understand the reasons behind that. I just wanted to make that point :)

imobach:
LGTM

PR #361, opened 2015-05-29
Don't log an error when fs snapshot has no date/time

imobach:
Avoid logging an error when date/time is just an empty string.


jreidinger:
LGTM

imobach:
Just as we have discussed, the only snapshot that have no date/time is the 'current' (and really is not an snapshot). So I'm closing this PR without merging (and I've open a new one: https://github.com/yast/yast-yast2/pull/362).

PR #362, opened 2015-05-29
FsSnapshot ignores current snapshot

imobach:
* It assumes that its id is '0'.


library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -198,8 +198,9 @@ def self.all
       end
       lines = out["stdout"].lines.grep(VALID_LINE_REGEX) # relevant lines from output.
       log.info("Retrieving snapshots list: #{LIST_SNAPSHOTS_CMD} returned: #{out}")
-      lines.map do |line|
+      lines.each_with_object([]) do |line, snapshots|
         data = line.split("|").map(&:strip)
+        next if data[1] == "0"
```
jreidinger: please add comment, as it is not so obvious

imobach: Yes, you're right. Done.

jreidinger:
LGTM

PR #363, opened 2015-06-01
Added Linuxrc.value_for

kobliha:
- We need to read Linuxrc commandline to find out whether Yast should create pre/post-install snapshots during Upgrade/Installation
- This new functionality enables us to simply call Linuxrc.value_for("disable_snapshots") and it will have the same result independently on the way how it's written on Linuxrc commandline - users don't need to follow strict guidelines and usually do not follow them (matches are case-insensiteve, dots, hyphens and underscores are ignored)

### DIFF
--- a/library/general/src/modules/Linuxrc.rb
+++ b/library/general/src/modules/Linuxrc.rb
@@ -222,18 +222,18 @@ module Yast
       feature_key = polish(feature_key)
       install_inf_features = deep_copy(@install_inf)
 
-      # Some values written on Linuxrc commandline are not know to Linuxrc
+      # Some values written on Linuxrc commandline are not known to Linuxrc
       # Unless they are also mentioned as `PTOptions`, they will appear as "Cmdline" entry
-      cmdline_values = install_inf_features.fetch("Cmdline", "").split.map do |cmdline_entry|
+      install_inf_features.fetch("Cmdline", "").split.map do |cmdline_entry|
         key_val = cmdline_entry.split("=")
         install_inf_features[key_val[0]] = key_val[1]
       end
 
-      matching_keys = install_inf_features.keys.select{|key| polish(key) == feature_key}
+      matching_keys = install_inf_features.keys.select { |key| polish(key) == feature_key }
       return nil if matching_keys.empty?
 
       # The last key wins as in Linuxrc
-      matching_keys.map{|key| install_inf_features[key]}.last
+      matching_keys.map { |key| install_inf_features[key] }.last
     end
 
     publish function: :ResetInstallInf, type: "void ()"
@@ -260,7 +260,6 @@ module Yast
     def polish(key)
       key.downcase.tr("-_\.", "")
     end
-
   end
 
   Linuxrc = LinuxrcClass.new
--- a/library/general/src/modules/Linuxrc.rb
+++ b/library/general/src/modules/Linuxrc.rb
@@ -211,6 +211,31 @@ module Yast
       true
     end
 
+    # Returns value of a given Linxurc key/feature defined on commandline
+    # and written into install.inf
+    #
+    # @param [String] key
+    # @return [String] value of a given key
+    def value_for(feature_key)
+      ReadInstallInf()
+
+      feature_key = polish(feature_key)
+      install_inf_features = deep_copy(@install_inf)
+
+      # Some values written on Linuxrc commandline are not know to Linuxrc
+      # Unless they are also mentioned as `PTOptions`, they will appear as "Cmdline" entry
+      cmdline_values = install_inf_features.fetch("Cmdline", "").split.map do |cmdline_entry|
+        key_val = cmdline_entry.split("=")
+        install_inf_features[key_val[0]] = key_val[1]
+      end
+
+      matching_keys = install_inf_features.keys.select{|key| polish(key) == feature_key}
+      return nil if matching_keys.empty?
+
+      # The last key wins as in Linuxrc
+      matching_keys.map{|key| install_inf_features[key]}.last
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Linuxrc.rb
```diff
@@ -211,6 +211,31 @@ def SaveInstallInf(root)
...
+    #
+    # @param [String] key
+    # @return [String] value of a given key
+    def value_for(feature_key)
+      ReadInstallInf()
+
+      feature_key = polish(feature_key)
+      install_inf_features = deep_copy(@install_inf)
+
+      # Some values written on Linuxrc commandline are not know to Linuxrc
```
lslezak: NP: `not known`

kobliha: true :) thx

library/general/src/modules/Linuxrc.rb
```diff
@@ -211,6 +211,31 @@ def SaveInstallInf(root)
...
+      install_inf_features.fetch("Cmdline", "").split.map do |cmdline_entry|
+        key_val = cmdline_entry.split("=")
+        install_inf_features[key_val[0]] = key_val[1]
+      end
+
+      matching_keys = install_inf_features.keys.select { |key| polish(key) == feature_key }
+      return nil if matching_keys.empty?
+
+      # The last key wins as in Linuxrc
+      matching_keys.map { |key| install_inf_features[key] }.last
```
lslezak: That's inefficient, you evaluate all matching keys but use just the last value. You can use the last matching key directly:

```ruby
install_inf_features[matching_keys.last]
```

library/general/src/modules/Linuxrc.rb
```diff
@@ -211,6 +211,31 @@ def SaveInstallInf(root)
...
+    # @return [String] value of a given key
+    def value_for(feature_key)
+      ReadInstallInf()
+
+      feature_key = polish(feature_key)
+      install_inf_features = deep_copy(@install_inf)
+
+      # Some values written on Linuxrc commandline are not known to Linuxrc
+      # Unless they are also mentioned as `PTOptions`, they will appear as "Cmdline" entry
+      install_inf_features.fetch("Cmdline", "").split.map do |cmdline_entry|
```
lslezak: If you do not need the result then replace `map` with `each`, this avoids creating an unused `Array`.

But the code still looks kind of ugly or weird for me, I'll try improving it a bit...

library/general/src/modules/Linuxrc.rb
```diff
@@ -224,6 +249,17 @@ def SaveInstallInf(root)
...
+    publish function: :value_for, type: "string (string)"
+
+  private
+
+    # Removes characters ignored by Linuxrc and turns all to downcase
+    #
+    # @param [String]
+    # @return [String]
+    def polish(key)
+      key.downcase.tr("-_\.", "")
```
lslezak: Backslash (`\`) is used for escaping special characters, dot (`.`) is not a special character thus the backslash is ignored:

```
irb(main):001:0> "-_\."
=> "-_."
```

Either escape the backslash (`"-_\\."`) or use single quotes (`'-_\.'`).

library/general/src/modules/Linuxrc.rb
```diff
@@ -211,6 +211,31 @@ def SaveInstallInf(root)
...
+      install_inf_features = deep_copy(@install_inf)
+
+      # Some values written on Linuxrc commandline are not known to Linuxrc
+      # Unless they are also mentioned as `PTOptions`, they will appear as "Cmdline" entry
+      install_inf_features.fetch("Cmdline", "").split.map do |cmdline_entry|
+        key_val = cmdline_entry.split("=")
+        install_inf_features[key_val[0]] = key_val[1]
+      end
+
+      matching_keys = install_inf_features.keys.select { |key| polish(key) == feature_key }
```
lslezak: Another issue is that order of `install_inf_features.keys` is not guaranteed, currently `Hash#keys` returns the keys in the insert order (http://stackoverflow.com/questions/5208512/ruby-what-is-the-order-of-keys-values-returned-by-hash-keys-and-hash-values-met) but it was different with Ruby 1.8 and might change anytime in the future....

library/general/src/modules/Linuxrc.rb
```diff
@@ -211,6 +211,31 @@ def SaveInstallInf(root)
...
+      ReadInstallInf()
+
+      feature_key = polish(feature_key)
+      install_inf_features = deep_copy(@install_inf)
+
+      # Some values written on Linuxrc commandline are not known to Linuxrc
+      # Unless they are also mentioned as `PTOptions`, they will appear as "Cmdline" entry
+      install_inf_features.fetch("Cmdline", "").split.map do |cmdline_entry|
+        key_val = cmdline_entry.split("=")
+        install_inf_features[key_val[0]] = key_val[1]
```
lslezak: This will trim the value when `=` is present, e.g. `foo=http://example.com?bar=42` will be stored as `foo` => `http://example.com?bar`.

library/general/src/modules/Linuxrc.rb
lslezak: See https://github.com/kobliha/yast-yast2/pull/1 for my proposed changes.

lslezak:
See PR https://github.com/kobliha/yast-yast2/pull/1 before commenting here...

kobliha:
@lslezak I'm not interested in the implementation, but the functionality. Both implementations have pros and cons, but the result is more important. Please approve so I can continue.

lslezak:
LGTM

PR #364, opened 2015-06-02
Backported patch for running IPv4, IPv6 check

mchf:
https://github.com/yast/yast-yast2/pull/189

bnc#932446, bnc#868001


jreidinger:
LGTM

mchf:
Thanks for review

PR #365, opened 2015-06-02
Reduced snapshotting

jreidinger:


imobach:
LGTM

package/yast2.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jun  2 11:26:50 UTC 2015 - jreidinger@suse.com
+
+- reduce count of extending snapper for snapshotting (fate#317973)
+- 3.1.128
```
aschnell: The changelog is completely misleading: You are not extending snapper. Instead you extend the instsys with snapper.

jreidinger: yeahm you are right, I change it

PR #366, opened 2015-06-02
Implemented possibility to temporarily disable creating filesystem snapshots

kobliha:
- User can define the behavior on Linuxrc commandline disable_snapshots=(single|around|all)
- Also any comma-separated combination is allowed
- Including full-coverage test case
- Thanks to Linuxrc.get_value user can define this as any other Linuxrc parameter (with dots, hyphens, underscores and case insensitive)
- Values behave the very same as the parameter name

### DIFF
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -109,13 +109,16 @@ module Yast2
       end
     end
 
-    # Creates a new 'single' snapshot
+    # Creates a new 'single' snapshot unless disabled by user
     #
     # @param description [String] Snapshot's description.
     # @return [FsSnapshot] The created snapshot.
     #
     # @see FsSnapshot.create
+    # @see FsSnapshot.create_snapshot?
     def self.create_single(description)
+      return nil unless create_snapshot?(:single)
+
       create(:single, description)
     end
 
@@ -125,11 +128,14 @@ module Yast2
     # @return [FsSnapshot] The created snapshot.
     #
     # @see FsSnapshot.create
+    # @see FsSnapshot.create_snapshot?
     def self.create_pre(description)
+      return nil unless create_snapshot?(:around)
+
       create(:pre, description)
     end
 
-    # Creates a new 'post' snapshot
+    # Creates a new 'post' snapshot unless disabled by user
     #
     # Each 'post' snapshot corresponds with a 'pre' one.
     #
@@ -138,8 +144,12 @@ module Yast2
     # @return [FsSnapshot] The created snapshot.
     #
     # @see FsSnapshot.create
+    # @see FsSnapshot.create_snapshot?
     def self.create_post(description, previous_number)
+      return nil unless create_snapshot?(:around)
+
       previous = find(previous_number)
+
       if previous
         create(:post, description, previous)
       else
@@ -148,7 +158,7 @@ module Yast2
       end
     end
 
-    # Creates a new snapshot
+    # Creates a new snapshot unless disabled by user
     #
     # It raises an exception if Snapper is not configured or if snapshot
     # creation fails.
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -98,14 +98,14 @@ module Yast2
       # Feature is not defined on Linuxrc commandline
       return true if disable_snapshots.nil? || disable_snapshots.empty?
 
-      disable_snapshots = disable_snapshots.downcase.tr("-_.", "").split(",").map(&:to_sym)
+      disable_snapshots = disable_snapshots.downcase.tr("-_.", "").split(",")
 
       if [:around, :single].include?(snapshot_type)
-        return false if disable_snapshots.include?(:all)
-        return !disable_snapshots.include?(snapshot_type)
+        return false if disable_snapshots.include?("all")
+        return !disable_snapshots.include?(snapshot_type.to_s)
       else
         raise ArgumentError, "Unsupported snapshot type #{snapshot_type.inspect}, " \
-          << "supported are :around and :single"
+              "supported are :around and :single"
       end
     end
 
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -58,6 +58,8 @@ module Yast2
   class FsSnapshot
     include Yast::Logger
 
+    Yast.import "Linuxrc"
+
     FIND_CONFIG_CMD = "/usr/bin/snapper --no-dbus --root=%{root} list-configs | grep \"^root \" >/dev/null"
     CREATE_SNAPSHOT_CMD = "/usr/lib/snapper/installation-helper --step 5 --root-prefix=%{root} --snapshot-type %{snapshot_type} --description \"%{description}\""
     LIST_SNAPSHOTS_CMD = "LANG=en_US.UTF-8 /usr/bin/snapper --no-dbus --root=%{root} list"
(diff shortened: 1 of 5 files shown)


library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -82,13 +84,41 @@ def self.configured?
...
+
+    # Creates a new 'single' snapshot unless disabled by user
     #
     # @param description [String] Snapshot's description.
     # @return [FsSnapshot] The created snapshot.
     #
     # @see FsSnapshot.create
+    # @see FsSnapshot.create_snapshot?
     def self.create_single(description)
+      return nil unless create_snapshot?(:single)
```
kobliha: No, I could not put it into `create` method, because self.create_post searches for `previous` snapshot first. So, to keep consistency with the other methods, it's three times "almost" the same code.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -82,13 +84,41 @@ def self.configured?
...
+    # Returns whether creating the given snapshot type is allowed
+    # Information is taken from Linuxrc (DISABLE_SNAPSHOTS)
+    #   * "all" - all snapshot types are temporarily disabled
+    #   * "around" - before and after calling YaST
+    #   * "single" - single snapshot at a given point
+    #
+    # @param [Symbol] one of :around (for :post and :pre snapshots) or :single
+    # @return [Boolean] if snapshot should be created
+    def self.create_snapshot?(snapshot_type)
+      disable_snapshots = Yast::Linuxrc.get_value(Yast::LinuxrcClass::DISABLE_SNAPSHOTS)
```
jreidinger: well, it is consistent with rest of linuxrc, but actually I do like it and prefer to add method like `Yast::Linuxrc.disable_snapshots`

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -82,13 +84,41 @@ def self.configured?
...
+
+      # Feature is not defined on Linuxrc commandline
+      return true if disable_snapshots.nil? || disable_snapshots.empty?
+
+      disable_snapshots = disable_snapshots.downcase.tr("-_.", "").split(",").map(&:to_sym)
```
jreidinger: I am not sure how big risk it is, but calling `to_sym` is potentially risky, as it create entry in memory that cannot be cleaned. So long malicious string can cause memory outage, but I expect attacker do not have access to cmd line.

jreidinger: another idea is that maybe such transformation can be in Linuxrc method I suggest above. so it by default return array of symbols.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -82,13 +84,41 @@ def self.configured?
...
+
+      if [:around, :single].include?(snapshot_type)
+        return false if disable_snapshots.include?(:all)
+        return !disable_snapshots.include?(snapshot_type)
+      else
+        raise ArgumentError, "Unsupported snapshot type #{snapshot_type.inspect}, " \
+          << "supported are :around and :single"
```
imobach: The `#<<` is not needed here (you could just remove it).

library/system/src/lib/yast2/fs_snapshot.rb
kobliha: Linuxrc has nothing to do with snapshots, it even doesn't know the feature. Adding just another API doesn't look like a good idea. Code should be close to other similar functionality. it might be worth refactoring, but that is out of scope of this PR, IMO.

library/system/test/fs_snapshot_test.rb
```diff
@@ -84,10 +84,12 @@ def logger
 
     before do
       allow(Yast2::FsSnapshot).to receive(:configured?).and_return(configured)
+      allow(Yast2::FsSnapshot).to receive(:create_snapshot?).and_return(create_snapshot)
```
imobach: I'd add `with(:single)` to narrow the arguments that could be passed.

jreidinger:
otherwise it looks good

library/system/test/fs_snapshot_test.rb
```diff
@@ -365,4 +401,59 @@ def logger
...
     end
   end
+
+  describe ".create_snapshot?" do
+    before do
+      Yast.import "Linuxrc"
+    end
+
+    context "when single value is defined on Linuxrc commandline" do
+      it "returns whether given snapshot is requested" do
```
imobach: I'd change 'snapshot' for 'snapshot type' and maybe 'requested' by 'allowed' (because at this point we only want to know which types are 'allowed').

library/system/test/fs_snapshot_test.rb
```diff
@@ -365,4 +401,59 @@ def logger
...
+        expect(described_class.create_snapshot?(:single)).to eq(false)
+
+        allow(Yast::Linuxrc).to receive(:get_value).with(/snapshot/).and_return("all")
+        expect(described_class.create_snapshot?(:around)).to eq(false)
+        expect(described_class.create_snapshot?(:single)).to eq(false)
+      end
+    end
+
+    context "when more values are defined on Linuxrc commandline" do
+      it "returns whether given snapshot is not withing disabled snapshots" do
```
imobach: Typo: 'withing' by 'within'?

library/system/src/lib/yast2/fs_snapshot.rb
imobach: Just curious: why do you have to substitute `-`, `_` and `.`? Sure it has something to do with Linuxrc.

jreidinger: see previous downcase. It is über cool feature which allows you to write `"All, a_round, sin_gle....around"` and it is still valid...actually linuxrc itself allows it only for keys so you can write `DUD=` or `dud=` or `autoupgrade=` or `Auto_Upgrade=`. So it allows you to write nicer words for you.

imobach: @jreidinger: oh, thanks. Anyway it sounds like we'd have to do some common method that could do that 'cleaning'.

jreidinger: yes, sounds reasonable, thats why I suggest it above, but its usage is spread over all modules, so hard to do bigger changes, maybe only add new API, but old will live with us for long time

imobach: @jreidinger If you're referring to 'symbol DoS attacks', Ruby 2.2 ships [garbage collection of symbols](http://www.sitepoint.com/symbol-gc-ruby-2-2/).

jreidinger: @imobach Cool, I do not know it. Thanks for info

library/system/test/fs_snapshot_test.rb
kobliha: OK, why not

library/system/src/lib/yast2/fs_snapshot.rb
kobliha: Frankly, these symbols are used just because the same (previous implementation) symbols were used for different snapshot types.

You are right, that the symbol stays in memory, but I could not imagine anyone, who has access to Linuxrc comandline, who would like to spoil their own memory by super-long string. They already have root, so there's no reason for that. I'll check whether some [String] comparison would be better though.

library/system/test/fs_snapshot_test.rb
kobliha: Yes

library/system/test/fs_snapshot_test.rb
kobliha: Hmm, it seems I was working on the test case before changing the whole method name.

kobliha:
Wishes and requests were added / implemented, LGTMed already, so merging...

PR #367, opened 2015-06-04
Fix value_for typo

imobach:
Just call `value_for` instead of `get_value`.


jreidinger:
LGTM

jreidinger:
BTW rspec 3 provide nice feature how to avoid such problems: http://stackoverflow.com/questions/21426216/forbid-mocking-methods-that-are-undefined-on-original-object

imobach:
@jreidinger Nice. I'll take that into account for future tests. Thanks.

PR #368, opened 2015-06-11
Dropped no longer used group from YCC

mchf:
follow up for https://github.com/yast/yast-network/pull/313


mvidner:
LGTM

mchf:
Rebased on top of current master, going to merge once Travis is done.

Thanks for review.

PR #369, opened 2015-06-11
Added Yast::Directory#find_data_file

ancorgs:


library/general/src/modules/Directory.rb
```diff
@@ -117,13 +118,24 @@ def Directory
       nil
     end
 
+    # Find an existing file in any of the YaST data directories
+    #
+    # @param relative_path [String] name of the file to find
+    # @return [String] full path of the file if it's found, nil otherwise
+    def find_data_file(relative_path)
+      possible_paths = Yast.y2paths.map { |p| File.join(p, "data", relative_path) }
+      possible_paths.find { |p| File.exist?(p) }
```
jreidinger: Question is if we should use absolute `File.exist?` or something that respect SCR like `FileUtils.Exist`. I am not sure as you maybe want both...maybe option with some reasonable default?

library/general/src/modules/Directory.rb
```diff
@@ -117,13 +118,24 @@ def Directory
...
+
     publish variable: :bindir, type: "string"
     publish variable: :ybindir, type: "string"
     publish variable: :logdir, type: "string"
     publish variable: :vardir, type: "string"
     publish variable: :etcdir, type: "string"
     publish variable: :agentdir, type: "string"
+    # @deprecated since it does not honor Y2DIR. Use #find_data_file instead
     publish variable: :datadir, type: "string"
+    # @deprecated since it's an alias for the deprecated #datadir
```
jreidinger: do we want to publish new method to be usable also from perl? I am fine if not and add it in future if needed.

library/general/src/modules/Directory.rb
ancorgs: Actually I was wondering the same.

library/general/src/modules/Directory.rb
ancorgs: I wouldn't publish it until is needed.

test/test_helper.rb
```diff
@@ -1,5 +1,7 @@
 root_location = File.expand_path("../../", __FILE__)
 inc_dirs = Dir.glob("#{root_location}/library/*/src")
+# Needed to test Y2DIR support in Yast::Directory
+inc_dirs << "#{root_location}/library/general/test"
```
jreidinger: to be honest I do not like it...better add support to respect previous Y2DIR option like 
```
  if ENV["Y2DIR"]
    ENV["Y2DIR"] << ":"
  else
    ENV["Y2DIR"] = ""
  end
  ENV["Y2DIR"] << inc_dirs.join(":")
```

and set Y2DIR in respective test, because if more tests need own testing location if can quite quickly become horrible mess

library/general/src/modules/Directory.rb
ancorgs: Actually, reading a file from the data directory is conceptually closer to requiring a ruby file than to doing operations with the agents. I would not expect the SCR chrooting to affect anything else than the calls to SCR.

test/test_helper.rb
ancorgs: Yes, that was my first attempt and it works if you do ```rspec directory_test.rb```. But it does not work with ```rake test:unit```, since all the tests are called in a single command so you cannot set Y2DIR in the tests in a reliable way.

library/general/src/modules/Directory.rb
jreidinger: agreed, we can extend it later as it will be api compatible change to add something like `def find_data_file(relative_path, chroot: false)`

test/test_helper.rb
jreidinger: ah, thats right :(
ok, so please at least use constant with explanation and use it like
```
# Y2DIRs location needed for testing purpose
ADDITIONAL_Y2DIRS = [
  # Needed to test Y2DIR support in Yast::Directory
  "#{root_location}/library/general/test"
]
```

ancorgs:
@jreidinger Updated with the suggested constant and command

jreidinger:
LGTM

PR #370, opened 2015-06-13
Changed default TLD

mchf:
bnc#922765

hotfix for SLE-11-SP4


mchf:
Only moved domain into the constant. Intention is that all places where the TLD is needed should depend on this module ```Hostname``` and the constant.

jreidinger:
LGTM

mchf:
thanks for review

PR #375, opened 2015-06-25
Starting the Installer (YaST2.First-Stage) directly from yast2

kobliha:
- FATE#317637, bnc#877447
- Explanation and discussion at http://lists.opensuse.org/yast-devel/2015-06/msg00024.html
- Tested IRL with dud=${path_to}/yast2.rpm - the package is downloaded, used, new message "Starting Installer" is shown and then Installation starts as usually

### DIFF
--- a/scripts/yast2
+++ b/scripts/yast2
@@ -14,10 +14,11 @@
 # the configuration and administration menu.
 
 # FATE#317637, bsc#877447: In installation system, we call the installer
-# program directly
+# script directly and then exit
 if [ -f /.instsys.config ]; then
   echo "Starting Installer"
-  exit `/usr/lib/YaST2/startup/YaST2.First-Stage`
+  /usr/lib/YaST2/startup/YaST2.First-Stage
+  exit $?
 fi
 
 export PATH=/sbin:/usr/sbin:$PATH
--- a/scripts/yast2
+++ b/scripts/yast2
@@ -13,6 +13,13 @@
 # wise ncurses. It starts then the module 'menu' which implements
 # the configuration and administration menu.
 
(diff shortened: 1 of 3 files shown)


kobliha:
@shundhammer Thanks :) I'll test this IRL tomorrow and merge if that works well

scripts/yast2
```diff
@@ -13,6 +13,14 @@
 # wise ncurses. It starts then the module 'menu' which implements
 # the configuration and administration menu.
 
+# FATE#317637, bsc#877447: In installation system, we call the installer
+# program directly
+if [ -f /.instsys.config ]; then
+  echo "Starting Installer"
+  exec /usr/lib/YaST2/startup/YaST2.First-Stage
+  exit $?
```
mvidner: BTW this `exit` will only be run if the `exec ` failed, but that is actually what we want :)

mvidner:
You really want to overflow people's branch name buffer, don't you?

scripts/yast2
```diff
@@ -13,6 +13,14 @@
 # wise ncurses. It starts then the module 'menu' which implements
 # the configuration and administration menu.
 
+# FATE#317637, bsc#877447: In installation system, we call the installer
+# script directly and then exit
+if [ -f /.instsys.config ]; then
+  echo "Starting Installer"
+  /usr/lib/YaST2/startup/YaST2.First-Stage
+  exit $?
```
mvidner: This is simple and works as intended, since we don't want to get the control center if executing the installation failed.

mvidner:
LGTM.

kobliha:
Tested IRL and worked well, see the updated PR description.

PR #378, opened 2015-07-07
Huha snapshot cleanup

shundhammer:
https://trello.com/c/YDDWzV5X

### DIFF
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -65,6 +65,9 @@ module Yast2
     LIST_SNAPSHOTS_CMD = "LANG=en_US.UTF-8 /usr/bin/snapper --no-dbus --root=%{root} list"
     VALID_LINE_REGEX = /\A\w+\s+\| \d+/
 
+    # Predefined snapshot cleanup strategies (the user can define custom ones, too)
+    CLEANUP_STRATEGY = { number: "number", timeline: "timeline" }
+
     attr_reader :number, :snapshot_type, :previous_number, :timestamp, :user,
       :cleanup_algo, :description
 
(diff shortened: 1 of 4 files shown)


library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -65,6 +65,9 @@ class FsSnapshot
     LIST_SNAPSHOTS_CMD = "LANG=en_US.UTF-8 /usr/bin/snapper --no-dbus --root=%{root} list"
     VALID_LINE_REGEX = /\A\w+\s+\| \d+/
 
+    # Predefined snapshot cleanup strategies (the user can define custom ones, too)
```
jreidinger: I do not see how user can use his own defined one

shundhammer: Not in the context of this Ruby class, but in general, he can. Snapper comes with those predefined ones you see here, but a user can write his own scripts and make snapper use them on request. Arvin knows more details about that.

library/system/test/fs_snapshot_test.rb
```diff
@@ -118,6 +121,46 @@ def logger
           expect(snapshot).to be(dummy_snapshot)
         end
       end
+
+      context "when a cleanup strategy is set" do
+        let(:output) { { "stdout" => "2", "exit" => 0 } }
+        let(:dummy_snapshot) { double("snapshot") }
+        let(:snapshot_command) { CREATE_SINGLE_SNAPSHOT + OPTION_CLEANUP_NUMBER }
+
+        it "returns the created snapshot" do
```
jreidinger: when you run `rspec --format doc` it sounds bit strange, that all when have same result. More important here is to mention that strategy is used like 
```
it "creates snapshot with given strategy and returns the created snapshot" do
 ...
end
```
and similar below, so it is obvious how cleanup strategy affect behavior.

library/system/src/lib/yast2/fs_snapshot.rb
jreidinger: Ah a bit confusing as I think it mean, that you can pass to parameter also other strategies, which do not work.

PR #380, opened 2015-07-13
Agent infrastructure

mvidner:
https://trello.com/c/MMlaM6sK/6-better-agent-infrastructure

### DIFF
--- a/library/general/src/lib/config_file/augeas_parser.rb
+++ b/library/general/src/lib/config_file/augeas_parser.rb
@@ -1,8 +1,62 @@
 require "augeas"
 
 module ConfigFile
-  # TODO: figure out a way to bind to Augeas, and to replace ConfigFile::KeyValueParser with it
+  # TODO: handle nested structures
+  # TODO: use for data smarted structure that can search
+  # @example read, print, modify and serialize again
+  #
+  #  parser = ConfigFile::AugeasParser.new("sysconfig.lns")
+  #  data = parser.parse(File.read("/etc/default/grub"))
+  #  puts data
+  #  data.find { |d| d[:key] == "GRUB_DISABLE_OS_PROBER" }[:value] = "true"
+  #  puts parser.serialize(data)
   class AugeasParser
+    def initialize(lense)
+      @lense = lense
+    end
 
+    def parse(raw_string)
+      @old_content = raw_string
+
+      Augeas::open(nil, nil, Augeas::NO_MODL_AUTOLOAD) do |aug|
+        aug.set("/input", raw_string)
+        if !aug.text_store(@lense, "/input", "/store")
+          error = aug.error
+          raise "Augeas error #{error[:message]}. Details: #{error[:details]}."
+        end
+        matches = aug.match("/store/*")
+
+        return matches.map do |aug_key|
+          key = aug_key.sub(/^\/store\//, "")
+          key.sub!(/\[\d+\]/, "[]")
+          {
+            key:   key,
+            value: aug.get(aug_key)
+          }
+        end
+      end
+    end
+
+    def serialize(data)
+      arrays = {}
+
+      Augeas::open(nil, nil, Augeas::NO_MODL_AUTOLOAD) do |aug|
+        aug.set("/input", @old_content || "")
+        data.each do |entry|
+          key = entry[:key]
+          if key.end_with?("[]")
+            array_key = key.sub(/\[\]$/, "")
+            arrays[array_key] ||= 0
+            arrays[array_key] += 1
+            key = array_key + "[#{arrays[array_key]}]"
+          end
+          aug.set("/store/#{key}", entry[:value])
+        end
+
+        aug.text_retrieve(@lense, "/input", "/store", "/output")
+
+        return aug.get("/output")
+      end
+    end
   end
 end
--- a/library/general/src/lib/config_file/augeas_parser.rb
+++ b/library/general/src/lib/config_file/augeas_parser.rb
@@ -1,7 +1,8 @@
 require "augeas"
 
-module ConfigAgent
+module ConfigFile
+  # TODO: figure out a way to bind to Augeas, and to replace ConfigFile::KeyValueParser with it
   class AugeasParser
-    
+
   end
 end
--- a/library/general/src/lib/config_file/augeas_parser.rb
+++ b/library/general/src/lib/config_file/augeas_parser.rb
@@ -1,59 +1,106 @@
 require "augeas"
 
 module ConfigFile
+
+  # Represent parsed augeas config tree with user friendly methods
+  class AugeasTree
+    def initialize
+      @data = {}
+    end
+
+    def [](key)
+      entry = @data.find{|d| d[:key] == key}
+      return nil unless entry
+
+      entry[:value]
+    end
+
+    def []=(key, value)
+      entry = @data.find{|d| d[:key] == key}
+      if entry
+        entry[:value] = value
+      else
+        @data << {
+          key:   key,
+          value: value
+        }
+      end
+    end
+
+    def load_from_augeas(aug, prefix)
+      matches = aug.match("#{prefix}/*")
+
+      @data = matches.map do |aug_key|
+        key = aug_key.sub(/^#{prefix}\//, "")
+        key.sub!(/\[\d+\]/, "[]")
+        {
+          key:   key,
+          value: aug.get(aug_key)
+        }
+      end
+    end
+
+    def save_to_augeas(aug, prefix)
+      arrays = {}
+
+      @data.each do |entry|
+        key = entry[:key]
+        if key.end_with?("[]")
+          array_key = key.sub(/\[\]$/, "")
+          arrays[array_key] ||= 0
+          arrays[array_key] += 1
+          key = array_key + "[#{arrays[array_key]}]"
+        end
+        aug.set("#{prefix}/#{key}", entry[:value])
+      end
+    end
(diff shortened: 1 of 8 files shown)


library/general/src/lib/config_file/augeas_parser.rb
```diff
@@ -0,0 +1,7 @@
+require "augeas"
+
+module ConfigAgent
+  class AugeasParser
```
mvidner: > \# TODO: figure out a way to bind to Augeas, and to replace ConfigFile::KeyValueParser with it

library/general/src/lib/config_file/augeas_parser.rb
```diff
@@ -0,0 +1,7 @@
+require "augeas"
+
+module ConfigAgent
```
mvidner: s/ConfigAgent/ConfigFile/ ?

library/general/src/lib/config_file/config_file.rb
```diff
@@ -0,0 +1,29 @@
+module ConfigFile
+  class ConfigFile
```
mvidner: About naming:
- `ConfigFile` for the whole module sounds good.
- `ConfigFile::ConfigFile` -> `ConfigFile::ModelBase`
- `GrubConfig` -> `GrubModel`
- `KeyValue_Parser` -> `KeyValueParser`
- `AugeasParser` sounds good

library/general/src/lib/config_file/grub_config.rb
```diff
@@ -0,0 +1,47 @@
+require "config_file/config_file"
+require "config_file/keyvalue_parser"
+
+module ConfigFile
+  PARSER = Keyvalue_Parser
+  PATH = "/etc/default/grub"
+
+  class GrubConfig < ConfigFile
```
mvidner: `GrubModel`

library/general/src/lib/config_file/keyvalue_parser.rb
```diff
@@ -0,0 +1,37 @@
+module ConfigFile
+  class Keyvalue_Parser
```
mvidner: It's a Camelo_Snake! `KeyValueParser`

library/general/src/lib/config_file/config_file.rb
```diff
@@ -0,0 +1,29 @@
...
+    def initialize(parser, file_path, file_class: File)
+      @file_class = file_class
+      @parser = parser
+      @file_path = file_path
+    end
+
+    def write(changes_only: false)
```
mvidner: So `read`, `write` are part of the public API... Wouldn't it be better to adopt an ActiveModel API, with `save` instead?

library/general/src/lib/config_file/config_file.rb
mvidner: > \# A base class for models. Represents a configuration file as an object with domain-specific attributes/methods.
Uses
- a **file** (usually plain `::File`) to convert between the on-disk storage and a `String`,
- a **parser** (usually specific for each derived Model) to convert between that `String` and a private representation

library/general/src/lib/config_file/keyvalue_parser.rb
mvidner: > \# A `ConfigFile::ParserBase` can `parse` and `serialize` some data between a String and some structured representation of that string. According to the type of that representation it is coupled with a class derived from `ConfigFile::ModelBase`

library/general/src/lib/config_file/config_file.rb
mvidner: What is our position on caching and flushing? (compare the explicit flushes required with some SCR agents, (not) re-reading the config on repeated Read)

library/general/src/lib/config_file/augeas_parser.rb
jreidinger: yeah, typo, original name and then I recognize I do not like agent term, when it is in fact configuration file.

library/general/src/lib/config_file/grub_config.rb
```diff
@@ -0,0 +1,47 @@
...
+      data["GRUB_CMDLINE_LINUX_RECOVERY"][:commented_out] = true
+    end
+
+    def enable_recovery_entry(kernel_params)
+      data["GRUB_DISABLE_RECOVERY"] ||= {}
+      data["GRUB_DISABLE_RECOVERY"][:value] = "true"
+      data["GRUB_DISABLE_RECOVERY"][:commented_out] = false
+      data["GRUB_CMDLINE_LINUX_RECOVERY"] ||= {}
+      data["GRUB_CMDLINE_LINUX_RECOVERY"][:value] = kernel_params
+      data["GRUB_CMDLINE_LINUX_RECOVERY"][:commented_out] = false
```
mvidner: For the record, this part of the model is repetitive and would get refactored for production use, but it isn't our focus now.

library/general/src/lib/config_file/config_file.rb
jreidinger: I do not like `GrubModel` as it sounds like model for whole grub, which is not true...so maybe `GrubConfModel`?

library/general/src/lib/config_file/config_file.rb
```diff
@@ -0,0 +1,29 @@
...
+      merge_changes if changes_only
```
mvidner: The merging feature sounds... interesting and complicated. Too early to say more.

library/general/src/lib/config_file/keyvalue_parser.rb
jreidinger: yeah, creating name is the biggest challenge :wink:

library/general/src/lib/config_file/config_file.rb
jreidinger: whatever agent needs. In general common base support read/write, no caching, only parsed tree is stored in memory, so you can call more writes...if you need you can call more reads to reset.

mvidner:
Overall it looks good. I think we should work on the external interface before wrestling with the augeas internals: see how it actually replaces existing SCR code, for more than one model.

library/general/src/lib/config_file/config_file.rb
mvidner: In general, a matching FooModel and FooParser. Now, does `GrubConfModel` describe /etc/grub.conf? `EtcDefaultGrubModel`? But that's not important for the general API.

mvidner: > \# A base class for models. Represents a configuration file as an object with domain-specific attributes/methods. For persistent storage, use `load` and `save`,
Non-responsibilities: actual storage and parsing (both delegated).

library/general/src/lib/config_file/config_file.rb
mvidner: \# No extra caching (unlike some SCR agents)

library/general/src/lib/config_file/config_file.rb
```diff
@@ -0,0 +1,29 @@
...
+      @file_class.write(@file_path, @parser.serialize(data))
+    end
+
+    def read
```
mvidner: \# No extra caching (unlike some SCR agents)

library/general/src/lib/config_file/config_file.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  protected
+
+    attr_accessor :data
+
+    def merge_changes
+      new_data = data.dup
+      read
+      # TODO recursive merge
+      date.merge(new_data)
```
mvidner: data

jreidinger:
testing script:
```
jreidinger@linux:~/yast/yast2/library/general/src/lib/config_file> cat ~/test.rb 
$: << "/home/jreidinger/yast/yast2/library/general/src/lib"

require "config_file/grub_model"
require "config_file/memory_file"

config = ConfigFile::GrubModel.new
config.load

puts"config: " + config.inspect
puts ""
puts "os prober:  #{config.os_prober_enabled?}"

memory_file = ConfigFile::MemoryFile.new
config2 = ConfigFile::GrubModel.new(file_class:  memory_file)

config2.disable_os_prober
config2.enable_recovery_entry "\"kernel_do_your_job=HARD!\""

config2.save

puts
puts "Testing output:"
puts memory_file.content

```

library/general/src/lib/config_file/memory_file.rb
```diff
@@ -0,0 +1,17 @@
+module ConfigFile
+  class MemoryFile
+    attr_accessor :content
+
+    def initialize(content = "")
+      @content = content
+    end
+
+    def read(path)
```
mvidner: This being an instance method means you have to pass an instance to a parameter named "class": 
```ruby
memory_file = ConfigFile::MemoryFile.new
config2 = ConfigFile::GrubModel.new(file_class:  memory_file)
```

That's not nice, but I do not see a better option :(

jreidinger: maybe just naming issue. I plan to use it this way from beginning.

mvidner:
TODO: try and describe how we handle (un)escaping of values.

mvidner:
OK, so value escaping is probably not necessary with respect to Augeas core, and depends on each individual lens.
But Augeas core gets into play for keys: what if a key contains a slash or a space or a bracket. See https://bugzilla.suse.com/show_bug.cgi?id=925225

jreidinger:
lets close it as now it lives in separate gem at https://github.com/jreidinger/config_files

PR #381, opened 2015-07-14
Fail tests if mocked method does not exist

mvidner:
WIP

If you misremember a method name both in code and in tests,
will save you.
https://relishapp.com/rspec/rspec-mocks/v/3-0/docs/verifying-doubles/partial-doubles


jreidinger:
looks good, I would like to see it in yast/rspec as default

jreidinger:
LGTM, general enablement is not possible now as it can break build of too much modules.

mvidner:
The new Travis test ran with RSpec 3; and an OBS build has succeeded too.

PR #382, opened 2015-07-17
Remove references to snapper extension

imobach:
* Fixes bsc#938377.

### DIFF
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -77,11 +77,10 @@ module Yast2
     def self.configured?
       return @configured unless @configured.nil?
 
-      out = with_snapper do
-        Yast::SCR.Execute(Yast::Path.new(".target.bash_output"),
-          format(FIND_CONFIG_CMD, root: target_root)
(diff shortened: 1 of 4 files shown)


ancorgs:
:+1:

imobach:
Thanks!

jreidinger:
so if I get it snapper is already part of insts-sys and not extension only, right?

imobach:
@jreidinger Yes. The problem is that now we need to ship a new version of snapper because the current one does not have support for setting the clean-up algorithm nor the 'important' flag through the installation helpers.

jreidinger:
it should be auto added when accepted to SP1 or TW, not?

imobach:
@jreidinger I think that snapper releasing is a little bit different, isn't it? I'll take a look at what you wrote in the `README`.

@aschnell added the `--cleanup` option to installation helpers in https://github.com/openSUSE/snapper/pull/171 but we haven't released a new version since then, so the one included in the DVD is not able to do that.

imobach:
I've take a look at OBS at it seems that they is a pending request (since 10 days ago). I'll investigate it further. Thanks!

PR #383, opened 2015-07-20
Fix create pre fs snapshot

imobach:
`Yast2::FsSnapshot.create_pre`, as used in https://github.com/yast/yast-update/blob/7aea14b5c5a8632f8bd86855df245ba16b90902f/src/modules/RootPart.rb#L1808, is supposed to receive `cleanup` and `important` parameters (instead of only the description). This commit should fix this issue.


jreidinger:
LGTM

imobach:
Thanks!

PR #384, opened 2015-07-21
Support alternative input

jreidinger:


library/general/src/lib/ui/event_dispatcher.rb
```diff
@@ -46,6 +46,44 @@ def event_loop
       end
     end
 
+    # Methods gets input for event dispatching
+    # Its main purpose is to allow change user input obtaining method like
+    # using user input with timeout.
+    # @example use user input with timeout
```
ancorgs: I like the code. But this description is a little bit confusing. Even having a whole example makes things look more complicated that they really are. I'd go for something like

```ruby
# Reads input for next event dispatching
# Redefine to use an alternative method to get user input, like introducing a timeout.
# Default implementation uses Yast::UI.UserInput which waits indefinitely for user input.
```

Either keeping the example or even removing it. I don't have strong feelings about that point.

ancorgs:
Other than the small typo (actually my fault), LGTM

PR #386, opened 2015-07-28
Fix handling of error reporting at Commandline.rb

ancorgs:


library/commandline/src/modules/CommandLine.rb
```diff
@@ -1491,45 +1491,13 @@ def UniqueOption(options, unique_options)
...
-        Ops.greater_than(Report.NumErrors, 0),
-        Ops.greater_than(Report.NumMessages, 0),
-        Ops.greater_than(Report.NumYesNoMessages, 0)
-      )
-      if Ops.greater_than(Builtins.size(report), 0)
-        Yast.import "RichText"
-        Print(RichText.Rich2Plain(report))
+    def run_if_exists(commandline, index)
+      meth = commandline[index]
+      if meth
```
lslezak: NP: maybe this one liner is better `meth ? meth.call : false`

package/yast2.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Tue Jul 28 10:50:05 UTC 2015 - ancor@suse.com
+
+- Properly display error information when running in command line
```
mvidner: "Error" can mean many things. AFAIK the fix concerns only exceptions, so we should say 
> Do not hide exceptions when running in command line mode (boo#939565).

 right?

library/commandline/test/clients/dummy_cmdline.rb
```diff
@@ -0,0 +1,68 @@
...
+      puts "Initialize called"
+      true
+    end
+
+    def echo_handler(options)
+      puts options["text"]
+      true
+    end
+
+    def crash_handler(*)
```
lslezak: There is a small inconsistency, here you allow any params (via `*`), but the `funref()` above contains `boolean ()` signature.

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1491,45 +1491,13 @@ def UniqueOption(options, unique_options)
...
-      report = Report.GetMessages(
-        Ops.greater_than(Report.NumWarnings, 0),
-        Ops.greater_than(Report.NumErrors, 0),
-        Ops.greater_than(Report.NumMessages, 0),
-        Ops.greater_than(Report.NumYesNoMessages, 0)
-      )
-      if Ops.greater_than(Builtins.size(report), 0)
-        Yast.import "RichText"
-        Print(RichText.Rich2Plain(report))
+    def run_if_exists(commandline, index)
```
mvidner: This entire method is unnecessary :) because wherever it is called the caller has already checked that commandline[index] exists. I guess the `fake_false` logic was meant just to fit into the YCP type system.

package/yast2.changes
ancorgs: Well, actually it hides everything reported by means of `Report.Error`. That includes exceptions (in the current implementation of WFM) but also more things. Other than that, I find "error" to be a more end-user friendly word than "exception".

mvidner: Yes, *error* is more general but my point is that the ones via Report.Error *were* reported, all at once in the end. The problem is just the exceptions, since they escaped this collector and went via the still disabled Report.Error.

library/commandline/test/clients/dummy_cmdline.rb
ancorgs: Even with the original implementation of CommandLine.rb, if I remove the `(*)` the test fails. So I probably misunderstood something about how ComandLine and/or funref works. I'll take a closer look.

ancorgs: Well, actually the signature of funref is completely ignored by `CommandLine` and assumed to be `boolean (map <string, string>)` no matter what. So I will correct the dummy client, even if it's just kind of cosmetic. :-)

mvidner:
LGTM

lslezak:
LGTM

PR #387, opened 2015-08-03
Prevent premature call to SetApplicationIcon (bnc#940341)

ancorgs:


library/wizard/src/modules/Wizard.rb
```diff
@@ -1192,9 +1166,8 @@ def SetDesktopIcon(file)
 
       return false unless icon
 
-      icon_path = File.join(@icon_path, icon + ".png")
```
mvidner: So this @icon_path was never defined and wouldn't work. Nice catch.

library/wizard/src/modules/Wizard.rb
```diff
@@ -1881,6 +1851,24 @@ def SetProductName(name)
...
+    # arguments
+    def open_wizard_dialog(*args)
+      set_icon
+      UI.OpenDialog(
+        Opt(:wizardDialog),
+        Wizard(*args)
+      )
+    end
+
+    # Sets the application icon according to the value of @icon_name
```
mvidner: "This should be called only before an actual dialog is opened; premature UI calls can interfere with the CommandLine mode"

ancorgs: Good point

mvidner:
LGTM.
Last time Rubocop complained about the indentation of `protected` BTW.

PR #388, opened 2015-08-10
Keep routing state when disabling/stopping firewall

mchf:
obsoletes https://github.com/yast/yast-yast2/pull/371

minor fix + changelog + rebased


kobliha:
LGTM

PR #391, opened 2015-08-19
Read fqdn from /etc/hostname when needed

mchf:
bnc#941881


library/types/src/modules/Hostname.rb
```diff
@@ -159,7 +160,7 @@ def CurrentFQ
         SCR.Execute(path(".target.bash_output"), "hostname --fqdn")
       )
       if hostname_data.nil? || Ops.get_integer(hostname_data, "exit", -1) != 0
-        fqhostname = if SCR.Read(path(".target.stat"), "/etc/hostname").empty?
+        fqhostname = if FileUtils.Exists("/etc/hostname")
```
ancorgs: Isn't this implementable using plain ruby? We want to get rid of Yast::Fileutils at some point in future.

jreidinger:
LGTM

library/types/src/modules/Hostname.rb
jreidinger: I think it is problem during upgrade, where you want to read hostname from target machine

library/types/test/hostname_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    allow(Yast::SCR)
+      .to receive(:Read)
+      .with(path(".target.string"), "/etc/hostname")
+      .and_return(TEST_HOSTNAME)
+    allow(Yast::FileUtils)
+      .to receive(:Exists)
+      .with("/etc/hostname")
+      .and_return(true)
+
+    expect(Yast::Hostname.CurrentFQ).to eq TEST_HOSTNAME
```
ancorgs: For the shake of completeness, what about adding also a test for a successfully `hostname --fqdn` (that will prove, OTOH that /etc/hostname is not directly read in that case).

library/types/test/hostname_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    allow(Yast::SCR)
+      .to receive(:Read)
+      .with(path(".target.string"), "/etc/hostname")
+      .and_return(test_hostname)
+    allow(Yast::FileUtils)
+      .to receive(:Exists)
+      .with("/etc/hostname")
+      .and_return(true)
+
+    expect(Yast::Hostname.CurrentFQ).to eq test_hostname
```
ancorgs: Better, thanks.

Although being NP, I would prefer to have two different names for each case, to make more clear they are coming from different places. Something like thi
```ruby
let(:command_hostname) { "host.suse.cz" }
let(:etc_hostname) { "host.suse.de" }
```
And then
```ruby
expect(...).to eq command_hostname
...
expect(...).to eq etc_hostname
```

As said. Just NP. I can leave with the current implementation.

ancorgs:
LGTM

mchf:
Thanks for reviews.

PR #392, opened 2015-08-20
"yast2" script - handle also the reboot flag (bsc#942461)

lslezak:
...the same way as the installation script

### Notes ###

- The installation script handles the reboot this way: https://github.com/yast/yast-installation/blob/master/startup/YaST2.call#L368, in installed system the support is missing
- No version change (yet), it's needed for online migration, it will be released as part of the "online migration" update later (or sooner if some other urgent issue appears until that)


imobach:
LGTM

PR #394, opened 2015-08-26
[Review] Request from 'schubi2' @ 'yast/yast-yast2/review_150826_checking_cpuinfo_flags_correctly'

schubi2:
  * efa8443 packaging
  * 027b583 checking cpuinfo_flags correctly

### DIFF
--- a/library/system/src/modules/Kernel.rb
+++ b/library/system/src/modules/Kernel.rb
@@ -362,7 +362,7 @@ module Yast
         cpuflags = []
 
         # bugzilla #303842
-        if !cpuflags.nil?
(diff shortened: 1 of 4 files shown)


library/system/src/modules/Kernel.rb
```diff
@@ -362,7 +362,7 @@ def ProbeKernel
         cpuflags = []
 
         # bugzilla #303842
-        if !cpuflags.nil?
+        if !cpuinfo_flags.nil?
```
jreidinger: better to use simple `if cpuinfo_flags`

jreidinger:
in general LGTM

PR #397, opened 2015-08-26
Firewall keeps imported data

mchf:
bnc#897129

alternative approach to https://github.com/yast/yast-network/pull/275

notes, see above PR for details I refer to:
- the issue is that when SuSEFirewall#Import is followed by SuSEFirewall#Read than the later overwrites imported data
- keep_install_network should address only network setup (setup which can be done by yast2-network or <networking /> section in case of AY profile; just my POV)
- keep_install_network can be a bit misleading in above PR. Its semantic is that network configuration used in linuxrc should be transferred to target. In case of firewall it makes no sense - linuxrc do not allow to setup firewall and in the above PR it is used to force AY setup.
- in my POV, SuSEFirewall should take care of protecting imported data.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Aug 26 19:25:07 UTC 2015 - mfilka@suse.com
+
+- bnc#897129
+  - AutoYaST will no longer ignore firewall settings if
+    keep_install_network is enabled
+- 3.1.108.8
+
+-------------------------------------------------------------------
 Wed Aug 26 15:27:51 CEST 2015 - schubi@suse.de
 
 - Checking cpuinfo_flags correctly while evaluating kernel packages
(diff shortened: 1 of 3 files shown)


schubi2:
LGTM. Please drop the other one.

package/yast2.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug 26 19:25:07 UTC 2015 - mfilka@suse.com
+
+- bnc#897129
+  - protect imported data agains overwriting by read
```
mvidner: This description is too implementation oriented. We should rephrase it from the user's POV.

mvidner: How about
> AutoYaST will no longer ignore firewall settings if keep_install_network is enabled.

mvidner:
Except for the changelog it looks good, thank you!

package/yast2.changes
mchf: interaction with user is not my cup of tea ;-) I'll rephrase it as suggested. Thanks

mchf:
Thanks for reviews

PR #399, opened 2015-09-11
Disable snapshots when running the "migration" module (bsc#944019)

lslezak:
The migration module creates the snapshots explicitly, avoid too many
snapshots created during the online migration.

https://bugzilla.suse.com/show_bug.cgi?id=944019

### DIFF
--- a/scripts/yast2
+++ b/scripts/yast2
@@ -182,7 +182,7 @@ SNAPPERBIN=/usr/bin/snapper
 
 snapshot_pre()
 {
-    if [ "$1" != "menu" ] ; then
+    if [ "$1" != "menu" -a "$1" != "migration" ]; then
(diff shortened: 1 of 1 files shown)


ancorgs:
Well, I can leave with it as it is, so LGTM.

On the other hand, I'd prefer to see something like
```
if wanted_snapshots_for_module $1; then
```
with a better name for the function, of course :-)

lslezak:
IMHO for two modules it is OK, if would need to add more later it would make sense to replace it by a function.

mvidner:
@lslezak you haven’t bumped the version on this one, so the package 3.1.108.9 is not submitted, (see https://build.suse.de/package/show/SUSE:SLE-12:Update/yast2 ). Is that OK?

PR #401, opened 2015-10-02
Move SrvStatusComponent from dns-server to yast2

ancorgs:
Goes with https://github.com/yast/yast-dns-server/pull/51

Remember is moving of files from one module to another, so please don't re-review them for code correctness unless is related to the moving.


ancorgs:
Umm, Rubocop in yast2 is more demanding than in dns-server. I'll check

library/general/src/lib/ui/srv_status_component.rb
```diff
@@ -0,0 +1,236 @@
...
+          reload_widget,
+          VSpacing(),
+          on_boot_widget
+        )
+      )
+    end
+
+    # Handles the input triggered by the widgets, this method must be called in
+    # the event loop of the dialog using the component.
+    def handle_input(input)
```
jreidinger: do you have way how to use it in new dialogs, that is based on Dialog class?

library/general/src/lib/ui/srv_status_component.rb
```diff
@@ -0,0 +1,236 @@
...
+        stop_service
+        refresh_widget
+      when "#{id_prefix}_start"
+        start_service
+        refresh_widget
+      when "#{id_prefix}_reload"
+        @reload = Yast::UI.QueryWidget(Id(input), :Value)
+      when "#{id_prefix}_enabled"
+        @enabled = Yast::UI.QueryWidget(Id(input), :Value)
+        @enabled_callback.call(@enabled) if @enabled_callback
```
jreidinger: If there is Sandi Metz she will say you how she hates ifs :)...what is possible to do it to use EmptyCallback instead of nil in default constructor.

jreidinger: or `Proc.new`

library/general/src/lib/ui/srv_status_component.rb
```diff
@@ -0,0 +1,236 @@
...
+  protected
+
+    attr_reader :id_prefix
+
+    # Checks if the service is currently running
+    #
+    # Must be redefined for services not following standard procedures
+    #
+    # @return [Boolean]
+    def service_running?
```
jreidinger: that exactly show why Service API is not nice...this method should not be needed and instead it should call `service.running?`

library/general/src/lib/ui/srv_status_component.rb
```diff
@@ -0,0 +1,236 @@
...
+      Yast::Service.Stop(@service_name)
+    end
+
+    # Reloads the configuration of a running service
+    #
+    # Must be redefined for services not following standard procedures
+    def reload_service
+      log.info "Default implementation of SrvStatusComponent#reload_service for #{@service_name}"
+      Yast::Service.Reload(@service_name)
+    end
```
jreidinger: again here I expect usage of `@service` object, so you do not need to use redefinition for non-standard services :(

library/general/src/lib/ui/srv_status_component.rb
```diff
@@ -0,0 +1,236 @@
...
+        ]
+      else
+        [
+          # TRANSLATORS: status of a service
+          Label(_("stopped")),
+          Label(" "),
+          PushButton(Id("#{id_prefix}_start"), _("Start now"))
+        ]
+      end
+    end
```
jreidinger: so basically I see there mixture of two functionalities - widget one ( good ) and service wrappers ( wrong ). I think we should consider using some nicer API for services.

library/general/test/srv_status_component_test.rb
```diff
@@ -0,0 +1,147 @@
...
+
+# Some helpers to test the UI
+
+def matches_id_and_text?(widget, id, text)
+  return false unless widget.is_a?(Yast::Term)
+  return false unless widget.params
+  return false unless widget.params.any? do |p|
+    p.is_a?(Yast::Term) && p.value == :id && p.params.first =~ id
+  end
+  widget.params.any? { |p| p.is_a?(::String) && p =~ text }
```
jreidinger: term implement enumerable so using `.params` is usually useless.

jreidinger:
NP, LGTM

ancorgs:
I will wait for https://github.com/yast/yast-dns-server/pull/51 review and will merge both together

PR #404, opened 2015-10-08
Stack fix

jreidinger:


lslezak:
LGTM

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -3052,7 +3052,7 @@ def load_config
     def autoeject
       load_config unless @config
 
-      @config[autoeject] || false
+      @config["automatic_eject"] || false
```
mvidner: Better `@config.fetch("automatic_eject", false)` otherwise people may change the default value to `|| true` and get a surprise.

jreidinger: ok, make sense

mvidner:
LGTM, thank you!

jreidinger:
thanks, merged

PR #405, opened 2015-10-14
Fixed clipped labels in Arabic on some widgets

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=880701

The actual bug seems to be between Qt and Harfbuzz. Telling Qt to use an
older Harfbuzz solves the problem.
Using that for all languages risks breaking other languages with complex
scripts (Indic?) so let's limit this to Arabic.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Oct 14 15:28:37 UTC 2015 - mvidner@suse.com
+
+- Fixed clipped labels in Arabic on some widgets (bsc#880701).
+- 3.1.108.9
+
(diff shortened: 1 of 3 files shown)


ancorgs:
LGTM

nilxam:
LGTM. And I will report qt5 and harfbuzz issue to upstream.

kobliha:
Discussed offline: Martin will add reference to upstream bug here, because we need to be able to find out that the `workaround is not needed anymore`. Yast contains several similar workarounds and I'd like to keep the number of `I don't know what this does but it doesn't work without it` low. Thanks in advance.

kobliha:
And BTW, congrats for finding and implementing this workaround! :)

PR #406, opened 2015-10-19
Fixed clipped labels in Arabic on some widgets (bsc#880701).

mvidner:
This is a merge from SLE-12-GA, of #405.

The AArch64 commits that are visible in the commit list but not in the log are fine, because they were a back-port of  f493b69.


mvidner:
Tested to work, see https://trello.com/c/KRbg4TjD/365-3-sled12-sp1-p2-880701-sled12-loc-ar-group-labels-are-truncated-qgroupbox-yframe

scripts/yast2-funcs
```diff
@@ -111,6 +111,21 @@ function set_qt_home_dir()
     fi
 }
 
+function set_qt_arabic_workaround()
+{
+    # Work around clipped Arabic text, bsc#880701
+    # Upstream bug: https://bugreports.qt.io/browse/QTBUG-41450
+    # Locale check: first of the variables wins; man 7 locale
```
lslezak: I'd prefer some `TODO` or `FIXME` tag so we can easily find this workaround by `grep` later and verify whether it's still needed.

lslezak:
Thanks, LGTM.

mvidner:
Publicity: http://mvidner.blogspot.cz/2015/11/arabic-text-bugfix.html

PR #407, opened 2015-10-19
Hotfix. Propose correct default hostname / domain when network is set by linuxrc

mchf:
bnc#946047

### DIFF
--- a/library/types/src/modules/Hostname.rb
+++ b/library/types/src/modules/Hostname.rb
@@ -151,14 +151,6 @@ module Yast
       Ops.add(Ops.add(hostname, "."), domain)
     end
 
-    # It checks if the hostname cannot be used for setting default fqdn
-    #
-    # See bnc#946047
-    # Basicaly, linuxrc sometimes resolves IP when querying "hostname --fqdn"
-    def invalid_hostname?(hostname)
-      Mode.installation && IP.Check(hostname)
-    end
-
     # Retrieve currently set fully qualified hostname
     # (uses hostname --fqdn)
     # @return FQ hostname
@@ -245,6 +237,18 @@ module Yast
     publish function: :CurrentFQ, type: "string ()"
     publish function: :CurrentHostname, type: "string ()"
     publish function: :CurrentDomain, type: "string ()"
+
+  private
+
+    # It checks if the hostname cannot be used for setting default fqdn
+    #
+    # FIXME: Hotfix for bnc#946047. This should be dropped as part of fate#319639
+    # implementation
+    #
+    # Basicaly, linuxrc sometimes resolves IP when querying "hostname --fqdn"
+    def invalid_hostname?(hostname)
+      Mode.installation && IP.Check(hostname)
+    end
   end
 
   Hostname = HostnameClass.new
--- a/library/types/src/modules/Hostname.rb
+++ b/library/types/src/modules/Hostname.rb
@@ -38,6 +38,7 @@ module Yast
       Yast.import "IP"
       Yast.import "String"
       Yast.import "FileUtils"
+      Yast.import "Mode"
 
       # i18n characters in domain names are still not allowed
       #
@@ -150,6 +151,14 @@ module Yast
       Ops.add(Ops.add(hostname, "."), domain)
     end
 
+    # It checks if the hostname cannot be used for setting default fqdn
+    #
+    # See bnc#946047
+    # Basicaly, linuxrc sometimes resolves IP when querying "hostname --fqdn"
+    def invalid_hostname?(hostname)
+      return Mode.installation && IP.Check(hostname)
+    end
+
     # Retrieve currently set fully qualified hostname
     # (uses hostname --fqdn)
     # @return FQ hostname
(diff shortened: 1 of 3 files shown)


mvidner:
Sorry, this looks bad.
Formerly, `CurrentFQ` had a simple task of reading the output of `hostname --fqdn`, then we added higher level logic to it ([back in the YCP days](https://github.com/mchf/yast-yast2/blame/4a9a454d212811f90ea4166ab2d6e2bd8be6c3ef/library/types/src/modules/Hostname.rb#L153), but missed that during reviews of later changes) and now we are adding even more logic. Luckily, the code also got uglier so I noticed ;-)

Also, shouldn't this go to a SLE-12-SP1 branch?

Anyway, I know this is supposed to fix a real problem, but give me a couple of hours to find a less ugly solution.

mchf:
it is ugly intentionally bcs it is hotfix for a bug in linuxrc. The plan is to create a fate on it to get it fixed in proper place.

mvidner:
I see. Meanwhile, here is a grep of `CurrentFQ` in master
```
yast/yast-ca-management/package/yast2-ca-management.changes:243:- use Hostname::CurrentFQ() as hostname fallback in proposal.                                                                 
yast/yast-ca-management/src/clients/ca_mgm_auto.rb:729:              Ops.set(retmap, "stdout", Hostname.CurrentFQ)                                                                            
yast/yast-ca-management/src/clients/ca_mgm_auto.rb:981:              Ops.set(retmap, "stdout", Hostname.CurrentFQ)                                                                            
yast/yast-ca-management/src/clients/ca_mgm_proposal.rb:101:            @retmap["stdout"] = Hostname.CurrentFQ                                                                                 
yast/yast-ca-management/src/clients/ca_mgm_proposal.rb:122:          @longhostname = Hostname.CurrentFQ                                                                                       
yast/yast-crowbar/src/include/crowbar/complex.rb:1737:            Hostname.CurrentFQ                                                                                                          
yast/yast-dns-server/src/modules/DnsZones.pm:337:       my $hostname = Hostname->CurrentFQ();                                                                                                 
yast/yast-network/src/modules/DNS.rb:256:      fqhostname = Hostname.CurrentFQ if fqhostname.nil? || fqhostname.empty?                                                                        
yast/yast-yast2/library/types/src/modules/Hostname.rb:156:    def CurrentFQ                                                                                                                   
yast/yast-yast2/library/types/src/modules/Hostname.rb:188:      fqhostname = CurrentFQ()                                                                                                      
yast/yast-yast2/library/types/src/modules/Hostname.rb:208:      fqhostname = CurrentFQ()                                                                                                      
yast/yast-yast2/library/types/src/modules/Hostname.rb:236:    publish function: :CurrentFQ, type: "string ()"                                                                                 
yast/yast-yast2/library/types/test/hostname_test.rb:20:    expect(Yast::Hostname.CurrentFQ).to eq cmd_hostname
yast/yast-yast2/library/types/test/hostname_test.rb:37:    expect(Yast::Hostname.CurrentFQ).to eq etc_hostname
yast/yast-yast2/library/types/test/hostname_test.rb:7:describe "Hostname#CurrentFQ" do
```

mchf:
btw it should go into sp1 branch but yast2 is not branched yet. At least I didn't find sp1 branch ;-)

mvidner:
Right, was confused about that. Until we need to commit things *not* for SP1, we don't need the branch.

mvidner:
Sorry, I did not find time to improve this. How about you create a bug/fate for the proper fix, put it in a comment in the code, and then merge?

kobliha:
@mvidner Creating a FATE was in the plan, we'll have a discussion with @mchf after planning, maybe you want to join us.

mvidner:
Code looks good to me, so please make it mergeable. And you may want to install rubocop-0.29.1 locally, because:
```
library/types/src/modules/Hostname.rb:241:3: C: Keep a blank line before and after private.
  private
  ^^^^^^^
library/types/src/modules/Hostname.rb:244:7: C: Annotation keywords should be all upper case, followed by a colon and a space, then a note describing the problem.
    # FIXME:
      ^^^^^^
library/types/src/modules/Hostname.rb:252:1: C: Extra empty line detected at class body end.
```

mchf:
well, exactly this rubocop's complain is total nonsense for me ... :-/

ok going to fix, rebase and so on.

mvidner:
It's a cost that we pay for the more sensible cops.

jreidinger:
well first one is good one. It ensures that global access modifier is visible ( because it have empty line before and after )

The second one is for unified notation of TODO, FIXME and so on, so it is easier to grep for it

And the third one is to avoid useless endline after final end as it do not provide any benefit.

So for me this make sense for me.

mchf:
It is matter of taste. If things are done "my" way, it is better readable to me. Especially in case of multiline FIXME notes and empty line after the end. But never mind I'll change it.

kobliha:
I have to agree with @mchf that sometimes RuboCop complains just too much. But it can be actually configured in a different way if that fits better. I suggest to listen to the cop now and open a new thread on ML to make cops work better for us. RuboCop is just a servant and he does exactly what we tell him to do.

mvidner:
Thank you!
LGTM

mchf:
Thanks for reviews.

PR #408, opened 2015-10-21
Replace "Skip" with "Ignore"

imobach:
Replace "Skip" with "Ignore" when some problem occurs when fetching packages.

replace-skip-with-ignore


lslezak:
LGTM

The bug still does not contain the feedback from the documentation team, but Stefan said OK, so I'm not sure if we need to still wait or just submit it... :grey_question:

kobliha:
Just submit it please, LGTM

imobach:
@lslezak @kobliha Thanks, submitting right now.

PR #409, opened 2015-10-22
Removed an unused file.

mvidner:
We accidentally stopped distributing Version.rb at the Ruby conversion
so it turns out no code uses it anyway.*

But it does fail RuboCop checks.

\* I have double checked with `grep -r -E Version..?yast2 ~/my-clean-github-checkout/yast/yast-*`


jreidinger:
LGTM, it was stupid file even before as there is better way to get version of given package.

PR #411, opened 2015-11-04
Fixed clipped dialog label (bsc#948381)

lslezak:
- 3.1.156

### Debugged and tested with a small script

    require "yast"
    Yast.import "Wizard"
    Yast.import "Progress"

    Yast::Wizard.CreateDialog
    Yast::Progress.New("Initializing Boot Loader Configuration", " ", 1, [], [], "")
    Yast::UI.UserInput
    Yast::Wizard.CloseDialog

### Note

The tricky part is that you have to test it in a low-res screen (less than 800x600) to force usage of the simple Wizard implementation in the `Wizard` module. In the full Wizard UI the problem does not happen.

### The original output

dialog_label_broken

### With the fix applied

dialog_label_fixed

### DIFF
--- a/library/wizard/src/modules/Progress.rb
+++ b/library/wizard/src/modules/Progress.rb
@@ -458,6 +458,7 @@ module Yast
 
       if !UI.WizardCommand(term(:SetDialogHeading, window_title))
         UI.ChangeWidget(Id(:title), :Value, window_title)
+        UI.RecalcLayout
       end
       Wizard.SetHelpText(help_text) if "" != help_text && nil != help_text
       Wizard.DisableBackButton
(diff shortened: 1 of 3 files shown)


kobliha:
LGTM

library/wizard/src/modules/Progress.rb
```diff
@@ -458,6 +458,7 @@ def New(window_title, progress_title, length, stg, tits, help_text)
 
       if !UI.WizardCommand(term(:SetDialogHeading, window_title))
         UI.ChangeWidget(Id(:title), :Value, window_title)
+        UI.RecalcLayout
```
kobliha: BTW: Nitpicking, but, do you want to add some comment here?

PR #415, opened 2015-11-24
Y2logs rewrite

wfeldt:


scripts/save_y2logs
```diff
@@ -7,110 +7,140 @@
...
+	;;
+
     *.tgz|*.tar.gz)
 	TARGET="$1"
-	COMPRESSION=--gzip
+	COMPRESSION=gzip
 	;;
 
-    *.tar.bz2)
+    *.tar.xz)
```
jreidinger: I also see usage of `txz` so maybe also support it?

scripts/save_y2logs
```diff
@@ -7,110 +7,140 @@
...
-	COMPRESSION=--bzip2
+	COMPRESSION=xz
+	;;
 
-	if [ ! -x /usr/bin/bzip2 ]; then
-	    echo "FATAL: /usr/bin/bzip2 not available" >&2
-	    # This might easily happen in the inst-sys
-            exit 3
-	fi
+    *.tar.bz2)
```
jreidinger: same here with `tbz2` as shortcut.

scripts/save_y2logs
```diff
@@ -7,110 +7,140 @@
...
+	TARGET="$1"
+	COMPRESSION=bzip2
 	;;
 
     *)
-	echo "FATAL: Uncompressed archives not supported" >&2
-	echo "Use one of: .tgz .tar.gz .tar.bz2"          >&2
+	echo "FATAL: argument is not a (compressed) tar file." >&2
+	echo "Use one of these suffixes: .tar .tgz .tar.gz .tar.xz .tar.bz2" >&2
 	exit 4
```
jreidinger: target in all reasonable branches is always same TARGET=$1, maybe it make sense to assign it outside of case statement.

scripts/save_y2logs
```diff
@@ -7,110 +7,140 @@
...
   esac
 fi
 
+# check for compression program
+if [ "$COMPRESSION" -a ! -x "/usr/bin/$COMPRESSION" ]; then
```
jreidinger: can we be a bit robust with `which $COMPRESSION` for location of compression algorithm? Who know when someone decide to drop /usr or move it to some funny location.

scripts/save_y2logs
```diff
@@ -7,110 +7,140 @@
...
-if [ -x /bin/dmesg ]; then
-    dmesg > $TMPDIR/dmesg
-    LIST="$LIST $TMPDIR/dmesg"
-fi
+# 1. gather some information in $TMPDIR
 
-if [ -f /var/log/zypp/history ]; then
-    LIST="$LIST $( cd /var/log/; ls zypp/history*)"
+# last kernel messages, if any
+if [ -x /bin/dmesg ]; then
```
jreidinger: what about `journalctl --dmesg` for journalctl? it can be useful on running system.

scripts/save_y2logs
```diff
@@ -7,110 +7,140 @@
...
+VAR_LOG_FILES='\
+  YaST2 \
+  Xorg.0.log \
+  zypper.log zypp/history* pk_backend_zypp \
+  pbl.log linuxrc.log wickedd.log \
+  evms-engine.* \
+'
+cd /var/log
+for i in $VAR_LOG_FILES ; do
+  [ -e "$i" ] && LIST_VAR="$LIST_VAR $i"
```
jreidinger: this does not work with asterisk. Try e.g.

```
jreidinger@pepa:/tmp> ls /tmp/y2log*
/tmp/y2log-1.gz.swp  /tmp/y2logs.tgz
jreidinger@pepa:/tmp> [ -e "/tmp/y2log*" ]
jreidinger@pepa:/tmp> echo $?
1
jreidinger@pepa:/tmp> [ -e "/tmp/y2logs.tgz" ]
jreidinger@pepa:/tmp> echo $?
0
```

scripts/save_y2logs
wfeldt: Uhg, I can go with txz, but tbz2 is a bit over-the-top. bzip2 is quite dead.

scripts/save_y2logs
wfeldt: it does work, $VAR_LOG_FILES gets expanded

scripts/save_y2logs
wfeldt: just tried and it's quite big (120MB on my machine); xz deals wiith it, though...

scripts/save_y2logs
jreidinger: ah, dark bash magic, with having it in `''` and then expanded for cycle. OK, I trust to verify it.

scripts/save_y2logs
jreidinger: point is that it is replacement for dmesg utility itself. At least on my pc, both return same output. Now it do not make much sense to have both, as both should return same lines.

jreidinger:
it looks good now, just we should store dmesg or journalctl --dmesg, not both as it have same content see https://github.com/yast/yast-yast2/pull/415#discussion_r45743809 just we should try both if any of it is available, then use it.

wfeldt:
dmesg and journalctrl --dmesg are not the same on my machine

jreidinger:
ok, then it make sense having both. LGTM

PR #416, opened 2015-11-24
SrvStatusComponent -> ServiceStatus

ancorgs:
Replace `UI::SrvStatusComponent` with `UI:ServiceStatus`. The new class has a much better API, as explained below.

Instead of the service_xx methods, it knows expect an object that complains to the SystemdService API. For non-standard services the user is expected to implement their own class.

Instead of using callbacks, handle_input now returns a meaningful symbol.

See https://github.com/yast/yast-dns-server/pull/52 for code changes in yast-dns-server

This PR also introduces `CWM.save_current_widgets` and `CWM.validate_current_widgets`, needed to implement an "apply" button.

### DIFF
--- a/library/general/src/lib/ui/service_status.rb
+++ b/library/general/src/lib/ui/service_status.rb
@@ -17,46 +17,37 @@
 #  you may find current contact information at www.suse.com
 
 require "yast"
-Yast.import "Service"
 Yast.import "UI"
 
 module UI
-  # Component encapsulating the widgets for managing the status of services (both
-  # currently and on system boot) and the behavior associated to those widgets
+  # Widgets for managing the status of services (both currently and on system
+  # boot) and the behavior associated to those widgets
   #
-  # TODO: The API is experimental and subject to change. This component is only
-  # used by yast2-dns-server and should not be used by any other module until
-  # an upcoming refactoring, planned for SLE12-SP2. To see some of the planned
-  # improvements, check https://github.com/yast/yast-yast2/pull/401
-  #
-  # As long as #handle_input is invoked in the event loop, the component will
-  # handle interactive starting and stopping of the service on user demand. In
-  # addition #reload can be used after saving the settings.
-  #
-  # To manage the status on boot, the component can be queried for the user
-  # selection using #enabled?. In addition enabled_callback (in constructor)
-  # can be used to observe the status of the corresponding field in the UI.
-  class SrvStatusComponent
+  # As long as #handle_input is invoked in the event loop, the widget will
+  # handle interactive starting and stopping of the service on user demand.
+  class ServiceStatus
     include Yast::UIShortcuts
     include Yast::I18n
     include Yast::Logger
 
-    # @param service_name [String] name of the service as expected by
-    #   Yast::Service
-    # @param reload [Boolean] initial value for the "reload" checkbox.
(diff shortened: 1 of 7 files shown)


jreidinger:
API looks much better. thanks

jreidinger:
still LGTM

library/general/src/lib/ui/service_status.rb
```diff
@@ -17,46 +17,40 @@
...
+    # @param service [Object] An object providing the following methods:
+    #   #name, #start, #stop, #enabled?, #running?
+    #   For systemd compliant services, just do
+    #   Yast::SystemdService.find("name_of_the_service")
+    # @param reload [Boolean] Initial value for the "reload" checkbox.
+    #   Keep in mind it will always be displayed as unchecked if the service
+    #   is not running, despite the real value.
+    # @param reload_label [Symbol] Type of label for the "reload" checkbox.
+    #   :restart means that the service will be restarted, not just reloaded.
+    def initialize(service, reload: true, reload_label: :reload)
```
mvidner: @ancorgs I don't understand this "reload" business. What is being reloaded? The status of the service or the service itself? `@service.reload` is missing but the comment of `reload_label` suggests otherwise. I am confused.

ancorgs: @mvidner See screenshot at https://cloud.githubusercontent.com/assets/3638289/11421803/2d793940-9437-11e5-918b-1d31b56307f6.png

There is a checkbox there for the users to specify their want to (reload/restart) after saving changes.
The argument `reload` is the initial value for that checkbox. The `reload_label` argument is used to customize (to some extend, I didn't wanted to allow arbitrary customization) the label of that checkbox.

Doing the real restart/reload is not direct responsibility of the widget, since it's something that will happen after clicking the ok/finish/next/whatever button on the caller dialog.

mvidner: Ah, OK

jreidinger: maybe descriptive name helps...like `checkbox_label` ? as I agree with mvidner that name variable by its default value is confusing

ancorgs: I don't think that having a parameter in the constructor named as the initial value of a variable is confusing, it's actually the common case.

What about renaming `reload` to `reload_flag`. Thus, the parameters would be `reload_flag` and `reload_flag_label` and the method to check it would be `ServiceStatus#reload_flag?`. To be consistent, I guess we should also use `ServiceStatus#reload_flag?`.

mvidner: I think an important improvement is to explain in the constructor description that the `reload` information is not passed to the `service` *by this object/widget*; that the caller is expected to do that.
To reinforce that concept, the argument could be renamed to `reload_suggested` (where I understand "suggested" to mean "initialize a boolean setting to true")

Ah, `reload_flag`, `reload_flag_label`, `reload_flag?` look good. I hadn't considered `enabled_flag?` until now, but if it means that the caller is expected to act on it, instead of this class, then yes.

ancorgs: Yes, the caller is expected to enable/disable the service. The component only takes care of starting and stopping (in addition to displaying an UI that makes sense and it's consistent with the current state).

PR #417, opened 2015-11-27
UI::ServiceStatus: rename reload and enabled to xxx_flag

ancorgs:
To address the concerns expressed in https://github.com/yast/yast-yast2/pull/416

### DIFF
--- a/library/general/src/lib/ui/service_status.rb
+++ b/library/general/src/lib/ui/service_status.rb
@@ -25,6 +25,10 @@ module UI
   #
   # As long as #handle_input is invoked in the event loop, the widget will
   # handle interactive starting and stopping of the service on user demand.
+  #
+  # It also provides checkboxes (reload_flag and enabled_flag) for the user
+  # to specify whether the service must be reloaded/restarted after
+  # configuration changes and whether it must be enabled at boot time.
   class ServiceStatus
     include Yast::UIShortcuts
     include Yast::I18n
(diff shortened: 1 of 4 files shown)


ancorgs:
@mvidner @jreidinger It would be nice if you could take a quick look, so we submit this (and the related modules) before the next TW snapshot.

library/general/src/lib/ui/service_status.rb
```diff
@@ -34,19 +34,19 @@ class ServiceStatus
     #   #name, #start, #stop, #enabled?, #running?
     #   For systemd compliant services, just do
     #   Yast::SystemdService.find("name_of_the_service")
-    # @param reload [Boolean] Initial value for the "reload" checkbox.
+    # @param reload_flag [Boolean] Initial value for the "reload" checkbox.
     #   Keep in mind it will always be displayed as unchecked if the service
     #   is not running, despite the real value.
-    # @param reload_label [Symbol] Type of label for the "reload" checkbox.
+    # @param reload_flag_label [Symbol] Type of label for the "reload" checkbox.
     #   :restart means that the service will be restarted, not just reloaded.
```
jreidinger: document both possible values, so also `:reload` value in comment

jreidinger:
LGTM

library/general/src/lib/ui/service_status.rb
```diff
@@ -34,19 +34,20 @@ class ServiceStatus
     #   #name, #start, #stop, #enabled?, #running?
     #   For systemd compliant services, just do
     #   Yast::SystemdService.find("name_of_the_service")
```
mvidner: > Note that this widget will #start and #stop the service by itself but the actions referenced by the checkboxes (reload and enable) are expected to be done by the caller, when the whole dialog is done.

mvidner:
I suggest that one additional comment, otherwise LGTM.

PR #418, opened 2015-12-11
added detection of zKVM to Arch.rb (for proper fix of bsc#956736)

jsrain:


library/general/src/modules/Arch.rb
```diff
@@ -446,6 +449,23 @@ def is_kvm
...
+
+    # zKVM means KVM on IBM System z
+    # true if zKVM is running
+    #
+    # @return true if we are running on zKVM hypervisor
+    def is_zkvm
+      if @_is_zkvm.nil?
+        # using different check than on x86 as recommended by IBM
+        @_is_zkvm = Yast::WFM.Execute(".local.bash", "egrep 'Control Program: KVM' /proc/sysinfo") == 0
+        @_is_zkvm &&= s390
```
jreidinger: maybe it make sense to have different order as you would like do quick check for s390 and then run some script if needed.

library/general/test/arch_test.rb
```diff
@@ -0,0 +1,47 @@
...
+
+require_relative "test_helper"
+
+Yast.import "Arch"
+
+require "yast"
+
+describe Yast::Arch do
+
+  describe "DetectzKVM" do
```
jreidinger: describe should match method, so 
```
describe ".is_zkvm" do
```
so if you run rspec --format doc it will write class, its methods and then its specification.

library/general/test/arch_test.rb
```diff
@@ -0,0 +1,47 @@
...
+    it "zKVM on S/390" do
+      allow(Yast::WFM).to receive(:Execute).and_return 0
+      allow(Yast::SCR).to receive(:Read).and_return "s390_64"
+
+      is_zkvm = Yast::Arch.is_zkvm
+
+      expect(is_zkvm).to eq(true)
+    end
+
+    it "zVM on S/390" do
```
jreidinger: two times same header and different result. it should probably looks like this:
```
it "returns true if on s390 and in zkvm environment" do
...
it "returns false if on s390 and outside of zkvm environment" do
...
it "returns false on other architectures" do
...
```

jreidinger:
LGTM

PR #420, opened 2016-01-08
initial widgets implementation with a lot of todos and missing doc

jreidinger:
do not merge yet, I just would like to get initial comments


ancorgs:
I would like to see a usage example

jreidinger:
@ancorgs OK, make sense, I will do it together with original WFM documentation and example.

mvidner:
What is the scope and intent of this? Do I understand right that this is an evolution of CWM in the sense of keeping the semantics and vocabulary, while improving the usability of the library in the Ruby language? Any other goals, non-goals?

jreidinger:
@mvidner well goal is to have nice object API to define widgets in CWM. So for modules that are CWM based, it makes good chance to write it clean and maybe it evolve enough to use it more widely.

mvidner:
OK. I would expect some links to the docs for the original CWM. Unfortunately the comments in the source are not too detailed, resulting in <http://www.rubydoc.info/github/yast/yast-yast2/Yast/CWMClass>. We have Docbook docs in https://github.com/yast/yast-yast2/tree/master/library/cwm/doc but I could not find any rendering online :( If you install docbook-css-stylesheets.rpm then the local copies will be rendered by the browser (but not ones served by raw.github... as text/plain).
So it looks like a side quest.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,425 @@
...
+    end
+  end
+
+  class PasswordWidget < AbstractWidget
+    include ValueBasedWidget
+
+    def description
+      unless respond_to?(:label)
+        raise "For input field widget '#{self.class}' label method have to be defined"
+      end
```
mvidner: What about this?
```rb
module LabelRequired
  def description
    raise "'#{self.class}' requires that you define a 'label' method" unless respond_to? :label
    super
  end
end

class PasswordWidget < AbstractWidget
  include ValueBasedWidget
  include LabelRequired
```

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,425 @@
...
+    #
+    def description
+      unless respond_to?(:label)
+        raise "For input field widget '#{self.class}' label method have to be defined"
+      end
+
+      res = {
+        "widget"        => :combobox,
+      }
+      res["items"] = items if respond_to?(:items)
```
mvidner: Does this have the same behavior as defining `ComboBoxWidget#items` returning an empty list? Then that method would be a much better place to document it.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,425 @@
...
+      res = {
+        "widget"        => :multi_selection_box,
+      }
+      res["items"] = items if respond_to?(:items)
+
+      res.merge(super)
+    end
+
+  protected
+    def values
```
mvidner: I would also expect a `value` accessor, also acting on an array.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,425 @@
...
+    def description
+      unless respond_to?(:label)
+        raise "For input field widget '#{self.class}' label method have to be defined"
+      end
+
+      res = {
+        "widget"        => :intfield,
+      }
+      res["minimum"] = minimum if respond_to?(:minimum)
+      res["maximum"] = minimum if respond_to?(:maximum)
```
mvidner: Aren't these mandatory? Looking up... CWM defaults to 0 and 2^31 - 1 but I think it is better to be explicit here.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,425 @@
...
+    end
+
+  private
+
+    def handle(widget, event)
+      return if event["ID"] != widget
+
+      @block.call
+
+      nil
```
mvidner: `CWM handle` function docs say

> It should return nil to continue processing or a symbol to be returned from the dialog.

If you differ intentionally, how do you intend to exit the event loop?

library/cwm/src/modules/CWM.rb
```diff
@@ -397,7 +397,7 @@ def cleanupWidgets(widgets)
       Builtins.foreach(widgets) do |w|
         @processed_widget = deep_copy(w)
         toEval = Convert.convert(
-          Ops.get(w, "clean_up"),
+          Ops.get(w, "cleanup"),
```
mvidner: Please put the clean_up clean-up to a separate commit with an explanation why this change is needed.

library/cwm/src/modules/CWM.rb
```diff
@@ -462,6 +462,7 @@ def ValidateMaps(widgets)
           if key != "label" ||
               Ops.get(v, "widget") != :radio_buttons &&
                   Ops.get(v, "widget") != :custom &&
+                  Ops.get(v, "widget") != :rich_text &&
```
mvidner: Also deserves a description

library/cwm/src/modules/CWM.rb
mvidner: Notably cwm/doc/widgets.xml documents "clean_up".

mvidner:
Thanks for the example! Um, does it work for you? I have only tried on 13.2* where both widgets end up as InputFields.

*) also needed commenting out UI.SetApplicationIcon in WizardClass.set_icon)

library/cwm/src/lib/cwm/widget.rb
mvidner: Or, why reinvent the wheel:
```rb
require 'abstract_method'
class PasswordWidget < AbstractWidget
  abstract_method :label
...
end

Comment (jreidinger) [inline library/cwm/src/lib/cwm/widget.rb]: make sense

File: library/cwm/src/lib/cwm/widget.rb (hunk 4 above)
Comment (jreidinger) [inline library/cwm/src/lib/cwm/widget.rb]: It is question if we want it as own API or just layer over CWM, which adds value, but do not modify defaults

File: library/cwm/src/lib/cwm/widget.rb (hunk 5 above)
Comment (jreidinger) [inline library/cwm/src/lib/cwm/widget.rb]: I remove this call, as I found it not so handful

**Comment by jreidinger** [comment]
@mvidner problem is that Y2DIR processing is done too early, so you need to copy new CWM.rb to /usr/share/YaST2/modules or set Y2DIR before running client to wfm/src as I write ancor on IRC

File: library/cwm/src/lib/cwm/widget.rb (hunk 1 above)
Comment (jreidinger) [inline library/cwm/src/lib/cwm/widget.rb]: well, on other hand abstract method is third party gem, right? and we want to have minimal dependencies in yast2.

Comment (mvidner) [inline library/cwm/src/lib/cwm/widget.rb]: If you are looking for a "gateway drug" gem, https://github.com/openSUSE/abstract_method is exactly it. The API is one (1) method, authored by David Majda, co-owned by a certain Josef Reidinger on Rubygems.org :smile:

File: library/cwm/examples/example_helper.rb [hunk 8]
```diff
@@ -0,0 +1,17 @@
+root_location = File.expand_path("../../../..", __FILE__)
```
mvidner: Note to myself: On 13.1 the helper does not help but I can use
```sh
Y2DIR=../../../library/cwm/src /usr/lib/YaST2/bin/y2base ./object_api.rb qt
```

jreidinger: helper helps with loading lib, but problem is that Y2DIR is read at start of yast2 before reading helper

library/cwm/src/lib/cwm/widget.rb
jreidinger: I know that gem :smile: I just saying that we need to at first approve it..it is used by cheetah, so we can use it when we start using cheetah

library/cwm/examples/example_helper.rb
mvidner: The helper should work if the binary is `ruby` instead of `y2base`, especially if we add
```rb
    if Yast.ui_component == ""
      Yast.ui_component = ARGV[0] || "ncurses"
    end
```
See also https://github.com/yast/yast-ruby-bindings/pull/142

jreidinger:
@mvidner updated. I hope I fix all your concerns

jreidinger:
@mvidner @ancorgs updated, I add quite useful feature for easy widget enable/disable

ancorgs:
Much better now (less code duplicate).

And the examples confirm it works as I suspected... and expected.

So beside @mvidner comments about `cleanup`, LGTM

library/cwm/examples/object_api.rb
```diff
@@ -0,0 +1,78 @@
...
+    self.widget_id = "generate"
+    @lucky_number = lucky_number
+  end
+
+  def label
+    _("Generate Lucky Number")
+  end
+
+  def handle(widget, event)
+    return unless my_event?(widget, event)
```
mvidner: This should really be the default behavior of AbstractWidget. If you think that some widgets will need to process other events then that can be enabled in their constructor.

library/cwm/examples/object_api.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def store(_widget, _event)
+    @result = value
+  end
+end
+
+class GenerateButton < CWM::PushButtonWidget
+  def initialize(lucky_number)
+    self.widget_id = "generate"
+    @lucky_number = lucky_number
```
mvidner: It is a widget, not a number, so it should be named `lucky_number_widget`

library/cwm/examples/object_api.rb
```diff
@@ -0,0 +1,78 @@
...
+      textdomain "example"
+
+      lucky_number_widget = LuckyNumberWidget.new
+      button_widget = GenerateButton.new(lucky_number_widget)
+
+      widgets = [lucky_number_widget, button_widget]
+
+      content = HBox(
+        button_widget.widget_id,
+        lucky_number_widget.widget_id
```
mvidner: Calling `widget_id` manually feels stupid.

library/cwm/examples/object_api_enabling.rb
```diff
@@ -0,0 +1,122 @@
...
+    return unless my_event?(widget, event)
+
+    Yast::Builtins.y2milestone("int handle called")
+
+    nil
+  end
+end
+
+class EnableButton < CWM::PushButtonWidget
+  def initialize(lucky_number, disable_button)
```
mvidner: See, the 2nd argument is a button, but the 1st one is not really a number.

library/cwm/examples/object_api_enabling.rb
```diff
@@ -0,0 +1,122 @@
...
+  class ExampleDialog
+    include Yast::I18n
+    include Yast::UIShortcuts
+    def run
+      textdomain "example"
+
+      lucky_number_widget = LuckyNumberWidget.new
+      disable_button_widget = DisableButton.new(lucky_number_widget)
+      enable_button_widget = EnableButton.new(lucky_number_widget, disable_button_widget)
+      disable_button_widget.enable_button = enable_button_widget
```
mvidner: It's not really related to the library, but this asymmetry spoils the beauty of your example :)
```rb
  disable_button_widget = DisableButton.new(lucky_number_widget, nil)
  enable_button_widget = EnableButton.new(lucky_number_widget, nil)
  disable_button_widget.enable_button = enable_button_widget
  enable_button_widget.disable_button = disable_button_widget
```
If this pattern turns out to be common we should add a helper for it.

library/cwm/examples/object_api_enabling.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      widgets = [lucky_number_widget, enable_button_widget, disable_button_widget]
```
mvidner: I've always thought that this enumeration is stupid because it can be computed from the `content`. There was some justification in YCP where the strings could mean labels or widget IDs but now that widgets are objects, `widgets` can be computed.

library/cwm/examples/object_api_enabling.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      content = HBox(
+        enable_button_widget.widget_id,
+        disable_button_widget.widget_id,
+        lucky_number_widget.widget_id
+      )
+
+      Yast::Wizard.CreateDialog
+      CWM.ShowAndRun(
+        "contents" => content,
```
mvidner: Use "contents" in both cases please.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
+require "yast"
+
+require "abstract_method"
+
+module CWM
+  # Represent base for any widget used in CWM. It can be passed as "widget" argument. For more
+  # details about usage see {CWM.ShowAndRun}
+  class AbstractWidget
```
mvidner: Why not simply "Widget"?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+    include Yast::UIShortcuts
+    include Yast::I18n
+
+    # @return [String] id used for widget
+    attr_accessor :widget_id
+
+    # defines widget type for CWM usage
+    def self.widget_type=(val)
```
mvidner: I don't get it. Why not simply define an instance method `widget_type` in the children of AbstractWidget, returning the appropriate string?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+    # - `#validate` [Boolean (String, Hash)] validate widget value, arguments are
+    #   widget key and event map causing validation, for other validation types, overwrite description
+    # - `#init` [nil (String)] initialize widget. like its value. Parameter is widget id
+    # - `#handle` [Symbol,nil (String, Hash)] handle widget changed value or press.
+    #   return value is usually nil, returning symbol can be used to send different event
+    # - `#store` [nil (String, Hash)] store widget value after user confirm dialog
+    # - `#cleanup` [nil (String)] cleanup after widget is destroyed
+    # @raise [RuntimeError] if required method is not implemented or widget id not set.
+    def description
+      raise "Widget '#{self.class}' does set its widget ID" if widget_id.nil?
```
mvidner: "does **not** set"

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+      if !respond_to?(:widget_type)
+        raise "Widget '#{self.class}' does set its widget type"
```
mvidner: "does not set"

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+
+    # shortcut from Yast namespace to avoid including whole namespace
+    # TODO: kill converts in CWM module, to avoid this workaround for funrefs
+    def fun_ref(*args)
+      Yast::FunRef.new(*args)
+    end
+
+  private
+
+    def init_method
```
mvidner: If it is named `*_method` then I would expect it to be a ::Method. The helpers should be renamed to `*_fun_ref`. All the more important because the `::Method`s are also involved.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+  #       when :reset then ...
+  #       when :undo then ...
+  #       else ...
+  #       end
+  #     end
+  #   end
+  class CustomWidget < AbstractWidget
+    self.widget_type = :custom
+    # custom witget without content do not make sense
+    abstract_method :content
```
mvidner: "contents"  for consistency with CWM

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+
+    def description
+      {
+        "custom_widget" => content
+      }.merge(super)
```
mvidner: In multiple places I see `some_hash.merge(super)`, which makes it impossible to override the superclass values. Please reverse all the merges.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+    end
+  end
+
+  # Empty widget useful mainly as place holder for replacement or for catching global events
+  #
+  # @example empty widget usage
+  #   widget = CWM::EmptyWidget("replace_point")
+  #   CWM.ShowAndRun(
+  #     "content" => VBox(widget.widget_id),
```
mvidner: contents

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+    # @return [Array<Array<String>>]
+    # @example items method in widget
+    #   def items
+    #     [
+    #       [ "Canada", _("Canada")],
+    #       [ "USA", _("United States of America")],
+    #       [ "North Pole", _("Really cold place")],
+    #     ]
+    #   end
+    def items
```
mvidner: `[]`

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,450 @@
...
+    end
+
+    def description
+      {
+        "items" => items
+      }.merge(super)
+    end
+
+    # change list of items offered in widget. Format is same as in {#items}
+    def change_items(items_list)
```
mvidner: `def items=(items_list)` ?

mvidner:
Duh! Only now I have realized that in the examples we are busy building subclasses, but that is quite unnatural. We should be creating instances instead, don't you think?

library/cwm/examples/object_api.rb
jreidinger: short id collide with `Object#id`. Having magic in content parsing is possible, but I do not want to modify too much CWM as it is not test covered and still quite widely used.

library/cwm/examples/object_api_enabling.rb
jreidinger: I do not think it is common pattern, symmatry can be possible, passing nil is not nice.

library/cwm/examples/object_api_enabling.rb
jreidinger: yes, even old WFM can extract widgets from content ( detecting strings ), but with objects it is more tricky, as it require modification of WFM itself.

library/cwm/src/lib/cwm/widget.rb
jreidinger: Widget can get impression that it can be used itself as widget, but it is not true. instace of Widget itself is useless, so I add there abstract to make it explicit.

library/cwm/src/lib/cwm/widget.rb
jreidinger: I used to have it that way, but found it not so nice. Compare
```
class A
  self.widget_type :button

#versus
class A
  def widget_type
     :button
  end
```
I feel it is more like constant and not method.

library/cwm/src/lib/cwm/widget.rb
jreidinger: in fact FunRef is_a ::Method just with additional info. and I hope in future when we modify CWM to not be type strict it can be common ::Method.

library/cwm/src/lib/cwm/widget.rb
jreidinger: you are right

library/cwm/src/lib/cwm/widget.rb
jreidinger: I am considering it, but it is confusing as
```
a.items = [:a, :b, :c]
a.items # => []  original value
```

jreidinger:
@mvidner regarding subclass it depends what is your goal. My goal is to have reusable widgets. Just check https://github.com/yast/yast-bootloader/pull/280/files#diff-399e7f5b6714a35ea085089f1537e2aeR83

especially store and init part is very annoying with instances. Just consider:

```
InputField.new(
  label: _("My label"),
  help: _("blablbalba" \
    "blablabla" \
    "blablabla"
  ),
  init: Proc.new do |widget|
     ...
   end,
   store: Proc.new do |widget, event|
     ...
   end,
   validate: Proc.new do |widget, event|
      ....
   end
end
```
and with subclass you have separated functionality for widget and in dialog, you only say something like `MyWidget.new` so for me it is better separated and better reusable same widget in more dialogs ( e.g. usable for services reload/restart widget )

ancorgs:
That was actually my motivation to ask for an example. To make the subclassing more obvious. I think it's a good approach, but since I understand it can be counterintuitive at a first glance, I wanted it to be explicitly explained.

library/cwm/src/lib/cwm/widget.rb
jreidinger: and it internal private method, so I do not consider name so important

library/cwm/src/lib/cwm/widget.rb
mvidner: Aaah, now I see. So that would be `items`, `items=` doing a QueryWidget/ChangeWidget and a `default_items` setting the `description`... which I still don't like much better.

jreidinger:
@mvidner I hope I address all your comments

ancorgs:
We were discussing offline in the Spanish office about the subclassing vs instances issue and I think we came up with a good solution. @imobach will write it down here this afternoon.

ancorgs:
In short (just as some previous food for thoughts), the idea is to extract the behavior (init, store, validate, etc.) into separate objects that will be passed to the widget instances. Let's call it composition.

imobach:
As @ancorgs has said, the idea would be to extract the behavior to separated classes. Something like this:

```ruby
class MyWidgetBehavior # just a plain ruby object.
  def init
    # ...
  end

  def store
    # ...
  end

  def validate
    # ...
  end
end

input = InputField.new(label: _("My label"), help: _("Help..."), behavior: MyWidgetBehavior))
```

For the behavior parameter, we could use a initialized object or maybe a string.

```ruby
input = InputField.new(label: _("My label"), help: _("Help..."), behavior: MyWidgetBehavior.new(...)))
input = InputField.new(label: _("My label"), help: _("Help..."), behavior: "my_widget_behavior")
```
This is just a draft, but I think that we could explore this path. What do you think about this?

library/cwm/test/widgets_test.rb
```diff
@@ -0,0 +1,28 @@
+require_relative "test_helper"
```
mvidner: Almost there: Travis ran this with `sh`. We need `#!/usr/bin/ruby`.

library/cwm/src/modules/CWM.rb
```diff
@@ -1026,6 +1060,30 @@ def validate_current_widgets(event)
...
     publish function: :InitNull, type: "void (string)"
     publish function: :StoreNull, type: "void (string, map)"
+
+  private
+
+    def widgets_in_contents(contents)
+      contents.each_with_object([]) do |arg, res|
+        case arg
+        when ::CWM::AbstractWidget then res << arg
+        when Yast::Term then res.concat(widgets_in_contents(arg))
```
mvidner: "Undefined method `each_with_object` in Yast::Term" when I run examples/object_api.rb

jreidinger: old ruby bindings. It is quite new feature of term

library/cwm/test/widgets_test.rb
jreidinger: I know, failed when I have to go yesterday, will fix.

library/cwm/src/modules/CWM.rb
```diff
@@ -1026,6 +1060,30 @@ def validate_current_widgets(event)
     publish function: :ShowAndRunOrig, type: "symbol (list <string>, map <string, map <string, any>>, term, string, string, string, map)"
     publish function: :InitNull, type: "void (string)"
     publish function: :StoreNull, type: "void (string, map)"
+
+  private
+
+    def widgets_in_contents(contents)
```
mvidner: If I add this line, the examples work again
```rb
      contents = contents.params if contents.is_a? Yast::Term
```

jreidinger:
@imobach @ancorgs interesting idea, but I still think that it should be together, as it is related code as help describe behavior and label is also related with functionality. Current bootloader have it separated and I am not much happy about it as when you change something or use new widget you have to:

1) write label and help in one file
2) write behavior in second file
3) at third location write description of widget
4) use it in another file in CWM content

Consider also that dialogs in bootloader have a lot of widget and that it depends on architecture and EFI usage. With your approach it can end up like

```
input = InputField.new(
  widget_id: "my_input_field",
  label: _("My label"), 
  help: _("Help..."), # helps are multiline often
  behavior: MyWidgetBehavior.new(...) # behavior often need another widget or know another widget id, so we have to pass it
)
# and repeat such line e.g. 30 times...so in the end I separate it to own methods like
def kernel_append_widget
  InputField.new(
 ....
)
end
# and in such case I again have three location when I have to use new widget 1) behavior 2) rest of widget 3) it usage in dialog, but with subclassing I need only two create new class and then use it
```

So to summary it, splitting behavior and instancing widget for me does now make sense alone as you cannot reuse behavior itself and it do not make sense to split it from related labels and help.

@ancorgs consider you will create widget for Service reload as we have it now in YaST2 for CWM based usage.
Do you think it is easier to use in dialogs that use it as
```
CustomWidget.new(
  contents: ServiceStatus.contents,
  help: ServiceStatus.help, #or write it again
  label: ..
  behavior: ServiceStatus.behavior(service_name) ... # or write own or subclass
)
```
 or with subclassign independent widget
```
ServiceStatusWidget.new(service_name) # and when needed to modify then subclass it
```

consider that there is 20 modules that use service status and CWM together and what is easier to use? This then say which way is easier to use and for having independent reusable parts of code

library/cwm/src/modules/CWM.rb
jreidinger: but it is not correct for more complex contents as you do not do searching in deep. As I write below problem is old ruby bindings...just install newer or at least use newer term code - https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/term.rb#L14

mvidner: I see, Term being Enumerable is from 2015-01 but I run 13.2. Sorry for the noise!

library/cwm/examples/object_api.rb
```diff
@@ -0,0 +1,70 @@
...
+Yast.import "CWM"
+Yast.import "Wizard"
+
+class LuckyNumberWidget < CWM::IntField
+  attr_reader :result, :minimum, :maximum
+
+  def initialize
+    @minimum = 0
+    @maximum = 1000
+    self.widget_id = "lucky_number"
```
mvidner: Optional improvement: use `self.class.to_s` by default, maybe also converting Yast::FooBar to foo_bar.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,565 @@
...
+  #
+  #     def init(widget)
+  #       ...
+  #     end
+  #
+  #     def store(widget, event)
+  #       ...
+  #     end
+  #
+  #     def validate(widget, event)
```
mvidner: We should also update the callback signatures in these examples to (not) include `widget` and `event` according to the final API.

library/cwm/examples/object_api.rb
jreidinger: hmm, yes good idea, in general you usually do not need widget_id as long as it is unique

library/cwm/src/lib/cwm/widget.rb
jreidinger: yes, thanks for notice

ancorgs:
@jreidinger Ok, you convinced me. Let's continue with the current "subclass everything" approach.

imobach:
I see your point. If behavior are not going to be reusable, it makes sense following the subclass approach.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    # - `#init` [nil ()] initialize widget like e.g. set initial value
+    # - `#handle` [Symbol,nil (String?)] handle widget changed value or press.
+    #   return value is usually nil, returning symbol can be used to send different event.
+    #   It support varient with parameter specified and without. If parameter is specified,
+    #   gets event Hash including "ID" key with value that specify widget ID that cause event.
+    #   It is mainly useful if handle_all_events is set to true to distinguish which event caused it.
+    # - `#store` [nil ()] store widget value after user confirm dialog
+    # - `#cleanup` [nil ()] cleanup after widget is destroyed
+    # @raise [RuntimeError] if required method is not implemented or widget id not set.
+    def description
```
mvidner: API: In UI context I understand a "description" as a "long label". Should we rename this to `definition` or even `cwm_definition` ? Or mimic its hash key and name it  `widget_descr`?

jreidinger: `cwm_definition` sounds good for me

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    # generates description for CWM with widget. Description is auto filled from defined methods.
+    #
+    # methods used to generate description:
+    #
+    # - `#help` [String] to get translated help text for widget
+    # - `#label` [String] to get translated label text for widget
+    # - `#opt` [Array<Symbol>] to get options passed to widget like `[:hstretch, :vstretch]`
+    # - `#validate` [Boolean ()] validate widget value. Returns false if validation failed.
+    # - `#init` [nil ()] initialize widget like e.g. set initial value
+    # - `#handle` [Symbol,nil (String?)] handle widget changed value or press.
```
mvidner: This line of the docs illustrates why I think we are better off declaring all of these as methods of AbstractWidget and write proper Yard comments for them. You simply cannot document a whole method signature in a single Yard type.

jreidinger: but it duplicates default values which I would like avoid, especially when cwm use for some widgets kinds specific defaults

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+  class AbstractWidget
+    include Yast::UIShortcuts
+    include Yast::I18n
+
+    # @return [String] id used for widget
+    attr_writer :widget_id
+    attr_writer :handle_all_events
+
+    # specify if widget handle all raised events or only its own
+    # By default only own values are handled
```
mvidner: > \# By default, {#handle} has no argument and it is called only for events of its own widget.
\# If `true`, {#handle}(event) is called for events of any widget.
\# @return [Boolean]

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+  #   end
+  #  # and usage in dialog
+  #  widget = MyWidget.new
+  # ```
+  #
+  class AbstractWidget
+    include Yast::UIShortcuts
+    include Yast::I18n
+
+    # @return [String] id used for widget
```
mvidner: >\# By default, the class name is used.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    def self.widget_type=(val)
+      define_method(:widget_type) { val }
+    end
+
+    # generates description for CWM with widget. Description is auto filled from defined methods.
+    #
+    # methods used to generate description:
+    #
+    # - `#help` [String] to get translated help text for widget
+    # - `#label` [String] to get translated label text for widget
```
mvidner: `label` is the only abstract method, the others have defaults, right?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    # - `#opt` [Array<Symbol>] to get options passed to widget like `[:hstretch, :vstretch]`
+    # - `#validate` [Boolean ()] validate widget value. Returns false if validation failed.
+    # - `#init` [nil ()] initialize widget like e.g. set initial value
+    # - `#handle` [Symbol,nil (String?)] handle widget changed value or press.
+    #   return value is usually nil, returning symbol can be used to send different event.
```
mvidner: I thought that returning non-nil from `handle` means to return with that value from `CWM::ShowAndRun`

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    end
+
+    # generates description for CWM with widget. Description is auto filled from defined methods.
+    #
+    # methods used to generate description:
+    #
+    # - `#help` [String] to get translated help text for widget
+    # - `#label` [String] to get translated label text for widget
+    # - `#opt` [Array<Symbol>] to get options passed to widget like `[:hstretch, :vstretch]`
+    # - `#validate` [Boolean ()] validate widget value. Returns false if validation failed.
```
mvidner: There are no args for `validate` and it is supposed to call `value` itself, right?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    # TODO: kill converts in CWM module, to avoid this workaround for funrefs
+    def fun_ref(*args)
+      Yast::FunRef.new(*args)
+    end
+
+  private
+
+    # @note all methods here use wrappers to modify required parameters as CWM
+    # have not so nice callbacks API
+    def init_method
```
mvidner: Non-API nitpicking, but it irritates me :)
`def init_fun_ref`
One day it may return a Method, but now it takes a Method and turns it into a FunRef, so it should have a name that does not lie.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+  #       case event["ID"]
+  #       when :reset then ...
+  #       when :undo then ...
+  #       else ...
+  #       end
+  #     end
+  #   end
+  class CustomWidget < AbstractWidget
+    self.widget_type = :custom
+    # custom witget without contents do not make sense
```
mvidner: But we need to document `contents`. Just declaring it abstract is not enough.
>\# @return [Term] ...

Is the term allowed to contain AbstractWidgets, or just plain UI terms?

library/cwm/src/lib/cwm/widget.rb
jreidinger: yes

library/cwm/src/lib/cwm/widget.rb
jreidinger: well, I need to try it. I do not know exact behavior, as I do not use it myself.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    end
+  end
+
+  # Represents custom widget, that have its UI content defined in method content.
+  # Useful mainly when specialized widget including more subwidget should be
+  # reusable at more places.
+  #
+  # @example custom widget child
+  #   class MyWidget < CWM::CustomWidget
+  #     def initialize
```
mvidner: The whole method unneeded since we have a default for widget_id.

library/cwm/src/lib/cwm/widget.rb
jreidinger: I write it above in example....I think it do not allow to contain AbstractWidget as it is processed later.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+      Yast::UI.ChangeWidget(Id(widget_id), :Items, val)
+    end
+  end
+
+  # Represents input field widget. `label` method is mandatory.
+  #
+  # @example input field widget child
+  #   class MyWidget < CWM::InputFieldWidget
+  #     def initialize(myconfig)
+  #       self.widget_id = "my_widget"
```
mvidner: zap!

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    end
+  end
+
+  # Widget representing button.
+  #
+  # @example push button widget child
+  #   class MyEvilWidget < CWM::PushButtonWidget
+  #     def initialize
+  #       self.widget_id = "my_evil_widget"
+  #     end
```
mvidner: and this one too

library/cwm/src/lib/cwm/widget.rb
mvidner: Hmm. Let's ask others which variant they prefer.

library/cwm/src/lib/cwm/widget.rb
jreidinger: Thinking about it, it is possibleto add processing of AbstractWidget there, but I prefer to do it separatelly when needed, this pr tend to be quite long and I would like to finish it

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+
+    include ValueBasedWidget
+    abstract_method :label
+
+    # @return [Boolean] true if widget is checked
+    def checked?
+      value
+    end
+
+    # @return [Boolean] true if widget is unchecked
```
mvidner: YUI also supports tristate checkboxes: <https://doc.opensuse.org/projects/YaST/openSUSE11.3/tdg/CheckBox.html>

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+      Yast::UI.ChangeWidget(Id(widget_id), :SelectedItems, val)
+    end
+  end
+
+  # Represents integer field widget. `label` method is mandatary. It supports
+  # additional `minimum` and `maximum` method for limiting selection.
+  # @see #{.cwm_definition} method for minimum and maximum example
+  #
+  # @see InputFieldWidget for example of child
+  class IntField < AbstractWidget
```
mvidner: All the other class names end with "Widget". Should we drop all the suffixes?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+  #     def handle(widget, _event)
+  #       return if widget != widget_id
+  #
+  #       Virus.install
+  #
+  #       nil
+  #     end
+  #   end
+  class PushButtonWidget < AbstractWidget
+    self.widget_type = :push_button
```
mvidner:
```
abstract_method :label
```
also here.

mvidner:
Good! I will make another pass to clarify the documentation comments, but for me, the API is almost done:
- Should all widget classes be named Foo or FooWidget?
- Do we support tristate CheckBox or not?
- What does the return value of `handle` mean?

jreidinger:
@mvidner good question, I think we are in CWM namespace, so we can drop Widget suffix except AbstractWidget one, do you agree?

I do not know there is tri-state checkbox, what is its purpose?

I will try handle with return value to check what happens.

jreidinger:
@mvidner I checked handle and really if result is not nil, it finish CWM.show with that result immediatelly

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+  # @see InputFieldWidget for example of child
+  class CheckboxWidget < AbstractWidget
+    self.widget_type = :checkbox
+
+    include ValueBasedWidget
+    abstract_method :label
+
+    # @return [Boolean] true if widget is checked
+    def checked?
+      value
```
mvidner: value == true

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,568 @@
...
+    end
+
+    # @return [Boolean] true if widget is unchecked
+    def unchecked?
+      !value
```
mvidner: value == false

mvidner:
Yes, drop the Widget suffix.

If we modify the CheckBox accessors like I suggested we can defer the question :)

mvidner:
I have improved the documentation.
Version and changelog need to be added.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -0,0 +1,631 @@
...
+  #       _("Win the lottery by clicking this.")
+  #     end
+  #
+  #     def handle
+  #       Virus.install
+  #       nil
+  #     end
+  #   end
+  class PushButton < AbstractWidget
+    self.widget_type = :push_button
```
mvidner: If `abstract_method :label` is omitted on purpose, please add a comment why. It seems like a mistake to me.

mvidner:
3.1.164, changes:
> Added CWM::AbstractWidget.

Then LGTM.

mvidner:
Widgets for lunch, widgets for dinner, widgets for breakfast!
LGTM already :smile: !

Delicious widgets, thank you!

mvidner:
@jreidinger @lslezak Now I found out that Travis fails, not for this repo but for [others](https://travis-ci.org/yast/yast-network/builds/104104240) and I *think* it is because abstract_method.gem is not installed. (Hard to know the actual cause. Travis.)

mvidner:
Yes, installing it fixed the Travis build for https://github.com/yast/yast-network/pull/381 .

jreidinger:
@mvidner so what is solution? install it everytime for travis? as we do not have dependency checking of dependencies

jreidinger:
@mvidner or maybe add it to ubuntu dpkg?

mvidner:
@jreidinger : @lslezak has already added it to yast2.deb on Friday but we have waited over the weekend for the repo to be published. I am rebuilding https://travis-ci.org/yast/yast-registration now to see whether it is fixed.

PR #412, opened 2015-11-17
Set default value for FW_BOOT_FULL_INIT

imobach:
Fixes [bsc#955400](https://bugzilla.suse.com/show_bug.cgi?id=955400).

PR #422, opened 2016-01-18
Set a default value for FW_BOOT_FULL_INIT

imobach:
This fix is taken from SLE-12-GA. As it's the only difference, I've also merged the SLE-12-GA branch into SLE-12-SP1.


mvidner:
(This is a port of #412)
LGTM

imobach:
Thanks @mvidner!

PR #425, opened 2016-01-20
CWM Tabs

jreidinger:


library/cwm/src/modules/CWM.rb
```diff
@@ -1061,11 +1065,10 @@ def validate_current_widgets(event)
     publish function: :InitNull, type: "void (string)"
     publish function: :StoreNull, type: "void (string, map)"
 
-  private
-
     def widgets_in_contents(contents)
       contents.each_with_object([]) do |arg, res|
         case arg
+        when ::CWM::CustomWidget then res.concat(arg.nested_widgets) << arg
```
mvidner: This method shouldn't do it. Instead, define AbstractWidget#widget_ids which by default returns just `[widget_id]` and the nested ones for CustomWidget.
```rb
if arg.respond_to? :widget_ids
  res.concat(arg.widget_ids)
else
  res.concat(widgets_in_contents(arg))
end

Comment (mvidner) [inline library/cwm/src/modules/CWM.rb]: Ugh, it should be the Widget objects, not the IDs, sorry.

Comment (jreidinger) [inline library/cwm/src/modules/CWM.rb]: your `else` cannot work, as in content is a lot of other objects like strings, numbers, etc. So we need detect term and do nesting.
Regarding widgets it is possible to add it.

File: library/cwm/examples/object_api_tabs.rb [hunk 2]
```diff
@@ -0,0 +1,153 @@
...
+    ]
+  end
+
+  def init
+    self.value = @chosen if @chosen
+  end
+
+  def validate
+    if value == :pc
+      Yast::Popup.Error(_("Human will be exterminated, pc cannot allow you to be your true love"))
```
mvidner: I think you will get true love of translators once it reaches them :)

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -285,14 +285,27 @@ class CustomWidget < AbstractWidget
...
+    # Returns all nested widgets used in contents
+    def nested_widgets
+      Yast.import "CWM"
+
+      Yast::CWM.widgets_in_contents(contents)
+    end
+
+  protected
+
+    def cwm_contents
```
mvidner: So what is the doc comment for `cwm_contents`, compared to `contents`?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -630,4 +643,182 @@ class RichText < AbstractWidget
 
     include ValueBasedWidget
   end
+
+  # Tab widget, usefull only with {CWM::Tabs}
+  # @see tabs example for usage
+  class Tab < CustomWidget
+    attr_accessor :initial
```
mvidner: \# @return [Boolean] is this the initially selected tab

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -630,4 +643,182 @@ class RichText < AbstractWidget
...
+
+    # content of tab, can contain another widgets
```
mvidner: \# @return [Yast::Term] contents of the tab, can contain {AbstractWidget}s

Can it be a single AbstractWidget too?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -630,4 +643,182 @@ class RichText < AbstractWidget
...
+    end
+
+  protected
+
+    # defines name of tab header
+    # It is recommended to rewrite it
+    def label
+      textdomain "base"
+
+      _("Tab")
```
mvidner: Oh no. Just leave it an abstract method.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -630,4 +643,182 @@ class RichText < AbstractWidget
...
+    end
+  end
+
+  # useful to have tabs as widget. It contained {CWM::Tab} with its content
+  # @see examples/object_api_tabs.rb
+  class Tabs < CustomWidget
+    # param [Array<CWM::Tab>] tabs to be shown
+    def initialize(*tabs)
+      @tabs = tabs
+      @current_tab = nil
```
mvidner: \# CWM::Tab

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -630,4 +643,182 @@ class RichText < AbstractWidget
...
+      self.handle_all_events = true
+    end
+
+    # gets visual order of tabs
```
mvidner: does this need to be public?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -630,4 +643,182 @@ class RichText < AbstractWidget
...
+    def store
+      store_tab(@current_tab.widget_id)
+    end
+
+    # validates current tab
+    def validate
+      Yast::CWM.validateWidgets(@current_tab.cwm_definition["widgets"], "ID" => @current_tab.widget_id)
+    end
+
+    # stores tab with given id
```
mvidner: Does `store_tab` need to be public?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -630,4 +643,182 @@ class RichText < AbstractWidget
...
+          Id(tab.widget_id),
+          :Label,
+          "#{Yast::UI.Glyph(:BulletArrowRight)}  #{tab.label}"
+        )
+      end
+    end
+
+    # gets id of initial tab
+    # This default implementation returns first tab from {#tabs} method
+    def initial_tab
```
mvidner: Rename to `initial_tab_id`

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -630,4 +643,182 @@ class RichText < AbstractWidget
...
+      initial = @tabs.find(&:initial)
+
+      (initial || @tabs.first).widget_id
+    end
+
+    def contents
+      textdomain "base"
```
mvidner: why?

library/cwm/src/lib/cwm/widget.rb
jreidinger: difference is that cwm_contents is content converted to format that CWM understand ( so widgets instance replaced by ids widget_id as string )

library/cwm/src/lib/cwm/widget.rb
jreidinger: you mean passing widget instead of term? I think it is not allowed

library/cwm/src/lib/cwm/widget.rb
jreidinger: it is compatible with CWMTab, but I am fine with having it as abstract one

library/cwm/src/lib/cwm/widget.rb
jreidinger: I don't get this comment

library/cwm/src/lib/cwm/widget.rb
jreidinger: it is a bit mistery, but when I move it to private it stop working without anything in logs

library/cwm/src/lib/cwm/widget.rb
jreidinger: this one probably not

library/cwm/src/lib/cwm/widget.rb
jreidinger: ah, not needed, right. There was `_("Tab")`

library/cwm/examples/object_api_tabs.rb
jreidinger: even translators need some fun :)

mvidner:
Thanks! I still see some room for improvement, but since this is blocking your other work and the API is good, I will make the improvements myself in a new PR.
LGTM.

jreidinger:
@mvidner thanks

PR #426, opened 2016-01-21
Count test coverage more accurately

mvidner:
and also fix an API bug in the process; need to verify that the fix is harmless

Coverage goes down from 56% to 45%, yay!

### DIFF
--- a/library/general/src/lib/installation/proposal_client.rb
+++ b/library/general/src/lib/installation/proposal_client.rb
@@ -89,7 +89,7 @@ module Installation
       when "Description"
         description
       when "Write"
-        write(param)
+        write
       else
         raise ArgumentError, "Invalid action for proposal '#{func.inspect}'"
       end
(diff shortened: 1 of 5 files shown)


jreidinger:
LGTM, just make rubocop happy: https://travis-ci.org/yast/yast-yast2/builds/103880372#L563

jreidinger:
LGTM

library/general/src/lib/installation/proposal_client.rb
```diff
@@ -270,5 +270,9 @@ def ask_user(_attrs)
     def description
       raise NotImplementedError, "Calling abstract method 'description'"
     end
+
+    def write
+      # doing nothing is OK. or is it? clarify the API, the docs, actual usage!
+    end
```
jreidinger: I worry it is not OK. If you really call it with write, then you expect it do something.

jreidinger: so possible to overlook bugs. Actual usage is without write, only in very specific modules like used to be network or cio proposal use it as it keep value, so finish client call proposal client with write.

PR #427, opened 2016-01-25
New UI::InstallationDialog class

ancorgs:
The unit tests are still missing. I'll update the PR tomorrow.

### DIFF
--- a/library/general/src/lib/ui/installation_dialog.rb
+++ b/library/general/src/lib/ui/installation_dialog.rb
@@ -97,7 +97,7 @@ module UI
   protected
 
     # Optional title icon
-    # 
+    #
     # @return [String] name of the icon to use
     def title_icon
     end
--- a/library/general/src/lib/ui/installation_dialog.rb
+++ b/library/general/src/lib/ui/installation_dialog.rb
@@ -41,6 +41,10 @@ module UI
   #       "Foo"
   #     end
   #
+  #     def help_text
+  #       "<p>Set the value for foo here.</p>
+  #     end
+  #
   #     def next_handler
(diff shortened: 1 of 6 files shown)


library/general/src/lib/ui/installation_dialog.rb
```diff
@@ -0,0 +1,139 @@
...
+    def dialog_title
+      ""
+    end
+
+    # Text to display when the help button is pressed
+    #
+    # @return [String]
+    def help_text
+      ""
+    end
```
jreidinger: are those two optional? if not, then you can use abstract method gem we already depend on

library/general/src/lib/ui/installation_dialog.rb
```diff
@@ -0,0 +1,139 @@
...
+
+    # Reimplementation of UI::Dialog#create_dialog for the installer
+    #
+    # It ignores UI::Dialog#dialog_options
+    def create_dialog
+      if dialog_options
+        log.info "This is an InstallationDialog, ignoring #dialog_options"
+      end
+
+      Yast::Wizard.CreateDialog if Yast::Mode.normal # allow manual testing
```
jreidinger: I know it is often used in dialog, but I found it quite hacky. I think better approach is to use `Wizard.IsWizardDialog` and only if it is not wizard, open it ( and of course store that you create it and close it when needed.

library/general/src/lib/ui/installation_dialog.rb
ancorgs: I wouldn't do them mandatory. It's fine if you want to have a dialog with empty help or empty headline.

jreidinger:
LGTM

PR #429, opened 2016-01-26
Skip store for

jreidinger:


library/cwm/src/modules/CWM.rb
```diff
@@ -917,8 +919,11 @@ def SetValidationFailedHandler(handler)
     # @param [String] back_button label for dialog back button
     # @param [String] next_button label for dialog next button
     # @param [String] abort_button label for dialog abort button
+    # @param [Array] skip_store_for list of events that skip storring of widgets.
```
ancorgs: "storring"... twice

library/cwm/src/modules/CWM.rb
```diff
@@ -779,8 +779,10 @@ def handleDebug
     # @param [Array<Hash{String => Object>}] widgets list of widget maps
     # @param [Hash] functions map initialize/save/handle fallbacks if not specified
     #   with the widgets.
+    # @param [Array<Object>] skip_store_for list of events that is cause to not store widgets
+    #   Useful mainly for non-standard redraw of widgets without storing like :reset or :redraw
```
ancorgs: list of events for which the value of the widget will not be stored
Useful mainly for non-standard redraw of widgets, like :reset or :redraw

library/cwm/src/modules/CWM.rb
```diff
@@ -917,8 +919,11 @@ def SetValidationFailedHandler(handler)
     # @param [String] back_button label for dialog back button
     # @param [String] next_button label for dialog next button
     # @param [String] abort_button label for dialog abort button
+    # @param [Array] skip_store_for list of events that skip storring of widgets.
+    #   Useful mainly when some widget return event that skip storring like
+    #   "reset" button or redrawing without storing
```
ancorgs: list of events for which the value of the widget will not be stored.
Useful mainly when some widget returns an event that should not trigger the storing, like a reset button or a redrawing

ancorgs:
Good code-wise. Just some English proposal (not really good but hopefully better than the originals).

ancorgs:
LGTM

jreidinger:
thanks

PR #432, opened 2016-02-01
Fix execute

jreidinger:


library/system/src/lib/yast2/execute.rb
```diff
@@ -62,6 +63,7 @@ def popup_error(&block)
         block.call
       rescue Cheetah::ExecutionFailed => e
         Yast.import "Report"
+        textdomain "base"
```
lslezak: NTH: I'd put the `textdomain` call into the main initialization code so you do not need to add another `textdomain` call later when adding a new translatable string somewhere else...

jreidinger: there is no initialization as it is class method

lslezak: OK, I meant just after the line `include Yast::I18n`...

library/system/src/lib/yast2/execute.rb
```diff
@@ -33,6 +33,7 @@ class Execute
     Cheetah.default_options = { logger: Y2Logger.instance }
 
     class << self
+      include Yast::I18n
```
lslezak: Does it work? For using `_()` in class methods you have to use `extend` instead of `include`.

This works for me:
```ruby
require "yast"
class Foo
    extend Yast::I18n

    textdomain "foo"

    def self.foo
        puts _("foo")
    end
end

Foo.foo
```

library/system/src/lib/yast2/execute.rb
```diff
@@ -32,48 +32,50 @@ class Execute
...
-            stderr:   e.stderr
-          }
-        )
-      end
+    def self.popup_error(&block)
+      block.call
+    rescue Cheetah::ExecutionFailed => e
+      Yast.import "Report"
+      textdomain "base"
+      Yast::Report.Error(
```
kobliha: Please, add also log.error - Report.Error is always localized

library/system/src/lib/yast2/execute.rb
```diff
@@ -32,48 +32,50 @@ class Execute
...
-            exitcode: e.status.exitstatus,
-            stderr:   e.stderr
-          }
-        )
-      end
+    def self.popup_error(&block)
+      block.call
+    rescue Cheetah::ExecutionFailed => e
+      Yast.import "Report"
+      textdomain "base"
```
lslezak: This `textdomain` call is not needed after the rewrite I guess...

library/system/src/lib/yast2/execute.rb
jreidinger: it is automatic logged by Cheetah, thanks why I set cheetah logger to yast one.

kobliha: I see, that's nice :)

lslezak:
LGTM, thanks!

BTW it would be nice to have an unit test here, the `include`/`extend` issue would have been easily found...

jreidinger:
yeah, I know. I hope there wasn't any bug in such thin layer :)

lslezak:
LGTM, thanks for test! :+1:

PR #433, opened 2016-02-03
Cleanup NetworkInterfaces.Read (bsc#982850)

teclator:
In general the goal of this PR is to improve the readability of ```NetworkInterfaces.Read``` and also take the opportunity to clean up some code.

https://trello.com/c/FXPO8MLz


library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -236,21 +237,21 @@ def device_type(dev)
 
     # Detects a subtype of Ethernet device type according /sys or /proc content
     def GetEthTypeFromSysfs(dev)
-      sys_dir_path = Builtins.sformat("/sys/class/net/%1/", dev)
+      sys_dir_path = "/sys/class/net/#{dev}/"
 
-      if FileUtils.Exists(Ops.add(sys_dir_path, "wireless"))
+      if FileUtils.Exists("#{sys_dir_path}wireless")
         return "wlan"
```
imobach: Just a matter of taste, but I'd remove the 'return' keyword.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -236,25 +237,25 @@ def device_type(dev)
...
-      elsif FileUtils.Exists(Ops.add(sys_dir_path, "bonding"))
-        return "bond"
-      elsif FileUtils.Exists(Ops.add(sys_dir_path, "tun_flags"))
-        return "tap"
-      elsif FileUtils.Exists(Ops.add("/proc/net/vlan/", dev))
-        return "vlan"
-      elsif FileUtils.Exists(Ops.add("/sys/devices/virtual/net/", dev)) &&
-          Builtins.regexpmatch(dev, "dummy.*")
-        return "dummy"
+      sys_dir_path = "/sys/class/net/#{dev}/"
```
mchf: NP: I'd prefer not to put trailing '/'. It is simply my personal preference to read subsequent usages as ```#{sys_dir_path}/whatever_file```

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -263,11 +264,11 @@ def GetIbTypeFromSysfs(dev)
       sys_dir_path = Builtins.sformat("/sys/class/net/%1/", dev)
 
       if FileUtils.Exists(Ops.add(sys_dir_path, "bonding"))
```
mchf: Why not to continue with conversion in the same way as above? I mean turn it into "#{sys_dir_path}/bondin" as you did above.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -599,30 +600,22 @@ def CanonicalizeIP(ifcfg)
       ifcfg = deep_copy(ifcfg)
       return nil if ifcfg.nil?
 
-      ip_and_prefix = Builtins.splitstring(
-        Ops.get_string(ifcfg, "IPADDR", ""),
-        "/"
-      )
-      ipaddr = Ops.get(ip_and_prefix, 0, "")
-      return deep_copy(ifcfg) if ipaddr == "" # DHCP or inconsistent
+      ipaddr, prefixlen = ifcfg["IPADDR"].to_s.split("/")
```
mchf: well, if you do
```(ifcfg["IPADDR"] || "/").to_s.split("/", -1)```
you can avoid those ```to_s``` further. It's on your decision if that is improvement.

File: library/network/src/modules/NetworkInterfaces.rb [hunk 5]
```diff
@@ -599,30 +600,22 @@ def CanonicalizeIP(ifcfg)
...
+      return ifcfg if ipaddr.to_s == "" # DHCP or inconsistent
 
-      if prefixlen == ""
-        prefixlen = Builtins.tostring(
-          Netmask.ToBits(Ops.get_string(ifcfg, "NETMASK", ""))
-        )
-      end
+      prefixlen = ifcfg["PREFIXLEN"].to_s if prefixlen.to_s == ""
+
+      prefixlen = Netmask.ToBits(ifcfg["NETMASK"].to_s).to_s if prefixlen == ""
```
mchf: I think that ```Netmask.ToBits``` always returns a number so the last ```to_s``` is useless

library/network/src/modules/NetworkInterfaces.rb
teclator: (ifcfg["IPADDR"] || "/") or ifcfg["IPADDR"].to_s is almost the same

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -236,38 +237,38 @@ def device_type(dev)
 
     # Detects a subtype of Ethernet device type according /sys or /proc content
     def GetEthTypeFromSysfs(dev)
```
imobach: Please, document the `dev` parameter.

library/network/src/modules/NetworkInterfaces.rb
imobach: Or maybe `ifcfg.fetch("IPADDR", "/").split("/")` but for me it's fine as it is.

library/network/src/modules/NetworkInterfaces.rb
imobach: It depends if `prefixlen` should be a number or a string.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -652,11 +639,7 @@ def filter_interfacetype(devmap)
     def ConcealSecrets1(ifcfg)
       ifcfg = deep_copy(ifcfg)
       return nil if ifcfg.nil?
-      out = Builtins.mapmap(ifcfg) do |k, v|
-        v = "CONCEALED" if Builtins.contains(@SensitiveFields, k) && v != ""
-        { k => v }
-      end
-      deep_copy(out)
+      ifcfg.each { |k, v| ifcfg[k] = "CONCEALED" if @SensitiveFields.include?(k) && v != "" }
```
imobach: You could split this part in:

1) finding what `ifcfg` elements should be `concealed`
2) setting `concealed` value

Anyway, if you want to keep it as it is, I would avoid the "if" at the end.

Just a suggestion.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -681,6 +664,38 @@ def ConcealSecrets(devs)
...
+      else
+        devices = allfiles.select { |file| file !~ /#{devregex}/ }
+      end
+      log.debug "devices=#{devices}"
+      devices
+    end
+
+    # Canonicalize IPADDR and STARTMODE of given config
+    # and nested _aliases
+    # @return
```
imobach: Please, fix `@return` documentation.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -681,6 +664,38 @@ def ConcealSecrets(devs)
...
+    def canonicalize_config(config)
+      config = deep_copy(config)
+      # canonicalize, #46885
+      caliases = (config["_aliases"] || {}).tap do |h|
```
imobach: np: use more than 1-letter identifiers when using the `do...end` form.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -681,6 +664,38 @@ def ConcealSecrets(devs)
...
+        h.map do |a, c|
```
imobach: Why a `map`?

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -681,6 +664,38 @@ def ConcealSecrets(devs)
...
+          h[a] = CanonicalizeIP(c)
+        end
+      end
+      config["_aliases"] = caliases if caliases != {} # unconditionally?
```
imobach: Maybe `unless caliases.empty?`?

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -693,79 +708,78 @@ def ConcealSecrets(devs)
       "IP_OPTIONS"
     ]
 
-    # Read devices from files
+    def generate_config(pth, values)
```
imobach: Please, doc :)

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -693,79 +708,78 @@ def ConcealSecrets(devs)
...
+      config = {}
+      values.each do |val|
+        item = SCR.Read(path("#{pth}.#{val}"))
+        log.debug("item=#{item}")
+        next if item.nil?
+        # No underscore '_' -> global
+        # Also temporarily standard globals
+        if !val.include?("_") || LOCALS.include?(val)
```
imobach: A `_` in any positions? Or only at the start?

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -693,79 +708,78 @@ def ConcealSecrets(devs)
...
+          config["_aliases"][s] ||= {}
+          config["_aliases"][s][v] = item
+        end
+      end
+      log.info("config=#{ConcealSecrets1(config)}")
+      config = canonicalize_config(config)
+      filter_interfacetype(config)
+    end
+
+    def set_devtype_config(device, config)
```
imobach: Please, doc :)

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -681,6 +664,38 @@ def ConcealSecrets(devs)
       deep_copy(out)
     end
 
+    # Get current sysconfig configured interfaces
+    #
+    # @param [String] regex to filter by
```
imobach: I would add argument names: `@param devregex [String] regex to filter by`

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -777,20 +791,13 @@ def CleanCacheRead
...
-        Ops.add("^(", Ops.get(@DeviceRegex, devregex, devregex)),
-        ")[0-9]*$"
-      )
-      Builtins.y2debug("regex=%1", regex)
-      devices = Builtins.filter(devices) do |file, _devmap|
-        Builtins.regexpmatch(file, regex) == true
-      end
-      Builtins.y2debug("devices=%1", devices)
-      deep_copy(devices)
+      regex = "^(#{@DeviceRegex[devregex] || devregex})[0-9]*$"
```
imobach: np: I prefer `\A` and `\z` over `^` and `$`.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1299,73 +1300,66 @@ def GetDevTypeDescription(type, longdescr)
     # @return dumped settings (later acceptable by Import())
     def Export(devregex)
       devs = Filter(@Devices, devregex)
```
imobach: You could replace `Filter` by Ruby `select`.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1299,73 +1300,66 @@ def GetDevTypeDescription(type, longdescr)
...
-          Map.Keys(Ops.get(@Devices, type, {})),
-          from: "list",
-          to:   "list <string>"
-        )
-      ) do |dev|
-        dev = device_num(dev) if Builtins.issubstring(dev, type)
-        curdevs = Builtins.add(curdevs, dev)
-      end
+      curdevs = @Devices.fetch(type, {}).keys
+      curdevs = curdevs.map { |d| d.include?(type) ? device_num(d) : d }
```
imobach: np: You could use a `map!` or a `each_with_object` if you prefer. Feel free to ignore this suggestion :)

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1299,73 +1300,66 @@ def GetDevTypeDescription(type, longdescr)
...
+        ret << ""
       end
 
       # Remaining numbered devices
-      while Ops.less_than(count, num)
-        ii = Builtins.sformat("%1", i)
-        if !Builtins.contains(curdevs, ii)
-          ret = Builtins.add(ret, ii)
-          count = Ops.add(count, 1)
+      while count < num
```
imobach: I doesn't look like proper ruby. What about using `(count...num).each`? You could also get rid of the `i` counter.

imobach: Well, that won't work: `(num - count).times` sounds better.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -693,79 +708,78 @@ def ConcealSecrets(devs)
...
       end
-      Builtins.y2debug("Devices=%1", @Devices)
+      log.debug("Devices=#{@Devices}")
 
       @OriginalDevices = deep_copy(@Devices)
       @initialized = true
-      true
+    end
+
+    def Reset
```
mchf: I'm not that sure that creating method which is used only in test is good idea. Or did I overlooked something again?

library/network/src/modules/NetworkInterfaces.rb
mchf: and I prefer ```^, $``` over ```\A, \z``` ;-)

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -777,20 +791,13 @@ def CleanCacheRead
...
-      )
-      Builtins.y2debug("regex=%1", regex)
-      devices = Builtins.filter(devices) do |file, _devmap|
-        Builtins.regexpmatch(file, regex) == true
-      end
-      Builtins.y2debug("devices=%1", devices)
-      deep_copy(devices)
+      regex = "^(#{@DeviceRegex[devregex] || devregex})[0-9]*$"
+      log.debug("regex=#{regex}")
+      devices.select! { |f, _d| f =~ /#{regex}/ }
```
mchf: probably, we may drop ```deep_copy``` from the beginning of the method and use ```select``` here

mchf: while ```f``` ?

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -799,48 +806,42 @@ def FilterDevices(devregex)
...
-      Builtins.y2debug("regex=%1", regex)
-      devices = Builtins.filter(devices) do |file, _devmap|
-        Builtins.regexpmatch(file, regex) != true
-      end
-      Builtins.y2debug("devices=%1", devices)
-      deep_copy(devices)
+      regex = "^(#{@DeviceRegex[devregex] || devregex})[0-9]*$"
+
+      log.debug("regex=#{regex}")
+      devices.select! { |f, _d| f !~ regex }
```
mchf: ok so we have two methods which are almost the same but one operator. Can it be somehow consolidated?

library/network/src/modules/NetworkInterfaces.rb
teclator: It was ```file``` originally so the natural change is to use ```f``` but we could discuss if maybe is better to use ```n``` for name device or d for device and rename ```_d``` as ```_m```

library/network/src/modules/NetworkInterfaces.rb
imobachgs: Maybe calling `main` would do the trick.

library/network/src/modules/NetworkInterfaces.rb
teclator: Exactly

library/network/src/modules/NetworkInterfaces.rb
teclator: Can be in the middle

library/network/src/modules/NetworkInterfaces.rb
teclator: I will maintain the current expression.

library/network/src/modules/NetworkInterfaces.rb
teclator: In that case the algorithm should be changed because there are some iterations without assignment because the interfaces is used.

library/network/src/modules/NetworkInterfaces.rb
teclator: So i replaced Reset by main call, and removed Reset method.

library/network/src/modules/NetworkInterfaces.rb
teclator: It is not possible to remove deep_copy, at least it should be added to the first return.

library/network/src/modules/NetworkInterfaces.rb
teclator: The differences are in the return and in the operator, or we could use same operator but reject.. but in that case we need other parameter. Is it worth it?

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -235,39 +236,52 @@ def device_type(dev)
...
+      elsif FileUtils.Exists("#{sys_dir_path}/bridge")
+        "br"
+      elsif FileUtils.Exists("#{sys_dir_path}/bonding")
+        "bond"
+      elsif FileUtils.Exists("#{sys_dir_path}/tun_flags")
+        "tap"
+      elsif FileUtils.Exists("/proc/net/vlan/#{dev}")
+        "vlan"
+      elsif FileUtils.Exists("/sys/devices/virtual/net/#{dev}") &&
+          dev =~ /dummy.*/
```
mchf: I think that ```.*``` is useless in the regexp

mchf: and I'd prefer to have whole condition on one line or use same indentation.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1299,73 +1331,74 @@ def GetDevTypeDescription(type, longdescr)
...
-          Map.Keys(Ops.get(@Devices, type, {})),
-          from: "list",
-          to:   "list <string>"
-        )
-      ) do |dev|
-        dev = device_num(dev) if Builtins.issubstring(dev, type)
-        curdevs = Builtins.add(curdevs, dev)
-      end
+      curdevs = @Devices.fetch(type, {}).keys
+      curdevs.map! { |d| d.include?(type) ? device_num(d) : d }
```
mchf: please drop call to ```device_num``` it is obsolete function and assumes that device names are in form ```<type><num>``` Since we support e.g. ```cool_name```s for ```eth``` device it makes no sense to use such function and we should replace all occurrences by something more reasonable.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1299,73 +1331,74 @@ def GetDevTypeDescription(type, longdescr)
...
       original_devs = Filter(@OriginalDevices, devregex)
-      Builtins.y2debug("OriginalDevs=%1", original_devs)
-      Builtins.y2debug("Devs=%1", devs)
+      log.debug("OriginalDevs=#{original_devs}")
+      log.debug("Devs=#{devs}")
       devs == original_devs
     end
 
+    # It returns an array with the next N devices available
+    # for the given device type.
```
mchf: ehm. No. It returns something like an array of N free indexes. Device name has to be constructed by caller. And it is silently expected that it will be ```<type><i>```

Honestly I dislike this function and it is candidate for removal for me.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1587,50 +1620,41 @@ def GetIP(device)
...
       end
 
-      deep_copy(ret)
+      ret
     end
 
     # Check if any device is using the specified provider
     # @param [String] provider provider identification
     # @return true if there is any
     def LocateProvider(provider)
```
mchf: drop. Support for DSL / Modem / ISDN was removed sometime ago

library/network/src/modules/NetworkInterfaces.rb
teclator: It make sense, and we have a lot of code that relies on olde device names so i will try also to think in this way in the future.

library/network/src/modules/NetworkInterfaces.rb
teclator: I was a litle thick trying to describe it :)

library/network/src/modules/NetworkInterfaces.rb
teclator: The problem is that the method in general does not make sense if we remove the device_num as it is concived to obtain next available device name based on the until now common device name so... in that case we should change also the way we obtain it to propose a device name when the user try to add a new device...

library/network/src/modules/NetworkInterfaces.rb
teclator: I also dislike it but it is still in use

mchf:
LGTM once the branche is rebased / merged / ... and travis is happy.

teclator:
Rebased branch, and Travis is now happy.

imobachgs:
The PR was reviewed and rebased changing only the version number and a few returns ([here](https://github.com/yast/yast-yast2/pull/433/files#diff-36f13a6d2bafc7ad7bcfa6f8a452e81dL244). So LGTM.

PR #434, opened 2016-02-05
AY profile generated at the end of installation contains valid hostname

mchf:
Bnc#960040

### DIFF
--- a/library/types/src/modules/Hostname.rb
+++ b/library/types/src/modules/Hostname.rb
@@ -155,30 +155,20 @@ module Yast
     # (uses hostname --fqdn)
     # @return FQ hostname
     def CurrentFQ
-      fqhostname = ""
+      hostname_data = SCR.Execute(path(".target.bash_output"), "hostname --fqdn")
 
-      hostname_data = Convert.to_map(
-        SCR.Execute(path(".target.bash_output"), "hostname --fqdn")
-      )
       if hostname_data.nil? || hostname_data["exit"] != 0 || invalid_hostname?(hostname_data["stdout"].to_s.strip)
-        fqhostname = if FileUtils.Exists("/etc/hostname")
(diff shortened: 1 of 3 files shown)


library/types/src/modules/Hostname.rb
```diff
@@ -155,30 +155,20 @@ def MergeFQ(hostname, domain)
...
-
-        if fqhostname == "" || fqhostname.nil?
-          # last resort (#429792)
-          fqhostname = "linux.#{@DefaultDomain}"
-        end
-        Builtins.y2warning("Using fallback hostname %1", fqhostname)
+        Builtins.y2warning("Using fallback hostname")
+
+        fqhostname = SCR.Read(path(".target.string"), "/etc/hostname") || ""
+        fqhostname = "linux.#{@DefaultDomain}" if fqhostname.nil? || fqhostname.empty?
```
teclator: NP: fqhostname.nil? is not needed

teclator:
LGTM

mvidner:
If you feel adventurous, you can even replace SCR with 
```rb
require "yast2/execute"
require "yast2/target_file"
Yast::Execute.on_target("hostname", "--fqdn")
Yast::TargetFile.read("/etc/hostname")
```

library/types/src/modules/Hostname.rb
mchf: ah sure, that was reason for the above change and in the end I forgot to delete it;-) Thanks for pointing that.

mchf:
@mvidner: This function is used even in SCR chroot magic parts of code and similar, so I don't feel that brave yet ;-)

mvidner:
LGTM

mchf:
Thanks for reviews.

kobliha:
No testcase?

PR #435, opened 2016-02-05
AY profile generated at the end of installation contains valid hostname

mchf:
Forgot to base previous PR on maintenance branch so this is a cherry pick of https://github.com/yast/yast-yast2/pull/434

### DIFF
--- a/library/types/src/modules/Hostname.rb
+++ b/library/types/src/modules/Hostname.rb
@@ -155,30 +155,20 @@ module Yast
     # (uses hostname --fqdn)
     # @return FQ hostname
     def CurrentFQ
-      fqhostname = ""
+      hostname_data = SCR.Execute(path(".target.bash_output"), "hostname --fqdn")
 
-      hostname_data = Convert.to_map(
-        SCR.Execute(path(".target.bash_output"), "hostname --fqdn")
-      )
       if hostname_data.nil? || hostname_data["exit"] != 0 || invalid_hostname?(hostname_data["stdout"].to_s.strip)
-        fqhostname = if FileUtils.Exists("/etc/hostname")
-                       SCR.Read(path(".target.string"), "/etc/hostname")
-                     else
-                       ""
-                     end
+        Builtins.y2warning("Using fallback hostname")
 
-        if fqhostname == "" || fqhostname.nil?
-          # last resort (#429792)
-          fqhostname = "linux.#{@DefaultDomain}"
-        end
-        Builtins.y2warning("Using fallback hostname %1", fqhostname)
+        fqhostname = SCR.Read(path(".target.string"), "/etc/hostname") || ""
+        fqhostname = "linux.#{@DefaultDomain}" if fqhostname.empty?
       else
-        fqhostname = Ops.get_string(hostname_data, "stdout", "")
+        fqhostname = hostname_data["stdout"] || ""
       end
 
       fqhostname = String.FirstChunk(fqhostname, "\n")
 
-      Builtins.y2debug("Current FQDN: %1", fqhostname)
+      Builtins.y2milestone("Current FQDN: %1", fqhostname)
       fqhostname
     end
 
--- a/library/types/src/modules/Hostname.rb
+++ b/library/types/src/modules/Hostname.rb
@@ -163,7 +163,7 @@ module Yast
         fqhostname = SCR.Read(path(".target.string"), "/etc/hostname") || ""
         fqhostname = "linux.#{@DefaultDomain}" if fqhostname.empty?
       else
-        fqhostname = hostname_data["stdout"] || ""
+        fqhostname = hostname_data["stdout"]
       end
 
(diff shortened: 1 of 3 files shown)


library/types/src/modules/Hostname.rb
```diff
@@ -155,30 +155,20 @@ def MergeFQ(hostname, domain)
...
-                     else
-                       ""
-                     end
-
-        if fqhostname == "" || fqhostname.nil?
-          # last resort (#429792)
-          fqhostname = "linux.#{@DefaultDomain}"
-        end
-        Builtins.y2warning("Using fallback hostname %1", fqhostname)
+        Builtins.y2warning("Using fallback hostname")
```
jreidinger: I think at least you should print what return previous command, so it is logged

library/types/src/modules/Hostname.rb
```diff
@@ -155,30 +155,20 @@ def MergeFQ(hostname, domain)
...
+
+        fqhostname = SCR.Read(path(".target.string"), "/etc/hostname") || ""
+        fqhostname = "linux.#{@DefaultDomain}" if fqhostname.empty?
       else
-        fqhostname = Ops.get_string(hostname_data, "stdout", "")
+        fqhostname = hostname_data["stdout"] || ""
```
jreidinger: why `""` is needed? I think above you checked that it is valid hostname...also yo remove to_s and strip, so above you check if stripped and converted version is correct and here you do nto use it.

library/types/src/modules/Hostname.rb
```diff
@@ -155,30 +155,20 @@ def MergeFQ(hostname, domain)
...
-                     end
-
-        if fqhostname == "" || fqhostname.nil?
-          # last resort (#429792)
-          fqhostname = "linux.#{@DefaultDomain}"
-        end
-        Builtins.y2warning("Using fallback hostname %1", fqhostname)
+        Builtins.y2warning("Using fallback hostname")
+
+        fqhostname = SCR.Read(path(".target.string"), "/etc/hostname") || ""
```
jreidinger: what if file contain some new line? I think at least strip here would be helpful. so maybe something like `SCR.Read(path(".target.string"), "/etc/hostname").to_s.strip`

library/types/src/modules/Hostname.rb
```diff
@@ -155,30 +155,20 @@ def MergeFQ(hostname, domain)
...
+        fqhostname = "linux.#{@DefaultDomain}" if fqhostname.empty?
       else
-        fqhostname = Ops.get_string(hostname_data, "stdout", "")
+        fqhostname = hostname_data["stdout"] || ""
       end
 
       fqhostname = String.FirstChunk(fqhostname, "\n")
```
jreidinger: ah and here it do new line cleaning. Interesting.

library/types/src/modules/Hostname.rb
```diff
@@ -247,7 +237,7 @@ def CurrentDomain
     #
     # Basicaly, linuxrc sometimes resolves IP when querying "hostname --fqdn"
     def invalid_hostname?(hostname)
-      Mode.installation && IP.Check(hostname)
+      Stage.initial && IP.Check(hostname)
```
jreidinger: so only invalid hostname is when it contain IP? Do not look much robust to me.

library/types/src/modules/Hostname.rb
mchf: yes. Some Ycp relics stayed ... I've focused on workflow (simplifying ```if ... if ... else``` constructs

library/types/src/modules/Hostname.rb
mchf: it doesn't hurt me ... so I put it there automatically.

library/types/src/modules/Hostname.rb
mchf: this is tracing note ... domain is printed bellow.

jreidinger: domain but not output of this command which can be useful.

library/types/src/modules/Hostname.rb
jreidinger: well, it confuse reader as it looks like you expected nil there. So use it only when needed :) Do not confuse network reader more then needed.

mchf: Well I think that there are much more things which can confuse reader of network code ;-) But ok I'll drop it.

library/types/src/modules/Hostname.rb
mchf: this method is a workaround / hotfix until fate#900647 is somehow handled.

library/types/src/modules/Hostname.rb
mchf: well I usually agree / do so. But in this case the output can be easily extracted from other output and I don't want to fill the log more than needed.

jreidinger:
LGTM

mchf:
Thanks for review.

PR #436, opened 2016-02-05
SuSEFirewall module reads system configuration during autoinstallation

imobach:
* Changes SuSEFirewall to read system's configuration during installation, instead of using `ReadDefaultConfiguration()`. It's should fix [bsc#963585](https://bugzilla.suse.com/show_bug.cgi?id=963585).
* Add a new method `read_and_import` which read the firewall configuration and merges a configuration hash (usually from an AutoYaST profile). In contrast to the regular `Import`, it performs a merge.


jreidinger:
LGTM

imobach:
Thanks!

PR #441, opened 2016-02-19
Add multi messages dialog

imobach:
As part of fixing [bsc#943805](https://bugzilla.suse.com/show_bug.cgi?id=943805) we need to show a set of messages during packages installation/upgrade. Those messages are quite unusual and have no constraints regarding their length.

So I decided to write a new kind of dialog to be used in this case:

* Inherits from `::UI::Dialog`.
* Available also through Report and Popup modules.
* Supports setting a time-out.

multi-messages-qt-timeout

Next/Back button won't be available when showing only 1 message:

multi-messages-no-timeout

It also looks ok in a ncurses interface:

multi-messages-ncurses

Take into account that the format of the message is set by the message itself. Using a presenter is the best option.


lslezak:
At first I have some comments from the UI/UX POV:

- There are too many frames. The Qt UI is acceptable, but the ncurses dialog has too many borders around, it looks strange. I'd remove the `Frame` widget put a simple `Label` instead.
- I'm not sure about the navigation. Personally I'd prefer a `ComboBox` with package names at the top and displaying the text for the selected package below.
With that you would easily see all affected packages (after clicking the combo box) so you  could directly switch to a package which is important for  you or simply press `Close` if you don't care about any package.

lslezak:
At first I have some comments from the UI/UX POV, it does not look bad, but I think it could be a bit better:

- There are too many frames. The Qt UI is acceptable, but the ncurses dialog has too many borders around, it looks strange. I'd remove the `Frame` widget put a simple `Label` at the top instead.
- I'm not sure about the navigation. Personally I'd prefer a `ComboBox` with package names at the top and displaying the text for the selected package below, like we have e.g. for the [license selection](https://openqa.opensuse.org/tests/122515/modules/welcome/steps/1) at installation.
With that you would easily see all affected packages (after clicking the combo box) so you could directly switch to a package which is important for  you or simply press `Close` if you don't care about any package.
- Too many navigation buttons at the bottom, with a combo box (as mentioned above) you could get rid of the `Next` and `Back` buttons.
- I'd add some description to the timeout label, it looks unclear what the plain number means or what happens. E.g. use something like `Close the dialog after X seconds.`

imobach:
* You're right. I'll follow your suggestions about removing frames (I think it's the first time since I joined that I write UI code). :)
* I like the `ComboBox` because of the fact that you can see all affected packages. They should be only 1 or 2 packages, but after all this is some kind of 'generic' dialog.
* Reducing the number of buttons also sounds good.
* About the timeout label, I didn't use one to be consistent with how other dialogs in the Report module look.

imobach:
@lslezak Thinking it twice about the `ComboBox` idea, from my point of view, it makes the navigation harder (specially when you're in the ncurses interface). So now I'm not so convinced about changing it :)

BTW, I have reduced the amount of frames as you suggested and it now looks way better :) Thanks!

lslezak:
Yes, it looks much better now, thanks! :+1: 

I'm still thinking about the navigation. The `Back` and `Next` buttons looks rather like the main navigation buttons for advancing in the workflow. Use something else? What about moving the buttons next to the label at the top (next to the `X out of Y` text)?

And what about using unicode arrow symbols `←` and `→` as the labels? Would it be better (or worse)?

I think we should involve the designers in this feature, this is quite complicated from the UX POV...

lslezak:
BTW the code coverage increased, nice! :+1:

library/general/src/lib/ui/multi_messages_dialog.rb
```diff
@@ -0,0 +1,247 @@
...
+    end
+
+    # Dialog's content
+    #
+    # Overrides UI::Dialog#dialog_content
+    #
+    # @return [Yast::Term] UI content for dialog
+    #
+    # @see UI::Dialog#dialog_content
+    def dialog_content
```
lslezak: NP: IMHO this and some other methods should be private (or protected). It should be only possible to create and run the dialog. The other methods are just implementation details and should be hidden from outside.

library/general/src/lib/ui/multi_messages_dialog.rb
```diff
@@ -0,0 +1,247 @@
...
+    def total
+      messages.size
+    end
+
+    # UI content describing the UI for the current message's body
+    #
+    # @return [Yast::Term] UI content for dialog
+    def message_body_ui
+      title =
+        if messages.size > 1
```
lslezak: NP: Um, I don't like much `if` condition in assignment, it's quite unusual in the YaST code. I'd rather prefer the ternary operator here.

library/general/src/lib/ui/multi_messages_dialog.rb
```diff
@@ -0,0 +1,247 @@
...
+          HStretch(),
+          PushButton(Id(:back), Opt(:disabled), Yast::Label.BackButton),
+          HStretch(),
+          PushButton(Id(:next), Opt(last_message? ? :disabled : :enabled), Yast::Label.NextButton)
+        ]
+      end
+      if timed?
+        buttons.insert(2, PushButton(Id(:stop), Yast::Label.StopButton), HStretch())
+      end
+      HBox(*buttons)
```
lslezak: Maybe a `ButtonBox` should be used here instead of the `HBox`...

BTW: A nice trick with creating the button list! :+1:

library/general/src/modules/Report.rb
```diff
@@ -39,6 +39,8 @@ module Yast
   # @TODO not all methods respect all environment, feel free to open issue with
   #   method that doesn't respect it.
   class ReportClass < Module
+    include Logger
```
lslezak: NP: I know this is correct, but I always prefer explicit `Yast::Logger` to not confuse it with the `Logger` from the Ruby stdlib.

library/general/src/lib/ui/multi_messages_dialog.rb
```diff
@@ -0,0 +1,247 @@
...
+    DEFAULT_MIN_WIDTH = 60
+    TIMEOUT_STEP = 1000 # miliseconds
+
+    attr_reader :messages, :headline, :min_height, :min_width, :timeout
+
+    # Constructor
+    #
+    # @param headline   [String]  Dialog's headline
+    # @param messages   [Array]   Array of messages. Objects in the array should
+    #                             respond to #title and #body methods.
```
lslezak: NP: Maybe this file should provide a generic message class? I'm not sure if we want to create a separate class at every place where this dialog is used...

imobach:
@lslezak Kenneth suggested showing all the messages in one single RichText control (separating all of them with titles). So it would remove all the navigation hassle. Moreover in this case we don't expect more than one or two messages.

About the timer, if it's used only in AutoYaST, it's ok.

As a side effect, I would get rid of a lot of code and I'd move this class to the yast-packager module.

lslezak:
OK, merging all texts into a `RichText` sounds as a good idea.

imobach:
I'm closing this PBI as it would be implemented in a easier way.

PR #442, opened 2016-02-22
Add commit result reporting

imobach:
This PR is the base one to solve [bsc#943806](https://bugzilla.suse.com/show_bug.cgi?id=943805):

* Add a method `PackagesUI.show_update_messages` to show messages coming from libzypp.
* Fixes the handling of richtext in the Popup module.
* Add tests for Report and Popup modules.

installation-messages-qt

installation-messages-ncurses


library/general/src/modules/Popup.rb
```diff
@@ -1913,6 +1910,20 @@ def ShowFile(headline, filename)
...
+
+    # Adapt a message to be shown as plaintext in a Richtext widget
+    #
+    # * Escape tags
+    # * Replace "\n" by "<br>"
+    #
+    # @param [String] message Message to be converted
+    # @return [String] Converted message
+    def convert_to_plaintext(message)
+      String.EscapeTags(message).split("\n").join("<br>")
```
lslezak: NTH: The `split` + `join` trick is a relic from the YCP times. Use native Ruby [`gsub`](http://ruby-doc.org/core-2.1.4/String.html#method-i-gsub) (or [`tr`](http://ruby-doc.org/core-2.1.4/String.html#method-i-tr) which is better in this case).

library/general/src/modules/Popup.rb
```diff
@@ -74,6 +74,8 @@ def popupLayoutInternalTypeWithLabel(headline, message, button_box, label, richt
       button_box = deep_copy(button_box)
       content = Empty()
 
+      body = richtext ? message : convert_to_plaintext(message)
```
lslezak: This always escapes plain text messages. But maybe some existing calls pass already escaped text, that would cause double escaping.

Could you grep the YaST sources whether there is a such case? It's probably OK to change it, but we should double check first...

imobach: Well, we could be more conservative and don't do any conversion at all. From my POV escaping tags when using richtext was wrong, so we could show the message as is in both cases.

library/general/test/popup_test.rb
```diff
@@ -36,4 +36,263 @@
...
+    it "shows a popup without escaping tags" do
+      expect(subject).to receive(:RichText).with("<h1>Title</h1>")
+      subject.Message("<h1>Title</h1>")
+    end
+  end
+
+  describe ".Warning" do
+    before { allow(ui).to receive(:OpenDialog) }
+
+    it "shows a popup with escaped tags" do
```
lslezak: This should be updated to `without escaping`.

library/general/test/popup_test.rb
```diff
@@ -36,4 +36,263 @@
...
+      expect(subject).to receive(:RichText).with("<h1>Title</h1>")
+      subject.Warning("<h1>Title</h1>")
+    end
+  end
+
+  describe ".Error" do
+    before { allow(ui).to receive(:OpenDialog) }
+
+    it "shows a popup with escaped tags" do
```
lslezak: And here as well.

library/general/test/report_test.rb
```diff
@@ -0,0 +1,195 @@
...
+
+      it "shows a timed popup" do
+        expect(Yast::Popup).to receive(:TimedLongError)
+          .with("Message", 1)
+        subject.LongError("Message")
+      end
+    end
+
+    context "when logging is enabled" do
+      it "logs the message" do
```
lslezak: This test is a pretty much a duplicate of the LongWarning/"when logging is enabled" test above. What about using the [shared examples](http://www.relishapp.com/rspec/rspec-core/v/3-3/docs/example-groups/shared-examples) RSpec feature?. See also this [registration example](https://github.com/yast/yast-registration/blob/e6b838a237d54430213959764356693af983a71c/test/registration_spec.rb#L76-L82).

library/packages/src/lib/packages/update_messages_view.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  private
+
+    # Convert one message to richtext
+    #
+    # @return [String] Message converted to richtext
+    def message_to_richtext(message)
+      location = format(_("This message will be available at %s"), message.installation_path)
+      "<h2>#{message.solvable}</h2><p><em>#{location}</em></p>" \
+        "<br>#{message.text.strip.gsub("\n", "<br>")}"
```
lslezak: NP: To be on the safe side you should also escape the solvable name and the file location. 

Just a note: Alternatively you could use an ERB template file. Then you could simply use the usual `h()` ERB shortcut for escaping, see this [registration example](https://github.com/yast/yast-registration/blob/master/src/data/registration/certificate_summary.erb).

But for such a short message an inline string is still acceptable. Just a hint if you would extend the message or formatting later... :wink:

library/packages/src/modules/PackagesUI.rb
```diff
@@ -774,6 +777,22 @@ def ShowInstallationSummary
...
     end
 
+    # Show messages coming from libzypp about installed packages
+    #
+    # This messages are retrieved from libzypp.
+    #
+    # @param [Array] result Result from package commit (as it comes from PkgCommit)
+    def show_update_messages(result)
+      commit_result = ::Packages::CommitResult.from_result(result)
+      if commit_result.update_messages.empty?
```
lslezak: I'd use a guard clause here (saves indentation):

```ruby
return false if commit_result.update_messages.empty?

view = ...

```

library/packages/test/update_messages_view_test.rb
```diff
@@ -0,0 +1,24 @@
...
+    [
+      Packages::UpdateMessage.new("pkg1", "message 1", "/var/adm/path-1", "/var/adm/path-1"),
+      Packages::UpdateMessage.new("pkg2", "message 2", "/var/adm/path-2", "/var/adm/path-2")
+    ]
+  end
+
+  describe "#richtext" do
+    it "concatenates information of all messages in a richtext string" do
+      expect(view.richtext)
+        .to eq("<h1>Packages messages</h1><h2>pkg1</h2><p><em>This message will be available at /var/adm/path-1</em></p><br>message 1<hr>" \
```
lslezak: NP: I'd be less strict about the message formatting here (to make it shorter and be more flexible if we decide to change the formatting later), we should really focus just on the presence of both messages here. I.e.:

```
.to match(/message 1.*message 2/)
```

library/packages/src/modules/PackagesUI.rb
imobach: I usually prefer to indent as it makes it easier to read (for me), specially when it's not the first line in the method. But I know that in YaST-land :) is quite usual, so I've no problem for changing it.

library/packages/src/lib/packages/update_messages_view.rb
imobach: Yes, I preferred to avoid ERB in this case :)

library/general/test/report_test.rb
imobach: Good point! Well, maybe I went too far with the latest refactor (see below :) ).

lslezak:
LGTM, thanks!

lslezak:
LEBTM (Looks Even Better To Me :wink:)

imobach:
:)

lslezak:
LGTM

imobach:
Thanks!

imobach:
I've improved how the dialog looks like and BTW I've fixed the handling of failures (I detected when pressing the abort button).

lslezak:
Great! With TOC it looks really much better! :+1: 

LGTM

PR #443, opened 2016-02-23
Updated testsuite for Hostname

mchf:


mvidner:
OK, but what is the context? Why do I have to dig around Trello to find the matching PR?

"This is a test for https://github.com/yast/yast-yast2/pull/434"

mchf:
i though that more tests is better and no special reference is needed ;-)

library/types/test/hostname_test.rb
```diff
@@ -10,30 +10,49 @@
 
   let(:etc_hostname) { "etc.hostname.cz" }
   let(:cmd_hostname) { "cmd.hostname.cz" }
+  let(:hostname)     { Yast::Hostname }
```
jreidinger: maybe easier to read is constant assignment on top of file like

```
Hostname = Yast::Hostname
```
this way it is still visible as constant and closer to real usage

library/types/test/hostname_test.rb
```diff
@@ -10,30 +10,49 @@
...
+
+  def allow_execute_hostname(result)
+    ret = { "stdout" => result, "exit" => 0 } if result
```
mvidner: Actually this makes `ret` nil if `result` is nil, but I don't think ".target.bash_output" ever returns nil.
I can get
```rb
{
"exit"=>1, 
"stderr"=>"sh: -c: line 0: syntax error near unexpected token `newline'\n" \
               "sh: -c: line 0: `nonsensecommand('\n", 
"stdout"=>""
}
```

or
```rb
{
"exit"=>127, 
"stderr"=>"sh: nonsensecommand: command not found\n", 
"stdout"=>""
}
```

So I think it is better to specify the entire return value of `.target.bash_output`, not just the stdout part.

library/types/test/hostname_test.rb
```diff
@@ -10,30 +10,49 @@
...
 
-  it "returns output of hostname --fqdn if available" do
     allow(Yast::SCR)
       .to receive(:Execute)
       .with(path(".target.bash_output"), "hostname --fqdn")
-      .and_return("stdout" => cmd_hostname, "exit" => 0)
-
-    expect(Yast::Hostname.CurrentFQ).to eq cmd_hostname
+      .and_return(ret)
```
jreidinger: so it returns nil if result if nil? I think it is not correct behavior of SCR. I think simple adding `result = ""` helps with issue ( and remove such if result ).
better solution for me if you want also test error states is
```
def allow_execute_hostname(stdout: "", exit: 0, stderr: "")
end
```

so you have reasonable defaults and can be in test easily changed.

library/types/test/hostname_test.rb
```diff
@@ -10,30 +10,49 @@
...
+  end
+
+  it "returns output of hostname --fqdn if available" do
+    allow_execute_hostname(cmd_hostname)
+
+    expect(hostname.CurrentFQ).to eq cmd_hostname
+  end
+
+  it "reads /etc/hostname when hostname --fqdn fails" do
+    allow_execute_hostname(nil)
```
jreidinger: and here it shows. I worry it is not common fail of reading, as it should fail with exit code and some stderr output.

library/types/test/hostname_test.rb
mchf: there is ```nil``` bcs when it is not valid value I don't care. But, you're right i should use one of 'valid'  return values for invalid case.

library/types/test/hostname_test.rb
```diff
@@ -10,30 +10,49 @@
 
   let(:etc_hostname) { "etc.hostname.cz" }
   let(:cmd_hostname) { "cmd.hostname.cz" }
+  let(:hostname)     { Yast::Hostname }
+
+  def allow_execute_hostname(stdout, code = 0, stderr = "")
```
jreidinger: you do not like named parameters? :)

library/types/test/hostname_test.rb
```diff
@@ -10,30 +10,49 @@
...
+  end
+
+  it "returns output of hostname --fqdn if available" do
+    allow_execute_hostname(cmd_hostname)
+
+    expect(hostname.CurrentFQ).to eq cmd_hostname
+  end
+
+  it "reads /etc/hostname when hostname --fqdn fails" do
+    allow_execute_hostname("", -1)
```
jreidinger: exit code can be -1? I prefer to use real exit code when it fails so maybe in future you can enhance it to catch expected exit code and act in given way.

library/types/test/hostname_test.rb
```diff
@@ -10,30 +10,49 @@
...
 
-    expect(Yast::Hostname.CurrentFQ).to eq etc_hostname
+    expect(hostname.CurrentFQ).to eq etc_hostname
+  end
+
+  it "returns default hostname proposal when generating AY profile at the end of installation" do
+    Yast.import "Stage"
+
+    allow_execute_hostname("1.1.1.1")                     # linuxrc sometimes provides IP as hostname
+    allow_read_hostname("")                                  # linuxrc do not provide /etc/hostname
```
jreidinger: so it do not fail and just return empty string, right?

jreidinger: ah, it is read, so nevermind.

library/types/test/hostname_test.rb
mchf: huh, there should be 1 ... I've no idea where I found - :-/

jreidinger:
in general just NP, so LGTM

library/types/test/hostname_test.rb
mchf: I've never used that -> study

mchf:
Thanks for reviews

PR #444, opened 2016-02-24
network: testsuite: SuSEFirewall: Fix Import/Export SuSEFirewall2 test

hwoarang:
Commit c6b32ea1656c "Move read_and_import functionality into
SuSEFirewall.Import" changed the behavior of the Import method to
not set the default values for the SF2 sysconfig options.
However, this broke the Import/Export test since the expected output
was changed to expect the default options as well. We fix that by
only checking for the options we explicitely passed to the test.


hwoarang:
I will submit a better fix

PR #446, opened 2016-02-26
System Role (FATE#317481)

mvidner:
https://fate.suse.com/317481 Add 'System Role' capabilities to installer

### DIFF
--- a/library/control/src/modules/ProductFeatures.rb
+++ b/library/control/src/modules/ProductFeatures.rb
@@ -43,6 +43,9 @@ module Yast
       # See defaults map below for sample contents
       @features = nil
 
+      # See SetOverlay and ClearOverlay
+      @backup_features = nil
+
       # Default values for features
       # two-level map, section_name -> [ feature -> value ]
       @defaults = {
@@ -316,6 +319,32 @@ module Yast
       @features = deep_copy(import_settings)
 
       nil
+    end
+
+    # Overlay, or override, some features.
+    # The intended use is to use SetOverlay to apply some features
+    # specified by a system role (FATE#317481).
+    # Clear would be called only when coming Back
+    # to the role selection dialog (and then SetOverlay with a different one)
+    # @param features [Hash{String => Hash{String => Object}}] in the same
+    #   format as `@features` or `@defaults`
+    # @return void
+    def SetOverlay(features)
+      @backup_features = deep_copy(@features)
+
+      features.each do |section_name, section|
+        section.each do |k, v|
+          SetFeature(section_name, k, v)
+        end
+      end
+    end
+
+    # Remove a `@features` overlay; does nothing if SetOverlay was never called.
+    # @see SetOverlay
+    # @return void
+    def ClearOverlay
+      return if @backup_features.nil?
+      @features = deep_copy(@backup_features)
     end
 
     publish function: :GetStringFeature, type: "string (string, string)"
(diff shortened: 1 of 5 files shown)


library/control/src/modules/ProductFeatures.rb
```diff
@@ -318,6 +318,25 @@ def Import(import_settings)
       nil
     end
 
+    # Overlay, or override, some features.
+    # The intended use is to use PushOverlay to apply some features
+    # specified by a system role (FATE#317481).
+    # PopOverlay would be called only when coming Back
```
jreidinger: Sounds a bit like overengineering for me. Why not when going back simply reconstruct original features? Something like InitFeatures(true)

library/control/src/modules/ProductFeatures.rb
```diff
@@ -318,6 +318,25 @@ def Import(import_settings)
...
+    # to the role selection dialog (and then PushOverlay with a different one)
+    # @param features [Hash{String => Hash{String => Object}}] in the same
+    #   format as `@features` or `@defaults`
+    def PushOverlay(features)
```
jreidinger: Why not simple override?

library/control/src/modules/ProductFeatures.rb
```diff
@@ -318,6 +318,25 @@ def Import(import_settings)
...
+      # FIXME: 1) the stack is missing 2) the merging is totally fragile
+      features.each do |k, v|
+        @features[k].merge!(v)
```
jreidinger: why not using something like SetFeature?

library/control/src/modules/ProductFeatures.rb
mvidner: [InitFeatures](https://github.com/yast/yast-yast2/blob/system-role/library/control/src/modules/ProductFeatures.rb#L95) is not the right one because it simply uses the hard-coded defaults.

jreidinger: OK, so let me restate. why we have to have pop/push and not simple reset? which return to product defaults?

library/control/src/modules/ProductFeatures.rb
mvidner: Because of the requirement to be able to go back and select a different role.

jreidinger: as I mention above `reset` and `set_overlay` do same stuff without need to implement any stack

mvidner: Ah, now I see. I think I thought of a stack because it is an easy way to implement it so I thought why not expose it in the API. But if your point is to expose as restricted API as we can afford, to do a Minimal Viable Product, then I can implement `SetOverlay` and `ResetOverlay`, with just a single `@saved_features` hash.

jreidinger: make sense. Only with `ResetOverlay` I think better name is `ClearOverlay`

library/control/src/modules/ProductFeatures.rb
```diff
@@ -43,6 +43,9 @@ def main
       # See defaults map below for sample contents
       @features = nil
 
+      # See SetOverlay and ClearOverlay
+      @backup_features = nil
```
jreidinger: actually not needed, but if you would like to have it here for documentation purpose, it is fine for me.

jreidinger:
now it looks good, so if you finish pending tests, then it is LGTM

mvidner:
Thanks!

PR #449, opened 2016-03-10
Introduce PowerNV platform

k0da:
We treated PowerNV as CHRP platform for a while, for compatibility
reasons.

This introduces board_powernv

### DIFF
--- a/library/general/src/modules/Arch.rb
+++ b/library/general/src/modules/Arch.rb
@@ -200,13 +200,11 @@ module Yast
         end
         Builtins.y2milestone("_board_compatible '%1' \n", @_board_compatible)
         @_board_compatible = "wintel" if i386 || x86_64
-        # hwinfo expects CHRP/PReP/iSeries/MacRISC* in /proc/cpuinfo
+        # hwinfo expects CHRP/PReP/iSeries/MacRISC*/PowerNV in /proc/cpuinfo
         # there is no standard for the board identification
         # Cell and Maple based boards have no CHRP in /proc/cpuinfo
         # Pegasos and Cell do have CHRP in /proc/cpuinfo, but Pegasos2 should no be handled as CHRP
         # Efika is handled like Pegasos for the time being
-        # Treat PowerNV as CHRP. It is harmless for now. Patch for hwinfo is sent but it is better to be safe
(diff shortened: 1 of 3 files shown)


library/general/src/modules/Arch.rb
```diff
@@ -298,6 +293,11 @@ def board_chrp
       ppc && board_compatible == "CHRP"
     end
 
+    # true for all "CHRP" ppc boards
```
jreidinger: so all chrp boards are powerNV? what is different to boarch_chrp? or just copy and paste?
Can you please also add to comment link to some documentation of this kind of ppc board?

package/yast2.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Mar 10 10:15:13 UTC 2016 - dvaleev@suse.com
+
+- Introduce OPAL platform (boo#970582)
```
jreidinger: and here is name opal platform. It is a bit confusing for me.

jreidinger:
pull request is not mergable. For me the most important note is about documentation for this board, so we know where to look if there is some problems with it.

package/yast2.changes
kobliha: +1 PLS be consistent, all the time

kobliha:
The bug in bugzilla says what, but doesn't say why or does not link any source of information. Moreover, FATE should be more appropriate as we need QA testing and documentation here.

library/general/src/modules/Arch.rb
k0da: Just copy paste.. Fixed

package/yast2.changes
k0da: Fixed

k0da:
What kind of information is required? 

We need that platform so we could partition baremetal systems differently. Basically those are kexec systems no stage1 is required

jreidinger:
@k0da I really would like to see in arch file link to some kind of platform/architecture specification by its vendor. So we can check it in future, if there is problem with such platform. Also to check if it is still supported platform by its vendor.

k0da:
updated

jreidinger:
LGTM, so just please merge master into branch so it is mergable and I will merge this one

k0da:
updated

jreidinger:
thanks. merged

PR #450, opened 2016-03-16
Moved cfg_mail.scr from yast2-mail (FATE#319711)

teclator:
- Added cfg_mail.scr due to circular dependencies as showed [here](https://build.opensuse.org/request/show/372909)


jreidinger:
LGTM

PR #451, opened 2016-03-17
Added missing make files to tarball generation. (FATE#319711)

teclator:
- Not added make files so new folders weren't created.


imobach:
LGTM

PR #452, opened 2016-03-18
File conflict callbacks

mvidner:
# Handle File Conflicts in Packages

* [bsc#923590](https://bugzilla.suse.com/show_bug.cgi?id=923590)
* [Trello card](https://trello.com/c/Sty1P76G/100-5-opensuse-923590-p2-yast2-pkg-bindings-yast2-obvious-file-conflicts-not-resolved-missing-file-conflicts-callbacks-in-yast)

A sample repository with testing packages that have file conflicts is [obs://YaST:testing](http://download.opensuse.org/repositories/YaST:/testing/openSUSE_Leap_42.1/).

*Note: The new added file has [100% test code coverage](https://coveralls.io/builds/5581508/source?filename=library%2Fpackages%2Fsrc%2Flib%2Fpackages%2Ffile_conflict_callbacks.rb)* :wink: 

## Background (for the Blog)

Until now the package installation in YaST ignored possible file conflicts in the installed packages. In contrast `zypper` already supports that check for some time.

File conflicts happen when two packages attempt to install files with the same name but different contents. If such conflicting packages are installed the conflicting files will be replaced losing the previous content. The final file content will also depend on the installation order so some issues might look "random". The package which file has been overwritten is actually broken.

YaST now displays a confirmation dialog which asks whether to continue with installation despite the conflicts or abort. Previously YaST silently continued with the package installation which could cause serious troubles later.

File conflicts should normally not happen, at least when you use the original distribution repositories. The OBS build checks for some file conflicts during package build and if there really is a file conflict that it should be marked on the RPM level (so you should not be allowed to select the conflicting packages for installation at first place).

It is up to the user to decide whether it is OK to ignore the conflict or not. If the conflict is for example in a documentation file then ignoring the conflict is usually no problem, but if the conflict is in a binary file or in a system library then it is potentially risky. If you are not sure "Abort" is the safe choice here.

## Screencasts

*(Click an image to display it in full size.)*

### Qt GUI

_In this demo there is a `sleep()` call used in the progress dialog to make the progress visible, in real installation it will be **much** faster. The Packman packages were added just to have more progress steps. The "Actions performed" log is empty as the packages were cached locally from the previous attempts._ :wink: 

file_conflicts

### NCurses UI

*Again, the progress contains `sleep()` to make it visible.*

file_conflicts_ncurses

### Command Line Mode

*The yast2-dhcp-server package has been manually modified to install the conflicting packages beside the usual `dhcp-server` package. In the command line mode the conflicts are ignored and the packages are installed without asking.*

file_conflicts_cmdline

### Installing Extra Packages at Runtime

*Some modules install missing packages when the module is started. In this case I patched the `dhcp-server` module to additionally install the conflicting packages.*

file_conflicts_extra_pkgs2

### DIFF
--- a/library/packages/src/lib/packages/file_conflict_callbacks.rb
+++ b/library/packages/src/lib/packages/file_conflict_callbacks.rb
@@ -1,0 +1,186 @@
+
+# ------------------------------------------------------------------------------
+# Copyright (c) 2016 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+#
+
+require "yast"
+
+module Packages
+  # Default file conflicts callbacks for package bindings. To register the
+  # callbacks in Yast::Pkg just call {Packages::FileConflictCallbacks.register}
+  class FileConflictCallbacks
+    # Widget ID (created by other code)
+    PKG_INSTALL_WIDGET = :progressCurrentPackage
+
+    class << self
+      include Yast::Logger
+      include Yast::I18n
+      include Yast::UIShortcuts
+
+      # register the file conflict callbacks
+      def register
+        Yast.import "Pkg"
+        Yast.import "UI"
+        Yast.import "Progress"
+        Yast.import "Mode"
+        Yast.import "CommandLine"
+        Yast.import "Report"
+        Yast.import "Label"
+        Yast.import "PackageCallbacks"
+
+        textdomain "base"
+
+        register_file_conflict_callbacks
+      end
+
+    private
+
+      # Helper function for creating an YaST function reference
+      def fun_ref(*args)
+        Yast::FunRef.new(*args)
+      end
+
+      # Register the default file conflicts callbacks
+      def register_file_conflict_callbacks
+        log.info "Registering file conflict callbacks"
+
+        Yast::Pkg.CallbackFileConflictStart(fun_ref(method(:start), "void ()"))
+        Yast::Pkg.CallbackFileConflictProgress(fun_ref(method(:progress),
+          "boolean (integer)"))
+        Yast::Pkg.CallbackFileConflictReport(fun_ref(method(:report),
+          "boolean (list<string>, list<string>)"))
+        Yast::Pkg.CallbackFileConflictFinish(fun_ref(method(:finish), "void ()"))
+
+        nil
+      end
+
+      # Is the package installation progress displayed?
+      # @return [Boolean] true if package installation progress is displayed
+      def pkg_installation?
+        Yast::UI.WidgetExists(PKG_INSTALL_WIDGET)
+      end
+
+      # Handle the file conflict detection start callback.
+      def start
+        log.info "Starting the file conflict check..."
+        # TRANSLATORS: progress bar label
+        label = _("Checking file conflicts...")
+
+        if Yast::Mode.commandline
+          Yast::CommandLine.PrintVerbose(label)
+        elsif pkg_installation?
+          # package slideshow with progress already present
+          Yast::UI.ChangeWidget(Id(PKG_INSTALL_WIDGET), :Value, 0)
+          Yast::UI.ChangeWidget(Id(PKG_INSTALL_WIDGET), :Label, label)
+        else
+          # TRANSLATORS: help text for the file conflict detection progress
+          help = _("<p>Detecting the file conflicts is in progress.</p>")
+          # Use the same label for the window title and the progressbar label
+          Yast::Progress.Simple(label, label, 100, help)
+        end
+      end
+
+      # Handle the file conflict detection progress callback.
+      # @param [Fixnum] progress progress in percents
+      # @return [Boolean] true = continue, false = abort
+      def progress(progress)
+        log.debug "File conflict progress: #{progress}%"
+
+        if Yast::Mode.commandline
+          Yast::CommandLine.PrintVerboseNoCR("#{Yast::PackageCallbacksClass::CLEAR_PROGRESS_TEXT}#{progress}%")
+        elsif pkg_installation?
+          Yast::UI.ChangeWidget(Id(PKG_INSTALL_WIDGET), :Value, progress)
+        else
+          Yast::Progress.Step(progress)
+        end
+
+        ui = Yast::UI.PollInput unless Yast::Mode.commandline
+        log.info "User input in file conflict progress (#{progress}%): #{ui}" if ui
+
+        ui != :abort && ui != :cancel
+      end
+
+      # Handle the file conflict detection result callback.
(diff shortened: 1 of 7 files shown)


library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -2759,6 +2854,13 @@ def SetProgressReportCallbacks
       nil
     end
 
+    def SetFileConflictsCallbacks
+      Pkg.CallbackFileConflictStart(fun_ref(method(:FileConflictStart), "void ()"))
+      Pkg.CallbackFileConflictProgress(fun_ref(method(:FileConflictProgress), "boolean (intreger)"))
```
mvidner: intreger -> integer

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -2487,6 +2487,110 @@ def ProcessDone
...
+      end
+    end
+
+    # Handle the file conflict detection progress callback.
+    # @param progress [Fixnum] progress in percents
+    # @return [Boolean] true = continue, false = abort
+    def FileConflictProgress(progress)
+      log.debug "File conflict progress: #{progress}%"
+
+      sleep(1)
```
mvidner: FIXME remove before merging

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -2487,6 +2487,110 @@ def ProcessDone
...
+      end
+    end
+
+    # Handle the file conflict detection finish callback.
+    def FileConflictFinish
+      log.info "File conflict check finished"
+      return if Mode.commandline
+
+      Progress.Finish unless UI.WidgetExists(:progressCurrentPackage)
+    end
```
jreidinger: Btw I think this class is already too big. So at least for this case I suggest to use its namespace like Yast::FileCallbacks and inside it define methods. Ideally in own file. And just use require and Set method here.

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -2487,6 +2487,110 @@ def ProcessDone
...
+    def FileConflictReport(excluded_packages, conflicts)
+      log.info "Excluded #{excluded_packages.size} packages in file conflict check"
+      log.debug "Excluded packages: #{excluded_packages.inspect}"
+      log.info "Found #{conflicts.size} conflicts: #{conflicts.join("\n\n")}"
+
+      # just continue installing packages if there is no conflict
+      return true if conflicts.empty?
+
+      # don't ask in autoyast or command line mode, just report/log the issues and continue
+      if Mode.autoinst || Mode.autoupgrade || Mode.commandline
```
jreidinger: you can and should use `Mode.auto`  https://github.com/yast/yast-yast2/blob/630cf98148facf439a4439c189d610ba031d1537/library/general/src/modules/Mode.rb#L288

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -2487,6 +2487,110 @@ def ProcessDone
...
+      UI.OpenDialog(dialog)
+
+      begin
+        UI.SetFocus(Id(:continue))
+        ret = UI.UserInput
+        log.info "User Input: #{ret}"
+        ret == :continue
+      ensure
+        UI.CloseDialog
+      end
```
jreidinger: method too long, what about separated method just for the dialog?

library/packages/src/lib/packages/file_conflict_callbacks.rb
```diff
@@ -0,0 +1,185 @@
...
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+#
+
+require "yast"
+
+module Packages
+  # Default file conflicts callbacks for package bindings. To register the
+  # callbacks in Yast::Pkg just call {Package::FileConflictCallbacks.register}
```
mvidner: Package**s**::

library/packages/src/lib/packages/file_conflict_callbacks.rb
```diff
@@ -0,0 +1,185 @@
...
+  class FileConflictCallbacks
+    PKG_INSTALL_WIDGET = :progressCurrentPackage
```
mvidner: Maybe a comment, "a widget created by other code"

library/packages/src/lib/packages/file_conflict_callbacks.rb
```diff
@@ -0,0 +1,185 @@
...
+      def dialog(conflicts)
+        button_box = ButtonBox(
+          PushButton(Id(:continue), Opt(:default, :okButton), Yast::Label.ContinueButton),
+          PushButton(Id(:abort), Opt(:cancelButton), Yast::Label.AbortButton)
+        )
+
+        # TRANSLATORS: A popup label, use max. 70 chars per line, use more lines if needed
+        label = _("File conflicts happen when two packages attempt to install\n" \
+          "files with the same name but different contents. If you continue,\n" \
+          "the conflicting files will be replaced losing the previous content.")
```
mvidner: "If you continue (no comma) the conflicting files will be replaced, (comma added) losing..."

library/packages/src/lib/packages/file_conflict_callbacks.rb
```diff
@@ -0,0 +1,185 @@
...
+        log.info "Found #{conflicts.size} conflicts: #{conflicts.join("\n\n")}"
+
+        # just continue installing packages if there is no conflict
+        return true if conflicts.empty?
+
+        # don't ask in autoyast or command line mode, just report/log the issues and continue
+        if Yast::Mode.auto || Yast::Mode.commandline
+          # TRANSLATORS: An error message, %s is the actual list of detected conflicts
+          Yast::Report.Error(_("File conflicts detected, these conflicting files will " \
+            "be overwritten:\n\n%s") % conflicts.join("\n"))
```
mvidner: `join "\n\n"` also here? IIUC a single conflict is a multiline description, like 
```
File /usr/share/conflict
  from install of
     pkgconflict1-0.0.1-1.1.noarch (testing-conflict)
  conflicts with file from install of
     pkgconflict2-0.0.1-2.1.noarch (testing-conflict)
```

library/packages/src/lib/packages/file_conflict_callbacks.rb
lslezak: I have shamelessly taken the [text from zypper...](https://github.com/openSUSE/zypper/blob/27980f701378abcb470bbf5f80abec404e9a6c4a/src/callbacks/rpm.h#L416)

library/packages/src/lib/packages/file_conflict_callbacks.rb
lslezak: Yes, I overlooked this...

mvidner:
Looks good, only version + changelog missing.

package/yast2.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Mar 29 13:37:40 UTC 2016 - lslezak@suse.cz
+
+- Display a confirmation dialog when a file conflict is detected
+  during package installation (bsc#923590)
+- 3.1.180
```
mvidner: Conflict in a conflict PR!!1!

mvidner:
(I have removed an extra blank line in .changes)

LGTM, thank you!

lslezak:
JFYI: I have squashed the commits to have a clean history.

PR #453, opened 2016-03-23
Add proc_modules.scr from yast-installation (bsc#972310)

cwh42:
This is to avoid yast2-sound to depend on yast2-installation.


shundhammer:
LGTM

PR #456, opened 2016-04-04
Spec file cleanup

pjanouch:
Used [spec-cleaner](https://github.com/openSUSE/spec-cleaner), no further changes seemed to be necessary.

First commit from me, more packages to be expected.

### DIFF
--- a/package/yast2.spec
+++ b/package/yast2.spec
@@ -1,7 +1,7 @@
 #
 # spec file for package yast2
 #
-# Copyright (c) 2015 SUSE LINUX GmbH, Nuernberg, Germany.
+# Copyright (c) 2016 SUSE LINUX GmbH, Nuernberg, Germany.
 #
 # All modifications and additions to the file contributed by third parties
 # remain the property of their copyright owners, unless otherwise agreed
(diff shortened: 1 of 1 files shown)


package/yast2.spec
```diff
@@ -19,114 +19,92 @@
...
-
+# moved cfg_security.scr
+Conflicts:      yast2-security <= 2.13.2
+Conflicts:      yast2-sshd < 2.16.1
+Conflicts:      yast2-storage < 2.16.4
+Conflicts:      yast2-tune < 2.15.6
+Conflicts:      yast2-update < 2.16.1
+Obsoletes:      yast2-mail-aliases <= 2.14.0
 Provides:       yast2-lib-sequencer
 Obsoletes:      yast2-lib-sequencer
```
jreidinger: those provides and obsoletes can be also removed. It is ancient noone should use it. So whole Provides and Obsoletes part can be completely removed.

jreidinger:
in general looks OK, just please do more aggressive Obsoletes Conflicts cleaning, as it is very ancient and for SLE12 nothing is used.

kobliha:
Josef is right, except for `Conflicts:      yast2-dns-server < 3.1.17` -- that one is for newer type of conflict and maybe should stay. Anything with version < 3.0.0 is most probably 99.99% obsolete.

package/yast2.spec
```diff
@@ -253,7 +231,7 @@ mkdir -p "$RPM_BUILD_ROOT"/etc/YaST2
 # documentation (not included in devel subpackage)
 %doc %dir %{yast_docdir}
 %doc %{yast_docdir}/COPYING
-%doc %{_mandir}/*/*
+%{_mandir}/*/*
```
kobliha: man pages are not a documentation or is %{_mandir} "always documentation"?

package/yast2.spec
```diff
@@ -139,44 +117,44 @@ Provides:       y2t_menu
...
-
-Requires:       yast2-ruby-bindings >= 3.1.33
-
-Summary:        YaST2 - Main Package
-License:        GPL-2.0
-Group:          System/YaST
+Provides:       yast2-network:%{_datadir}/YaST2/modules/Internet.ycp
+Provides:       yast2-packager:%{_prefix}/lib/YaST2/servers_non_y2/ag_anyxml
+BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+# for Punycode.rb (bnc#651893) - the idnconv tool is located in
```
kobliha: I'm afraid that this is expected to break in the future Maybe we should just require `/usr/bin/idnconv`?

package/yast2.spec
```diff
@@ -1,7 +1,7 @@
 #
 # spec file for package yast2
 #
-# Copyright (c) 2015 SUSE LINUX GmbH, Nuernberg, Germany.
+# Copyright (c) 2016 SUSE LINUX GmbH, Nuernberg, Germany.
```
kobliha: We usually say `2015 - 2016` instead of replacing `2015` with `2016`.

pjanouch: This is what spec-cleaner does automatically. Maybe @scarabeusiv can say something about this.

package/yast2.spec
lslezak: That would not work without touching the respective packages:

```
# cnf idnconv
                       
Program 'idnconv' is present in package 'bind-utils',
which is installed on your system.

Absolute path to 'idnconv' is '/usr/bin/idnconv'.
Please check your $PATH variable to see whether
it contains the mentioned path.

# rpm -q --provides bind-utils
bind-utils = 9.9.6P1-33.1
bind-utils(x86-64) = 9.9.6P1-33.1
bind9-utils
bindutil
config(bind-utils) = 9.9.6P1-33.1
dns_utils
```

Unfortunately package `bind-utils` does not explicitly provide `/usr/bin/idnconv`.

package/yast2.spec
pjanouch: Seems to be so. In spec-cleaner/spec_cleaner/rpmfiles.py it is documented as:
```
Remove all %doc %_mandir to -> %_mandir as it is pointless to do twice
```

package/yast2.spec
lslezak: Um, maybe I'm wrong:

```
# rpm -q --whatprovides /usr/bin/idnconv
bind-utils-9.9.6P1-33.1.x86_64
# zypper se --provides /usr/bin/idnconv
Loading repository data...
Reading installed packages...

S | Name       | Summary                         | Type   
--+------------+---------------------------------+--------
i | bind-utils | Utilities to query and test DNS | package

```

So maybe it would work...

package/yast2.spec
scarabeusiv: This is opensuse guideline, latest year spec was altered is to be put there, no ranges.

package/yast2.spec
scarabeusiv: It is not good idea to expect those path provides to work, always better to work fine.
Good solution would be for idnkit and it friends to provide bind-utils symbols, or bind-utils on the old distro to provide the idnkit symbol.

package/yast2.spec
scarabeusiv: Exactly if you have something in %mandir it is already %doc-ed

mvidner:
LGTM.
I think it is better to just apply `spec-cleaner` in this PR and not mix in other changes.

pjanouch:
I have managed to automate version clean ups against SLE11 (to support migrations) and updated the PR.

package/yast2.spec
```diff
@@ -186,26 +123,26 @@ installation with YaST2.
...
+mkdir -p %{buildroot}%{yast_scrconfdir}
+mkdir -p %{buildroot}%{yast_ybindir}
+mkdir -p %{buildroot}%{yast_ydatadir}
+mkdir -p %{buildroot}%{yast_yncludedir}
+mkdir -p %{buildroot}%{yast_libdir}
+mkdir -p %{buildroot}%{yast_vardir}
+mkdir -p %{buildroot}%{yast_vardir}/hooks
+mkdir -p %{buildroot}%{yast_schemadir}/control/rnc
+mkdir -p %{buildroot}%{yast_schemadir}/autoyast/rnc
+mkdir -p %{buildroot}%{_sysconfdir}/YaST2
```
scarabeusiv: I wonder if this is needed and why it is not done by the yast_install call itself.

lslezak: I think the directories are created automatically by `make install` (which is invoked from `yast_install`). I guess the reason it to have always created these directories even in case there is no file installed into that directory.

scarabeusiv: Well if there are empty folders they are ignored by rpm unless specified as %dir in filelist.

jreidinger:
in general it is improvement, so lets merge it

PR #459, opened 2016-04-12
Link to docs for Cheetah.run

mvidner:
Use a full URL so that YARD can make a link in the rendered docs.
It is especially important because Cheetah.run takes a great variety of
arguments.


jreidinger:
LGTM

imobach:
LGTM

jreidinger:
Just idea maybe we can add there examples how to convert old SCR.Execute and WFM.Execute to cheetah there?

PR #457, opened 2016-04-06
Add FirewallD support to SuSEFirewall2 backend

hwoarang:
Hi,

This patchset extends the SuSEFirewall2 backend with FirewallD ([website](http://www.firewalld.org/), [OBS package](https://build.opensuse.org/package/show/openSUSE:Factory/firewalld)) support. The goal is for yast2 modules to be able to use either SuSEfirewall2 or FirewallD to perform their regular firewall configurations depending on which one is running on the system. Not all methods have been added to FirewallD since some of them do not appear to be used anymore. If something is missing I will fix it in a subsequent commit as bugs or limitations are being identified by the broader adoption and use of the extended API.

The only known limitation so far is in the testsuite where firewalld tests can't run in parallel with SuSEfirewall2. The reason for that is that they both manipulate the Yast::SuSEFirewall and Yast::SuSEFirewallServices objects so if these tests run in parallel strange things can happen ;) The solution so far is to disable either SuSEfirewall2 or FirewallD tests based on which backend is running as I couldn't find a way to disable parallelism on library/network/test/* files. I believe that using "doubles" or using rspec mock to fake methods is not going to actually tests the backend properly so the testsuite will end up being mostly useless. If you happen to know of the proper way to disable parallelism in a specific subdir please let me know. In the meantime I didn't want to delay the PR because of that as I want to get some broader testing of the extended code.

Thank you

PR #462, opened 2016-04-22
Reviewed and fixed docs for SuSEFirewallServices.

mvidner:
Note that in most cases a *service* argument needs a "service:" prefix.
Replacing tabs by spaces fixes `@example` rendering.

This was done to understand the module better as it gets bigger changes in #457 (firewalld).

### DIFF
--- a/library/network/src/modules/SuSEFirewallServices.rb
+++ b/library/network/src/modules/SuSEFirewallServices.rb
@@ -272,8 +272,8 @@ module Yast
     # @return	[Boolean] whether service is defined by package
     #
     # @example
-    #	ServiceDefinedByPackage ("http-server") -> false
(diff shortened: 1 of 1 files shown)


library/network/src/modules/SuSEFirewallServices.rb
```diff
@@ -337,12 +338,13 @@ def GetMetadataAgent(filefullpath)
 
     # Returns service definition.
     # See @services for the format.
-    # If `silent` is not defined or set to `true`, function throws an exception
-    # SuSEFirewalServiceNotFound if service is not found on disk.
+    # If *silent* is `false` (the default), the method throws an exception
```
mvidner: `true` -> `false` is a factual correction. The method is only used in this file so it was easy to check.

imobach:
LGTM

PR #465, opened 2016-05-06
Add Repository and Product classes

imobach:
This PR add a `Packages::Repository` class which offers an API to search, query and enable/disabled repositories. It's quite incomplete at this time but we can extend it in the future.


library/packages/src/lib/packages/repository.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+    # Return repository scheme
+    #
+    # The scheme is determined using the URL
+    #
+    # @raise UnknownScheme
+    def scheme
+      return @scheme unless @scheme.nil?
+      data_from_url = Yast::URL.Parse(url)
```
lslezak: Use the native Ruby `URI` class if you do not need any `Yast::URL` specific feature (which is not the case here).

library/packages/src/lib/packages/repository.rb
```diff
@@ -0,0 +1,118 @@
...
+      return @products unless @products.nil?
+
+      # Filter products from this repository
+      candidates = Yast::Pkg.ResolvableProperties("", :product, "").select do |pro|
+        pro["source"] == repo_id
+      end
+      
+      # Build an array of Packages::Product objects
+      @products = candidates.map do |data|
+        Packages::RepositoryProduct.new(name: data["name"], version: data["version"],
```
lslezak: Just a note: Some time ago I have added also the parts of the `version` attribute into the response, see https://github.com/yast/yast-pkg-bindings/pull/42.

But I think we can keep it simple for now, the version parts can be added later when needed.

imobach: Yes, the idea is to extend it as long as we need new features.

library/packages/src/lib/packages/repository_product.rb
```diff
@@ -0,0 +1,17 @@
...
+    def initialize(name:, version:, arch:, status:, category:)
+      @arch = arch.to_sym
+      @name = name
+      @status = status.to_sym
+      @category = category.to_sym
+      @version = version
+    end
+
+    def ==(other)
+      arch == other.arch && name == other.name && version == other.version
```
imobach: I'm not sure at all about this part.

lslezak: Maybe we should also store the `vendor` attribute and compare it as well.

It's very unlikely to have the very same product (in the same version) from also from a different vendor, but...

library/packages/src/lib/packages/repository_product.rb
```diff
@@ -0,0 +1,17 @@
+module Packages
+  class RepositoryProduct
```
lslezak: I think we should define a base class for sharing the common resolvable data, see https://github.com/yast/yast-pkg-bindings/blob/d91b7dd06db0b7a316ef0ea2506abdba4664cdfb/src/Resolvable_Properties.cc#L187. At the beginning it adds the generic data defined for all resolvables, then the object class specific data.

library/packages/test/repository_product_test.rb
```diff
@@ -0,0 +1,51 @@
...
+      category: "addon", status: :available)
+  end
+
+  describe "==" do
+    let(:other_data) do
+      { name: "openSUSE", version: "20160405", arch: "x86_64",
+        category: "addon", status: :installed }
+    end
+
+    context "when name and arch" do
```
lslezak: "when name and arch **match**"

library/packages/src/lib/packages/repository_product.rb
```diff
@@ -0,0 +1,48 @@
+module Packages
+  # Represent a product which is present in a repository. At this
+  # time this class is responsible for finding out whether two
+  # products instances are the same (for example, coming from different
+  # repositories).
+  class RepositoryProduct
```
lslezak: NP: Maybe simple `Product` could be enough as it already is in the `Packages` name space, there should not be any conflict. IMHO `Packages::Product` looks better.

library/packages/src/lib/packages/repository.rb
```diff
@@ -0,0 +1,167 @@
...
+    #
+    # @return [Array<Packages::RepositoryProduct>] Addons in the repository
+    #
+    # @see #products
+    def addons
+      products.select { |p| p.category == :addon }
+    end
+
+    # Enable the repository
+    #
```
lslezak: It changes the repository status only in memory, to make the change persistent you need to call `Pkg::SourceSaveAll`. I think this should be mentioned in the documentation comment to make it clear.

The same applies to the `disable!` method.

lslezak:
LGTM

imobach:
Thanks!

PR #466, opened 2016-05-09
Fix "when" key in FinishClient#info

imobach:
Fixes "when" key in FinishClient#info.

I'm not releasing a new version (I'll do that as part of the #465 PR).


ancorgs:
LGTM

imobach:
Thanks!

PR #468, opened 2016-05-17
Filter out INTERFACETYPE from ifcfg files

mchf:
See [bsc#972575](https://bugzilla.suse.com/show_bug.cgi?id=972575)


mvidner:
LGTM

mchf:
Thanks for review

PR #469, opened 2016-05-18
Ported fix from Code-11-SP4

mchf:
Bnc972575
https://github.com/yast/yast-yast2/pull/468


library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -627,6 +627,23 @@ def CanonicalizeIP(ifcfg)
...
+    # by yast in the past. According wicked team it makes sense to use it only
+    # in two cases 1) lo device (when it's name is changed - very strongly discouraged)
+    # 2) dummy device
+    #
+    # This function silently modifies user's config files. However, it should make sense
+    # because:
+    # - INTERFACETYPE is usually not needed
+    # - other functions in this module modifies the config as well (see Canonicalize* functions)
+    # - using INTERFACETYPE is reported as a warning by wicked (it asks for reporting a bug)
+    # - it is often ignored by wicked
```
mvidner: \# @return [Hash] filtered devmap
`delete_if` returns the new hash and also modifies the argument. To reduce surprises, please start the body with
```rb
devmap = deep_copy(devmap)
```

mchf: I've just came to the same conclusion and have been working on that. It seems I know you ;-) 

Updated.

mvidner:
Thanks! LGTM.

mchf:
Thanks for review

PR #470, opened 2016-05-18
Merged SLE-12-SP1 branch

mchf:
contains patch for bnc#972575


mvidner:
(For reference, this merges #469 which is a port of #468)

LGTM.

mchf:
Thanks for review

PR #471, opened 2016-05-20
Split firewall classes

hwoarang:
Hi,

This patchset  addresses a comment during the PR#457 review which requires to have one class per file and that file should be named based on the class it contains. There should be no functional changes to the modules themselves or for their reverse dependencies.


jreidinger:
please do not use require_relative in anything under src. YaST allow to overwrite load paths, e.g. with y2update and with require_relative it won't work. So please use common require. Also I think it make sense to have require top level near other requires ( cannot comment to diff as it is too big ).

library/network/src/modules/SuSEFirewall2.rb
```diff
@@ -0,0 +1,2820 @@
...
+# Package: SuSEFirewall configuration
+# Summary: Interface manipulation of /etc/sysconfig/SuSEFirewall
+# Authors: Lukas Ocilka <locilka@suse.cz
+#
+# $Id$
+#
+# Module for handling SuSEfirewall2
+# DO NOT use this class directly. Use SuSEFirewallClass in your modules
+require "yast"
+require_relative "SuSEFirewall"
```
jreidinger: same here, please use common require

library/network/src/modules/SuSEFirewall2Services.rb
```diff
@@ -0,0 +1,608 @@
...
+        "UDP"       => "udp_ports",
+        "RPC"       => "rpc_ports",
+        "IP"        => "ip_protocols",
+        "BROADCAST" => "broadcast_ports"
+      }
+
+      @known_metadata = { "Name" => "name", "Description" => "description" }
+
+      # Services definitions for conversion to the new ones.
+      @OLD_SERVICES = {
```
jreidinger: this also looks like constant and maybe it make sense to store it somewhere as data in yaml e.g.?

library/network/src/modules/SuSEFirewall2Services.rb
```diff
@@ -0,0 +1,608 @@
...
+      if !FileUtils.Exists(SERVICES_DIR) ||
+          !FileUtils.IsDirectory(SERVICES_DIR)
+        log.error "Cannot read #{SERVICES_DIR}"
+        return false
+      end
+
+      all_definitions = SCR.Read(path(".target.dir"), SERVICES_DIR)
+      all_definitions.reject! do |service|
+        IGNORED_SERVICES.include?(service)
+      end
```
jreidinger: in ruby you can use operator `-` see http://ruby-doc.org/core-2.2.0/Array.html#method-i-2D
so it then look like `all_definitions -= IGNORED_SERVICES`

library/network/src/modules/SuSEFirewall2Services.rb
hwoarang: This is part of the old SF2 code. However, with all these splits I would prefer it for code fixes to be made in a separate commit because it is too painful to rebase, do code fixes and then get conflicts in the split files.

library/network/src/modules/SuSEFirewall2Services.rb
hwoarang: Same as above. Overall, no new code has been introduced. The one from SuSEFirewall.rb has been split to SuSEFirewall2.rb and SuSEFirewalld.rb based on the class that it contains.

library/network/src/modules/SuSEFirewalld.rb
```diff
@@ -0,0 +1,1301 @@
...
+    attr_reader :special_all_interface_zone
+
+    # Valid attributes for firewalld zones
+    # :interfaces = [Array<String>]
+    # :masquerade = Boolean
+    # :modified   = [Set<Symbols>]
+    # :ports      = [Array<String>]
+    # :protocols  = [Array<String>]
+    # :services   = [Array<String>]
+    @@zone_attributes = [:interfaces, :masquerade, :modified, :ports, :protocols, :services]
```
jreidinger: why it is class attribute when it is only read? Why not have it constant?

library/network/src/modules/SuSEFirewalld.rb
```diff
@@ -0,0 +1,1301 @@
...
+    # {enable,start}_firewall are "inherited" from SF2 so we can't use symbols
+    # there without having to change all the SF2 callers.
+    @@key_settings = ["enable_firewall", "logging", "routing", "start_firewall"]
```
jreidinger: same here, it can be constant

library/network/src/modules/SuSEFirewalld.rb
```diff
@@ -0,0 +1,1301 @@
...
+      interfaces: [],
+      masquerade: false,
+      modified:   Set.new,
+      ports:      [],
+      protocols:  [],
+      services:   []
+    }
+
+    # We need that for the tests. Nothing else should access the API
+    # directly
```
jreidinger: so here it make sense to use yardoc @api call. Btw if you need to access it from tests only you can use http://apidock.com/ruby/Object/instance_variable_get which do not need public access ( but use it only in tests, not in production code as it break encapsulation )

hwoarang:
@jreidinger I can't move the 'require' on top because then you have SuSEFirewall.rb require SuSEFirewall2 and SuSEFirewall but SuSEFirewall2 and SuSEFirewalld also require SuSEFirewall for inheritance and ruby does not seem able to handle that. Or perhaps I misunderstand the error. Could you suggest how we could avoid that?

hwoarang:
@jreidinger oh I believe that's because 'require' will look to load stuff from /usr/share/Yast2/lib but the modules are actually located in /usr/share/Yast2/modules. Any idea how to fix that? Or is it require_relative our only option here?

jreidinger:
@hwoarang you have two options:

1. place it to lib
2. use `Yast.import` which loading from modules

package/yast2.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri May 20 08:09:04 UTC 2016 - mchandras@suse.de
+
+- Split SuSEFirewall and SuSEFirewallServices classes to separate files
+  as recommended by the Yast coding style.
```
jreidinger: well, if you do not add there FATE number it won't be accepted by IBS as it require for each change have BNC or FATE number.

jreidinger:
ok, few comments. I will test it on tuesday as we need to be sure that nothing is broken

package/yast2.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri May 20 08:09:04 UTC 2016 - mchandras@suse.de
+
+- Split SuSEFirewall and SuSEFirewallServices classes to separate files
+  as recommended by the Yast coding style (fate#318356, gh#yast/yast-yast2#471)
```
kobliha: Nitpicking: Please try to keep `max 80 characters on oneline` rule in changes file

hwoarang: Yes but this is 79 characters long (at least that's what vim says:))

jreidinger: in fact suse changelog guide mention 67 see https://github.com/yast/yast-storage/blob/9af1ee62c9e5a23e24d60f25fa97e754da4e7274/src/modules/Storage.rb#L5558 basic hint is to not exceed dashes line above

jreidinger:
manual testing looks fine. It is a lot of code, but at least it is not now in one file. So I see some improvements. So lets merge it.

hwoarang:
@jreidinger thanks! yes there are improvements to be made of course but it should be easier now with all the classes split into different files.

PR #472, opened 2016-05-31
[Review] Request from 'schubi2' @ 'yast/yast-yast2/review_160531_just_a_test'

schubi2:
  * 71ec7cd syntax
  * 0d31e60 set forceboot after adding an additional product


schubi2:
not needed

PR #474, opened 2016-06-08
Fixed displaying the file conflicts callbacks (bsc#983464)

lslezak:
...when the Progress dialog is not displayed.

Open a new Wizard window and close it at the end, this ensures the original dialog content is restored after finishing the callbacks.

- 3.1.193


imobach:
LGTM

PR #478, opened 2016-06-17
Rubocop update

lslezak:
This is a preview for Rubocop update, let's see and discuss the changes.

Note: The commits `Manually fixed ...` were fixed manually (and thus *deserve extra careful review*), the rest was fixed automatically Rubocop.

See also https://github.com/yast/yast-devtools/pull/107

### DIFF
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1155,7 +1155,7 @@ module Yast
           if !Builtins.haskey(
             Ops.get_map(cmdlineinfo, "actions", {}),
             mapaction
-            )
+          )
             Builtins.y2error(
               "Command line specification maps undefined action '%1'",
               mapaction
@@ -1171,7 +1171,7 @@ module Yast
             if !Builtins.haskey(
               Ops.get_map(cmdlineinfo, "options", {}),
               Convert.to_string(mapopt)
-              )
+            )
               Builtins.y2error(
                 "Command line specification maps undefined option '%1' for action '%2'",
                 mapopt,
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1281,11 +1281,11 @@ module Yast
       res = nil
 
       res = if type == :nohistory
-        Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
+              Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
             elsif type == :noecho
-        Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
+              Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
             else
-        Convert.to_string(SCR.Read(path(".dev.tty")))
+              Convert.to_string(SCR.Read(path(".dev.tty")))
       end
 
       # set the default prompt
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1271,7 +1271,7 @@ module Yast
     def Scan
       res = Convert.to_string(SCR.Read(path(".dev.tty")))
       return nil if res.nil?
-      String.ParseOptions(res,  "separator" => " ")
+      String.ParseOptions(res, "separator" => " ")
     end
 
     # Set prompt and read input from command line
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1278,8 +1278,6 @@ module Yast
       # set the required prompt
       SCR.Write(path(".dev.tty.prompt"), prompt)
 
-      res = nil
-
       res = if type == :nohistory
               Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
             elsif type == :noecho
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -167,12 +167,10 @@ module Yast
         else
           SCR.Write(path(".dev.tty.nocr"), s)
         end
+      elsif newline
+        SCR.Write(path(".dev.tty.stderr"), s)
       else
-        if newline
-          SCR.Write(path(".dev.tty.stderr"), s)
-        else
-          SCR.Write(path(".dev.tty.stderr_nocr"), s)
-        end
+        SCR.Write(path(".dev.tty.stderr_nocr"), s)
       end
 
       nil
@@ -505,24 +503,22 @@ module Yast
                 @aborted = true if !@interactive
               end
             end
-          else
-            # type is missing
-            if v != ""
-              Builtins.y2error(
-                "Type specification for option '%1' is missing, cannot assign a value to the option",
-                o
+          # type is missing
+          elsif v != ""
+            Builtins.y2error(
+              "Type specification for option '%1' is missing, cannot assign a value to the option",
+              o
+            )
+            # translators: error message if option has a value, but cannot have one
+            Print(
+              Builtins.sformat(
+                _("Option '%1' cannot have a value. Given value: %2"),
+                o,
+                v
               )
-              # translators: error message if option has a value, but cannot have one
-              Print(
-                Builtins.sformat(
-                  _("Option '%1' cannot have a value. Given value: %2"),
-                  o,
-                  v
-                )
-              )
-              @aborted = true if !@interactive
-              ret = false
-            end
+            )
+            @aborted = true if !@interactive
+            ret = false
           end
         end
       end
@@ -1328,36 +1324,34 @@ module Yast
         @commandcache = {}
         @done = !@interactive
         return deep_copy(result)
-      else
-        # if in interactive mode, ask user for input
-        if @interactive
-          loop do
-            newcommand = []
-            newcommand = Scan() while Builtins.size(newcommand) == 0
+      # if in interactive mode, ask user for input
+      elsif @interactive
+        loop do
+          newcommand = []
+          newcommand = Scan() while Builtins.size(newcommand) == 0
 
-            # EOF reached
-            if newcommand.nil?
-              @done = true
-              return { "command" => "exit" }
-            end
-
-            @commandcache = Parse(newcommand)
-            break if !ProcessSystemCommands(@commandcache)
-            break if @done
+          # EOF reached
+          if newcommand.nil?
+            @done = true
+            return { "command" => "exit" }
           end
 
-          return { "command" => @aborted ? "abort" : "exit" } if @done
-
-          # we are not done, return the command asked back to module
-          result = deep_copy(@commandcache)
-          @commandcache = {}
-
-          return deep_copy(result)
-        else
-          # there is no further commands left
-          @done = true
-          return { "command" => "exit" }
+          @commandcache = Parse(newcommand)
+          break if !ProcessSystemCommands(@commandcache)
+          break if @done
         end
+
+        return { "command" => @aborted ? "abort" : "exit" } if @done
+
+        # we are not done, return the command asked back to module
+        result = deep_copy(@commandcache)
+        @commandcache = {}
+
+        return deep_copy(result)
+      else
+        # there is no further commands left
+        @done = true
+        return { "command" => "exit" }
       end
     end
 
@@ -1576,11 +1570,9 @@ module Yast
 
             # if it is not interactive, abort on errors
             Abort() if !Interactive() && res == false
-          else
-            if !Done()
-              Builtins.y2error("Unknown command '%1' from CommandLine", command)
-              next
-            end
+          elsif !Done()
+            Builtins.y2error("Unknown command '%1' from CommandLine", command)
+            next
           end
         end
 
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -488,11 +488,7 @@ module Yast
                 Ops.set(givenoptions, o, i)
               end
             else
-              ret = if v == ""
-                false
-              else
-                TypeRepository.is_a(v, opttype)
-              end
+              ret = (v == "") ? false : TypeRepository.is_a(v, opttype)
 
               if ret != true
                 # translators: error message, %2 is expected type, %3 is the value given
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1276,9 +1276,9 @@ module Yast
 
       res = if type == :nohistory
               Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
-            elsif type == :noecho
+      elsif type == :noecho
               Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
-            else
+      else
               Convert.to_string(SCR.Read(path(".dev.tty")))
       end
 
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1275,11 +1275,11 @@ module Yast
       SCR.Write(path(".dev.tty.prompt"), prompt)
 
       res = if type == :nohistory
-              Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
+        Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
       elsif type == :noecho
-              Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
+        Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
       else
-              Convert.to_string(SCR.Read(path(".dev.tty")))
+        Convert.to_string(SCR.Read(path(".dev.tty")))
       end
 
       # set the default prompt
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1282,9 +1282,9 @@ module Yast
 
       res = if type == :nohistory
         Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
-      elsif type == :noecho
+            elsif type == :noecho
         Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
-      else
+            else
         Convert.to_string(SCR.Read(path(".dev.tty")))
       end
 
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -488,10 +488,10 @@ module Yast
                 Ops.set(givenoptions, o, i)
               end
             else
-              if v == ""
-                ret = false
+              ret = if v == ""
+                false
               else
-                ret = TypeRepository.is_a(v, opttype)
+                TypeRepository.is_a(v, opttype)
               end
 
               if ret != true
@@ -1284,12 +1284,12 @@ module Yast
 
       res = nil
 
-      if type == :nohistory
-        res = Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
+      res = if type == :nohistory
+        Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
       elsif type == :noecho
-        res = Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
+        Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
       else
-        res = Convert.to_string(SCR.Read(path(".dev.tty")))
+        Convert.to_string(SCR.Read(path(".dev.tty")))
       end
 
       # set the default prompt
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1320,13 +1320,7 @@ module Yast
     #  @see #Parse
     def Command
       # if we are done already, return the result
-      if @done
-        if @aborted
-          return { "command" => "abort" }
-        else
-          return { "command" => "exit" }
-        end
-      end
+      return { "command" => @aborted ? "abort" : "exit" } if @done
 
       # there is a command in the cache
       if Builtins.size(@commandcache) != 0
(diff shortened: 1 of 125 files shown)


library/cwm/src/modules/CWM.rb
```diff
@@ -856,7 +856,7 @@ def Run(widgets, functions, skip_store_for: [])
 
         next if ret.nil?
 
-        ret = nil if !validateWidgets(widgets, event_descr) if save
+        ret = nil if save && (!validateWidgets(widgets, event_descr))
```
jreidinger: :+1: it was actually from my POV style bug of ruby conversion

library/commandline/src/modules/CommandLine.rb
```diff
@@ -488,10 +488,10 @@ def Parse(arguments)
...
               end
             else
-              if v == ""
-                ret = false
+              ret = if v == ""
+                false
               else
-                ret = TypeRepository.is_a(v, opttype)
+                TypeRepository.is_a(v, opttype)
               end
```
jreidinger: I worry this is not much readable, but later conversion helps more with ternary operator.

library/control/src/modules/ProductControl.rb
```diff
@@ -572,13 +572,13 @@ def RequiredFiles(stage, mode)
...
+        client = if Stage.firstboot
+          Ops.get_string(m, "name", "dummy")
         else
-          if Builtins.issubstring(Ops.get_string(m, "name", "dummy"), "inst_")
-            client = Ops.get_string(m, "name", "dummy")
+          client = if Builtins.issubstring(Ops.get_string(m, "name", "dummy"), "inst_")
+            Ops.get_string(m, "name", "dummy")
           else
-            client = Ops.add("inst_", Ops.get_string(m, "name", "dummy"))
+            Ops.add("inst_", Ops.get_string(m, "name", "dummy"))
```
jreidinger: this demonstrate that check is not nested...double assign to client and above also having empty string initializer looks very bogus for me.

library/commandline/src/modules/CommandLine.rb
```diff
@@ -488,11 +488,7 @@ def Parse(arguments)
                 Ops.set(givenoptions, o, i)
               end
             else
-              ret = if v == ""
-                false
-              else
-                TypeRepository.is_a(v, opttype)
-              end
+              ret = (v == "") ? false : TypeRepository.is_a(v, opttype)
```
jreidinger: what about `ret = v != "" && TypeRepository.is_a(v, opttype)`

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1282,9 +1282,9 @@ def GetInput(prompt, type)
 
       res = if type == :nohistory
         Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
-      elsif type == :noecho
+            elsif type == :noecho
         Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
-      else
+            else
```
jreidinger: :+1: good change, makes code more readable

library/control/src/modules/ProductControl.rb
```diff
@@ -574,10 +574,10 @@ def RequiredFiles(stage, mode)
         client = ""
         client = if Stage.firstboot
           Ops.get_string(m, "name", "dummy")
-        else
+                 else
           client = if Builtins.issubstring(Ops.get_string(m, "name", "dummy"), "inst_")
```
jreidinger: I expect it also indent line belows, otherwise it looks bogus.

library/network/src/lib/network/susefirewall.rb
```diff
@@ -65,16 +65,16 @@ def SetStartService(start_service)
...
       else
         # without set modified!
         Builtins.y2milestone(
           "start-firewall has been already set to %1",
           start_service
         )
-        Ops.set(@SETTINGS, "start_firewall", start_service)
       end
 
+      Ops.set(@SETTINGS, "start_firewall", start_service)
```
jreidinger: :+1: nice checker

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1281,11 +1281,11 @@ def GetInput(prompt, type)
...
       res = if type == :nohistory
-        Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
+              Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
             elsif type == :noecho
-        Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
+              Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
             else
-        Convert.to_string(SCR.Read(path(".dev.tty")))
+              Convert.to_string(SCR.Read(path(".dev.tty")))
       end
```
jreidinger: ok, now body looks good, so last part I do not like is end alignment :wink:

library/control/src/modules/ProductFeatures.rb
```diff
@@ -214,11 +214,11 @@ def GetFeature(section, feature)
...
       if Ops.is_string?(value)
-        return value
+        value
       elsif Ops.is_boolean?(value)
-        return Convert.to_boolean(value) ? "yes" : "no"
+        Convert.to_boolean(value) ? "yes" : "no"
       else
-        return Builtins.sformat("%1", value)
+        Builtins.sformat("%1", value)
       end
```
jreidinger: in fact you can use case here like
```
value = GetFeature(section, feature)
case value
when String
  value
when true, false
  value ? "yes" : "no"
else
  Builtins.sformat("%1", value)
end
```

library/control/src/modules/ProductFeatures.rb
```diff
@@ -249,11 +244,11 @@ def GetBooleanFeature(section, feature)
...
       if Ops.is_integer?(value)
-        return value
+        value
       elsif Ops.is_string?(value)
-        return Builtins.tointeger(Convert.to_string(value))
+        Builtins.tointeger(Convert.to_string(value))
       else
-        return nil
+        nil
       end
```
jreidinger: same as above

library/control/src/modules/WorkflowManager.rb
```diff
@@ -120,14 +120,14 @@ def main
...
-        return deep_copy(@additional_finish_steps_after_chroot)
+        deep_copy(@additional_finish_steps_after_chroot)
       elsif which_steps == "before_umount"
-        return deep_copy(@additional_finish_steps_before_umount)
+        deep_copy(@additional_finish_steps_before_umount)
       else
         Builtins.y2error("Unknown FinishSteps type: %1", which_steps)
-        return nil
+        nil
       end
```
jreidinger: another opurtinity to use case and maybe use deep_copy only once on result?

library/control/src/modules/WorkflowManager.rb
```diff
@@ -578,12 +573,10 @@ def ReplaceProposalModule(proposal, old, new)
...
 
-          if Ops.is_map?(m)
-            next Builtins.maplist(new) do |it|
-              Builtins.union(Convert.to_map(m),  "name" => it)
-            end
-          else
-            next deep_copy(new)
+          next deep_copy(new) unless Ops.is_map?(m)
+
+          next Builtins.maplist(new) do |it|
```
jreidinger: this and following next is not needed.

library/control/src/modules/WorkflowManager.rb
```diff
@@ -604,12 +597,9 @@ def ReplaceProposalModule(proposal, old, new)
             modules2 = Builtins.maplist(
               Ops.get_list(tab, "proposal_modules", [])
             ) do |m|
-              if m == old
-                next deep_copy(new)
-              else
-                next [m]
-              end
+              next (m == old) ? deep_copy(new) : [m]
```
jreidinger: again next is not needed ( ruby conversion adds it there only to be safe )

library/control/src/modules/WorkflowManager.rb
```diff
@@ -755,20 +745,18 @@ def ReplaceWorkflowModule(workflow, old, new, domain, keep)
...
-          new_list = Builtins.add(new_list, m) if keep
+        found = true
 
-          next deep_copy(new_list)
-        else
-          next [m]
-        end
+        new_list = Builtins.add(new_list, m) if keep
+
+        next deep_copy(new_list)
```
jreidinger: same here. next in block context is like return

library/general/src/modules/Hooks.rb
```diff
@@ -133,11 +133,8 @@ def to_s
       end
 
       def verify!
-        if path.exist?
-          path
-        else
-          raise "Hook search path #{path} does not exists"
-        end
+        return path if path.exist?
+        raise "Hook search path #{path} does not exists"
```
jreidinger: in fact I prefer opposite order as usually it is in verify - check, check, check, fine...so if there is more check in verify in your way it will be harder to extent.
so something like
```
raise "Hook search path #{path} does not exists" unless path.exist?

path
```

library/general/src/modules/ValueBrowser.rb
```diff
@@ -94,9 +94,10 @@ def FormatSimpleType(variable, indent)
     def BrowseTreeHelper(variable, indent)
       variable = deep_copy(variable)
       simple = FormatSimpleType(variable, indent)
-      if !simple.nil?
-        return Item(simple)
-      elsif Ops.is_list?(variable)
+
+      return Item(simple) if !simple.nil?
```
jreidinger: unless ?

library/network/src/modules/CWMFirewallInterfaces.rb
```diff
@@ -232,34 +232,29 @@ def Selected2Opened(ifaces, _nm_ifaces_have_to_be_selected)
...
         end
+        Builtins.y2milestone(
+          "Ifaces left in zone: %1",
+          ifaces_left_explicitely
+        )
+        # there are no interfaces left that would be explicitely mentioned in the EXT zone
+        next [] if ifaces_left_explicitely == []
+
+        # Hmm, some interfaces left
+        next deep_copy(ifaces_also_supported_by_any)
```
jreidinger: next not needed

library/network/src/modules/CWMFirewallInterfaces.rb
```diff
@@ -613,35 +608,30 @@ def CheckPossbilityToChangeFirewall(new_status)
...
+      )
+        # all known ifaces are buggy?
+        return Builtins.size(@buggy_ifaces) != Builtins.size(all_ifaces)
+      else
+        # cancel
+        @buggy_ifaces = deep_copy(@all_interfaces)
+        return false
       end
 
       false
```
jreidinger: this cannot happen

library/network/src/modules/NetworkService.rb
```diff
@@ -274,26 +274,24 @@ def StartStop
...
-      if !@already_asked_for_NetworkManager && network_manager?
-        # TRANSLATORS: pop-up question when reading the service configuration
-        cont = Popup.ContinueCancel(
-          _(
-            "Your network interfaces are currently controlled by NetworkManager\n" \
-              "but the service to configure might not work well with it.\n" \
-              "\n" \
-              "Really continue?"
-          )
+      return true unless !@already_asked_for_NetworkManager && network_manager?
```
jreidinger: please do not use unless with complex term, really hard to understand.

library/sequencer/src/modules/Sequencer.rb
```diff
@@ -133,12 +133,10 @@ def WS_check(aliases, sequence)
       # check all aliases in sequence
       ret0 = Builtins.maplist(sequence) do |key, val|
         if key == "ws_start"
-          if !Ops.is_symbol?(val) && Ops.get(aliases, val).nil?
-            Builtins.y2error(2, "sequencer check: alias not found: %1", val)
-            next false
-          else
-            next true
-          end
+          next true unless !Ops.is_symbol?(val) && Ops.get(aliases, val).nil?
```
jreidinger: same here, please do not use unless for complex conditions

library/system/src/modules/FileChanges.rb
```diff
@@ -170,56 +170,50 @@ def StoreFileCheckSum(file)
...
-              PushButton(Id(:ok), Opt(:okButton), Label.ContinueButton),
-              PushButton(Id(:cancel), Opt(:cancelButton), Label.CancelButton)
-            )
+        _("File %1 has been changed manually.\nYaST might lose some of the changes.\n")
+      end
+
+      msg = Builtins.sformat(msg, Builtins.mergestring(files, ", "))
+      popup_file = "/filechecks_non_verbose"
+
+      return true unless {} ==
```
jreidinger: please do not use trailing form for multiline conditions.

library/system/src/modules/FileChanges.rb
```diff
@@ -231,46 +225,44 @@ def CheckFiles(files)
...
-              popup_file_path,
-              ""
-            )
-          end
-          UI.CloseDialog
-          return ret == :ok
-        else
-          return true
-        end
+      return true unless new_files.size > 0
```
jreidinger: as it use size method, then it is not nil, so I suggest to use `return true if new_files.empty?`

library/system/src/modules/FileChanges.rb
```diff
@@ -231,46 +225,44 @@ def CheckFiles(files)
...
+
+      # Continue/Cancel question, %s is a file name
+      msg = _("File %s has been created manually.\nYaST might lose this file.")
+      if new_files.size > 1
+        # Continue/Cancel question, %s is a comma separated list of file names
+        msg = _(
+          "Files %s have been created manually.\nYaST might lose these files."
+        )
+      end
```
jreidinger: to be consistent above you use if/else construct, so you can also use it here like 
```
if new_files.size == 1
  ...
else
  ...
end

File: library/types/src/modules/IP.rb [hunk 24]
```diff
@@ -218,15 +218,10 @@ def BitsToIPv4(bits)
...
+      return false if network.nil? || network == ""
 
-        # all networks
-      elsif network == "0/0"
-        return true
-      end
-
-      nil
+      # all networks
+      network == "0/0" ? true : nil
```
jreidinger: nil looks like bug for me :)

library/types/src/modules/IP.rb
```diff
@@ -240,21 +235,18 @@ def CheckNetworkShared(network)
     #   CheckNetwork("172.55.0.0/33") -> false
     def CheckNetwork4(network)
       generic_check = CheckNetworkShared(network)
-      if !generic_check.nil?
-        return generic_check
+      return generic_check unless generic_check.nil?
```
jreidinger: ugh, ternary usage of boolean, nasty

library/types/src/modules/IP.rb
```diff
@@ -274,22 +266,19 @@ def CheckNetwork4(network)
     #   CheckNetwork("::1/257") -> false
     def CheckNetwork6(network)
       generic_check = CheckNetworkShared(network)
-      if !generic_check.nil?
-        return generic_check
+      return generic_check if !generic_check.nil?
```
jreidinger: unless?

library/types/src/modules/String.rb
```diff
@@ -738,13 +734,12 @@ def FormatFilename(file_path, len)
...
 
-        if ret.size > len
-          # still too long, remove the ellipsis and start a new iteration
-          dir.delete(ellipsis)
-        else
-          # the size is OK
-          break
-        end
+        # the size is OK
+        break unless ret.size > len
```
jreidinger: what about using more direct `break if ret.size <= len`

library/types/src/modules/TypeRepository.rb
```diff
@@ -52,11 +52,9 @@ def main
     def is_a(value, type)
       value = deep_copy(value)
       validator = Ops.get(@types, type)
-      if !validator.nil?
-        return validator.call(value)
-      else
-        Builtins.y2error("Request to validate unknown type %1", type)
-      end
+      return validator.call(value) if !validator.nil?
```
jreidinger: validator is proc object, so simple `if validator` can be used

jreidinger:
reviewed till "Manually fixed Style/GuardClause" now need to look at L3 for a moment.

lslezak:
OK, thanks.

My approach was to keep the code unchanged as much as possible. Therefore I used e.g. `unless` with the same condition instead of using `if` and negating a complex condition (where I could do a mistake possibly).

I'll try to improve it...

library/packages/src/modules/PackageSystem.rb
```diff
@@ -410,38 +410,35 @@ def CheckAndInstallPackages(packages)
...
-            success = Popup.ContinueCancel(
-              # continue/cancel popup
-              _(
-                "Installing required packages failed. If you continue\n" \
-                  "without installing required packages,\n" \
-                  "YaST may not work properly.\n"
-              )
-            )
-          end
+      return success if success
```
jreidinger: so it is in fact return true if success

library/systemd/test/test_helper.rb
```diff
@@ -24,10 +24,11 @@ def stub_systemctl(unit)
 
   def stub_execute(success: true)
     allow(Yast::Systemctl).to receive(:execute).and_return(
-      OpenStruct.new \
+      OpenStruct.new(
         stdout: "success",
         stderr: (success ? "" : "failure"),
         exit:   (success ? 0  : 1)
+      )
```
jreidinger: :+1: much better now

library/general/src/modules/ValueBrowser.rb
```diff
@@ -80,9 +80,7 @@ def FormatSimpleType(variable, indent)
         Builtins.sformat("%2%1 (symbol)", variable, indent)
       elsif Ops.is_path?(variable)
         Builtins.sformat("%2%1 (path)", variable, indent)
-      else
-        nil
-      end
+            end
```
jreidinger: indentation in whole this commit is very bogus

jreidinger: does it mean, that auto-correct doesn't work correctly?

library/network/src/modules/NetworkService.rb
```diff
@@ -125,7 +125,7 @@ def backend_available?(backend)
       PackageSystem.Installed(BACKEND_PKG_NAMES[backend])
     end
 
-    alias_method :is_backend_available, :backend_available?
+    alias is_backend_available backend_available?
```
jreidinger: to be honest, this is one of new cops which I really dislike ( I also disable it in my projects :) . Reason is that alias is very specific and easy to use incorrectly, very dangerous method. On other hand alias_method is common method, with reasonable parameters ( so not method names which looks like it should be called, but with symbol ) and also provide reasonable error if not found.

https://github.com/config-files-api/config_files_api/blob/master/.rubocop.yml#L55

library/network/src/lib/network/susefirewalld.rb
```diff
@@ -815,7 +815,7 @@ def GetIgnoreLoggingBroadcast(_zone)
     #	// Do not log broadcast packetes from DMZ
     #	SetIgnoreLoggingBroadcast ("DMZ", "yes")
     def SetIgnoreLoggingBroadcast(_zone = nil, bcast)
-      bcast = bcast.downcase == "no" ? "broadcast" : "off"
+      bcast = bcast.casecmp("no").zero? ? "broadcast" : "off"
```
jreidinger: to be honest I am not sure if `bcast.casecmp("no").zero?` have at least same readability as `bcast.downcase == "no"`

library/network/src/modules/NetworkPopup.rb
```diff
@@ -65,19 +65,23 @@ def ChooseItem(_title, items, selected)
...
           device_name = Ops.add(Builtins.substring(device_name, 0, 27), "...")
         end
+
         ip_addr = if NetworkInterfaces.GetValue(i, "BOOTPROTO") == "dhcp"
                     # TRANSLATORS: Informs that the IP address is assigned via DHCP
                     _("DHCP address")
                   else
                     NetworkInterfaces.GetValue(i, "IPADDR")
-                  end
+        end
```
jreidinger: this is in fact not so nice. I expect end where it used to be...to have same indentation as if/else.

library/network/src/modules/SuSEFirewallExpertRules.rb
```diff
@@ -172,7 +172,7 @@ def GetListOfAcceptRules(zone)
                           )
                         else
                           ""
-                        end
+          end
```
jreidinger: also here. It doesn't look nice.

library/runlevel/src/modules/Service.rb
```diff
@@ -61,19 +61,19 @@ def call(command_name, service_name)
...
+      when "stop"    then :stop
+      when "enable"  then :enable
+      when "disable" then :disable
+      when "restart" then :restart
+      when "reload"  then :reload
+      when "try-restart" then :try_restart
+      when "reload-or-restart" then :reload_or_restart
+      when "reload-or-try-restart" then :reload_or_try_restart
+      else
+        raise "Command '#{command_name}' not supported"
```
jreidinger: ugh, no, this is plainly wrong, why elsif is aligned with if, but when not aligned with case? this indentation looks broken to me.

library/system/src/modules/FileChanges.rb
```diff
@@ -225,7 +225,7 @@ def CheckFiles(files)
     def CheckNewCreatedFiles(files)
       new_files = files - @file_checksums.keys
 
-      return true unless new_files.size > 0
+      return true unless !new_files.empty?
```
jreidinger: `unless !`? Ah, my eyes hurts :wink:

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1276,9 +1276,9 @@ def GetInput(prompt, type)
 
       res = if type == :nohistory
               Convert.to_string(SCR.Read(path(".dev.tty.nohistory")))
-            elsif type == :noecho
+      elsif type == :noecho
               Convert.to_string(SCR.Read(path(".dev.tty.noecho")))
-            else
+      else
```
jreidinger: I am not sure if this is better, now it looks like relates to res assignment and not to if condition. so :-1:

library/packages/src/modules/PackageSystem.rb
lslezak: Well, yes, but I wanted to keep the original semantics just to not break something by mistake.

library/system/src/modules/FileChanges.rb
lslezak: Actually the correct solution is to use the plural form translation via `n_()`, see [more here](https://github.com/webyast/webyast/wiki/Localization#plural-forms).

jreidinger:
LGTM

PR #479, opened 2016-06-22
Use a high resolution window icon (64x64) (bsc#985432)

lslezak:
...instead of the low resolution (22x22) one.

The standalone icon does not look that bad, but when comparing it with the other applications it looks awful, see this screenshot from [bug 985432](https://bugzilla.suse.com/show_bug.cgi?id=985432):

## Screenshots

I have tested the fix using the Plasma 5 window switcher, the look obviously depends on the configured style.

### The Original Look

This is the original 22x22 icon, it looks a bit fuzzy without much details:

yast_lowres_icon

### The New Look

The new 64x64 icon is more sharp with more details:

yast_hires_icon


gabi2:
LGTM

PR #481, opened 2016-06-30
Firewall fix

jreidinger:


mvidner:
CI is failing because of https://github.com/yast/yast-ruby-bindings/pull/169, we should revert that bit.

We should add a test reproducing the reported bug (importing SuSEFirewallServices before SuSEFirewall).

jreidinger:
jenkins is now green, if you have idea how to write such bug when we load all tests via rspec and not single one due to coveralls, I will welcome this idea, but in general we need to submit this fix ASAP. So if you agree code wise, if you can LGTM it.

ancorgs:
We are a little bit in a hurry and this is good enough. LGTM

jreidinger:
thanks

PR #482, opened 2016-07-10
Define Arch.arm

afaerber:
So far only Arch.aarch64 existed but no 32-bit equivalent.

Will be needed for further yast-bootloader changes.

Signed-off-by: Andreas Färber <afaerber@suse.de>

### DIFF
--- a/library/general/src/modules/Arch.rb
+++ b/library/general/src/modules/Arch.rb
@@ -65,7 +65,7 @@ module Yast
     # ************************************************************
     # system architecture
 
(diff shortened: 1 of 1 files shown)


jreidinger:
change look OK, just do we support 32 bit arm? and only opensuse or also SP2?

afaerber:
openSUSE only. This being a public PR, we shouldn't be limiting us to what SLES supports.

openSUSE actually has both armv6hl and armv7hl, but YaST seems to treat them both as "arm", which so far works okay.

jreidinger:
ok, I mean also opensuse part, as we support both, so if opensuse start supporting it, we should start fixing it.

So lets merge it and in case of problem, we will contact you.

PR #483, opened 2016-07-13
improve documentation of auto client workflow

jreidinger:
Related to https://github.com/yast/yast-bootloader/pull/350


mvidner:
Thanks! I will add a commit with English improvements.

mvidner:
(The Travis failures are because of the new RuboCop but that is being solved in #478)

LGTM

PR #487, opened 2016-08-04
Declare textdomain to fix untranslated texts

mvidner:
found by y2tool y2makepot


lslezak:
Version update not needed?

mvidner:
OK, I've made a generic bug for the translations: https://bugzilla.opensuse.org/show_bug.cgi?id=992084

lslezak:
LGTM

PR #488, opened 2016-08-08
Fix cd/dvd URLs reassembling

imobachgs:
Before this patch:

```ruby
url = "cd:/?device=/dev/sr0"
Yast::URL.Build(Yast::URL.Parse(url)) # => "cd://?device=/dev/sr0"
```

After this patch it returns `cd:/?device=/dev/sr0`.

### DIFF
--- a/library/types/test/url_test.rb
+++ b/library/types/test/url_test.rb
@@ -25,6 +25,66 @@ describe Yast::URL do
     it "returns a hash containing the token extracted from the URL" do
       expect(subject.Parse(url)).to eq(tokens)
     end
+
+    context "given a CD/DVD with a file" do
+      let(:url) { "cd:/some/file" }
+      let(:tokens) do
+        {
+          "scheme"   => "cd",
+          "host"     => "",
+          "path"     => "/some/file",
+          "fragment" => "",
+          "user"     => "",
+          "pass"     => "",
+          "port"     => "",
+          "query"    => ""
+        }
+      end
+
+      it "returns a hash containing 'scheme' and 'path'" do
+        expect(subject.Parse(url)).to eq(tokens)
+      end
+    end
+
+    context "given a CD/DVD with a device" do
+      let(:url) { "cd:/?device=/dev/sr0" }
+      let(:tokens) do
+        {
+          "scheme"   => "cd",
+          "host"     => "",
+          "path"     => "/",
+          "fragment" => "",
+          "user"     => "",
+          "pass"     => "",
+          "port"     => "",
+          "query"    => "device=/dev/sr0"
+        }
+      end
+
+      it "returns a hash containing 'scheme', 'path' and 'query'" do
+        expect(subject.Parse(url)).to eq(tokens)
+      end
+    end
+
+    context "given a CD/DVD with a device and a path" do
+      let(:url) { "cd:/some/file?device=/dev/sr0" }
+      let(:tokens) do
+        {
+          "scheme"   => "cd",
+          "host"     => "",
+          "path"     => "/some/file",
+          "fragment" => "",
+          "user"     => "",
+          "pass"     => "",
+          "port"     => "",
+          "query"    => "device=/dev/sr0"
+        }
+      end
+
+      it "returns a hash containing 'scheme', 'path' and 'query'" do
+        expect(subject.Parse(url)).to eq(tokens)
+      end
+    end
   end
 
   describe ".Build" do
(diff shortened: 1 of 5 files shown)


lslezak:
BTW What does it return for `dvd:///` or `cd:///`?

imobachgs:
@lslezak I guess it's libzypp. I've tried this in a SP4:

```
# zypper ar cd:/?device=/dev/sr0 DVD
This is a changeable read-only media (CD/DVD), disabling autorefresh.
Adding repository 'DVD' [done]
Repository 'DVD' successfully added
Enabled: Yes
Autorefresh: No
GPG check: Yes
URI: cd:///?device=/dev/sr0
```

Note the resulting URI.

lslezak:
Yes, I know this zypper transformation, I was asking what happens to `Yast::URL.Build(Yast::URL.Parse("cd:///"))` if you use `install=cd:///` parameter (if that works with linuxrc...).

Also the question is if we need to call `Yast::URL.Build(Yast::URL.Parse(url))`, could we use the original `url` value directly?

imobachgs:
Linuxrc documented parameter should look like `install=cd:/`, but I've tried to use three slashes and it looks like it works.

If you use `Yast::URL.Build(Yast::URL.Parse("cd:///"))` you'll get `cd:/` (which looks ok to me). To keep the number of `slashes` we should store that information in the hash we get while parsing.

On the other hand, I agree that we should avoid doing such stuff like parsing/building, but in this case [it looks reasonable](https://github.com/yast/yast-packager/blob/master/src/modules/AddOnProduct.rb#L1525).

lslezak:
I see, that looks OK...

One more question: How does it work with a path in a cd/dvd URL? E.g. `dvd:/dir` or `dvd:///dir`? Is that parsed/built correctly? Maybe we should add a test for that as well to avoid regressions...

imobachgs:
I expect it to work. But I'll add a testcase to be on the safe side. Thanks!

imobachgs:
Oh, that's funny (current behavior, without the patch):

```ruby
Yast::URL.Parse("dvd:/dir")
# => 
# {"scheme"=>"dvd",
#  "path"=>"",
#  "query"=>"",
#  "fragment"=>"",
#  "user"=>"",
#  "pass"=>"",
#  "port"=>"",
#  "host"=>"dir"} <- host?
 ```

So if we make `Yast::URL.Build(Yast::URL.Parse("dvd:/dir"))` will get `dvd://dir`. With the patch it will work (but just as a coincidence because Parse is broken anyway).

The only way to make the `Yast::URL.Parse` to work correctly is using three slashes:

```ruby
Yast::URL.Parse("dvd:///dir")
# => 
# {"scheme"=>"dvd",
#  "path"=>"/dir", <- OK
#  "query"=>"",
#  "fragment"=>"",
#  "user"=>"",
#  "pass"=>"",
#  "port"=>"",
#  "host"=>""}
```

lslezak:
Um, I do not know if `dvd:/dir` actually works in linuxrc...

Anyway, URI in Ruby parses it correctly:

```
irb(main):011:0> URI("dvd:/dir").path
=> "/dir"
irb(main):012:0> URI("dvd:/dir").host
=> nil
```

In the long term we should drop `Yast::URL` and use the native `URI`.

But if it works with the patch let's use it for SP2 for now.

lslezak:
IMO the problem is that linuxrc and libzypp URLs are bit different and in addition to that the YaST parser/builder between them is broken. :disappointed: 

We should unify linuxrc, libzypp and YaST to use one compatible URL style.

imobachgs:
Ok, I've changed the overall approach and I've fixed the parsing (instead of the building). The code is only applied if we're dealing with `dvd` and `cd` schemes, nothing more.

On the other hand, I've added a set of `rebuilding URL` tests to check some typical scenarios:

```ruby
    URLS = {
      "dvd:/dir"                        => "dvd:///dir",
      "dvd://dir"                       => "dvd:///dir",
      "dvd:///dir"                      => "dvd:///dir",
      "cd:/?device=/dev/sr0"            => "cd:///?device=/dev/sr0",
      "cd:/some/file?device=/dev/sr0"   => "cd:///some/file?device=/dev/sr0",
      "cd:///some/file?device=/dev/sr0" => "cd:///some/file?device=/dev/sr0",
      "http://u:p@suse.de/a#b"          => "http://u:p@suse.de/a#b",
      "ftp://u:p@suse.de/a#b"           => "ftp://u:p@suse.de/a#b",
      "slp:/"                           => "slp://"
    }
```

library/types/test/url_test.rb
```diff
@@ -25,23 +25,123 @@
...
+      "cd:///some/file?device=/dev/sr0" => "cd:///some/file?device=/dev/sr0",
+      "http://u:p@suse.de/a#b"          => "http://u:p@suse.de/a#b",
+      "ftp://u:p@suse.de/a#b"           => "ftp://u:p@suse.de/a#b",
+      "slp:/"                           => "slp://"
+    }.freeze
 
-      it "returns a URL containing which a single slash to separate the schema from the rest" do
-        expect(subject.Build(tokens)).to eq("cd:/?device=/dev/sr0")
+    URLS.each do |url, rebuilt|
+      context "given #{url}" do
```
lslezak: NP: IMHO the context is misused here, the context should define the environment where the test is run (something like "in installation", "on S390",...).

The test input should be part of the `it`, e.g. `it "returns '#{rebuilt}' for '#{url}'" do ...`

lslezak:
LGTM, thanks! :thumbsup:

imobachgs:
Thanks!

PR #489, opened 2016-08-09
Popup.Error will escape tags when using richtext

imobachgs:
This patch fixes [bsc#992506](https://bugzilla.suse.com/show_bug.cgi?id=992506). When trying to contact a non-valid SCC/SMT server, YaST didn't escaped the error message.

not-escaped-error

After this fix, this situation should be handled properly:

escaped-error

The bug was introduced in this older [PR](https://github.com/yast/yast-yast2/pull/442). The reason is that we have two different use cases here:

* I want to show an error and we should escape the content (potentially unsafe).
* I want to show a long message and I don't want the content to be escaped (supposed to be safe).

So now we're escaping in case of showing an error. AutoYaST timed dialogs won't be affected as the implementation is different and does not try to switch to richtext in long error messages.

### DIFF
--- a/library/general/src/modules/Popup.rb
+++ b/library/general/src/modules/Popup.rb
@@ -48,6 +48,7 @@ module Yast
       Yast.import "Label"
       Yast.import "Mode"
       Yast.import "Directory"
+      Yast.import "String"
 
       @feedback_open = false
 
--- a/library/general/src/modules/Popup.rb
+++ b/library/general/src/modules/Popup.rb
@@ -48,7 +48,6 @@ module Yast
       Yast.import "Label"
       Yast.import "Mode"
       Yast.import "Directory"
-      Yast.import "String"
 
       @feedback_open = false
 
@@ -891,7 +890,7 @@ module Yast
               VSpacing(0.5),
               RichText(
                 Builtins.mergestring(
-                  Builtins.splitstring(String.EscapeTags(details), "\n"),
+                  Builtins.splitstring(ERB::Util.html_escape(details), "\n"),
                   "<br>"
                 )
               ),
--- a/library/general/src/modules/Popup.rb
+++ b/library/general/src/modules/Popup.rb
@@ -36,6 +36,7 @@
 # <br>
 # See also <a href="../wizard/README.popups">README.popups</a>
 require "yast"
+require "erb"
 
 module Yast
   class PopupClass < Module
@@ -47,7 +48,6 @@ module Yast
       Yast.import "Label"
       Yast.import "Mode"
       Yast.import "Directory"
-      Yast.import "String"
 
       @feedback_open = false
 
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Popup.rb
```diff
@@ -1325,7 +1325,7 @@ def Error(message)
           Ops.greater_than(Builtins.size(lines), @too_many_lines)
         anyMessageInternalRich(
           Label.ErrorMsg,
-          message,
+          String.EscapeTags(message),
```
lslezak: I'd prefer native Ruby [ERB::Util#html_escape](https://ruby-doc.org/stdlib-2.1.2/libdoc/erb/rdoc/ERB/Util.html#method-c-html_escape) here instead of the YaST own implementation.

library/general/test/popup_test.rb
```diff
@@ -62,9 +63,40 @@
...
+    context "when switch to richtext is allowed" do
+      let(:switch_to_richtext) { true }
+
+      it "escapes the tags if message is too long" do
+        message = line * limit
+        expect(subject).to receive(:RichText).with(Yast::String.EscapeTags(message))
+        subject.Error(message)
+      end
+
+      it "does not escape the tags if message is not too long" do
```
lslezak: NP: Too many `not` makes it less readable, IMHO *"keeps the original text if the message is short"* is better.

library/general/test/popup_test.rb
```diff
@@ -62,9 +63,40 @@
...
     before { allow(ui).to receive(:OpenDialog) }
 
-    it "shows a popup without escaping tags" do
-      expect(subject).to receive(:RichText).with("<h1>Title</h1>")
-      subject.Error("<h1>Title</h1>")
+    let(:switch_to_richtext) { true }
+    let(:line) { "<h1>Title</h1>\n" }
+    let(:limit) { subject.too_many_lines }
+
+    around do |example|
```
lslezak: I'd add a short comment before this `around` block:

```ruby
# backup and restore the original "switch_to_richtext" flag
```

library/general/test/popup_test.rb
```diff
@@ -1,8 +1,10 @@
 #! /usr/bin/env rspec
 
 require_relative "test_helper"
+require "erb"
 
 Yast.import "Popup"
+Yast.import "String"
```
lslezak: This can be removed now, right?

library/general/test/popup_test.rb
```diff
@@ -62,9 +63,41 @@
...
     before { allow(ui).to receive(:OpenDialog) }
 
-    it "shows a popup without escaping tags" do
-      expect(subject).to receive(:RichText).with("<h1>Title</h1>")
-      subject.Error("<h1>Title</h1>")
+    let(:switch_to_richtext) { true }
+    let(:line) { "<h1>Title</h1>\n" }
+    let(:limit) { subject.too_many_lines }
+
+    # Backup and restore the original switch_to_rich flag
```
lslezak: NP: the `switch_to_richtext` name got truncated

lslezak:
LGTM :tada:

imobachgs:
Thanks!

PR #491, opened 2016-08-24
Packages: remove warning Icon from package callbacks.

teclator:
How it looks before:

screenshot_sle12_update_2016-08-26_12 45 25

How looks now:

screenshot_sle12_update_2016-08-26_12 43 26

### DIFF
--- a/library/packages/src/modules/PackageCallbacks.rb
+++ b/library/packages/src/modules/PackageCallbacks.rb
@@ -2989,9 +2989,6 @@ module Yast
         HSpacing(50), # enforce width
         VSpacing(0.1),
         HBox(
-          # maybe more icon types could be used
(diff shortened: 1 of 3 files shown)


imobachgs:
LGTM

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -2989,9 +2989,6 @@ def layout_popup(message, button_box, info_on)
         HSpacing(50), # enforce width
         VSpacing(0.1),
         HBox(
```
lslezak: I think you can remove this `HBox` as it is not needed anymore. But double check how it would look...

lslezak:
LGTM

PR #492, opened 2016-08-25
[Review] Request from 'schubi2' @ 'yast/yast-yast2/review_160825_timeout_error_reports'

schubi2:
This PR also belongs to https://github.com/yast/yast-autoinstallation/pull/244
  * 8ec9e99 packaging
  * d5f10b5 cleanup Reports flag. espl. for AutoYaST


library/general/test/report_test.rb
```diff
@@ -121,4 +124,58 @@
       expect(subject.GetMessages(0, 1, 0, 0)).to match(/Message/)
     end
   end
+
+  describe ".Settings" do
```
imobachgs: You're checking `.Export` instead of `.Settings`, right?

library/general/test/report_test.rb
```diff
@@ -121,4 +124,58 @@
...
+    end
+
+    context "while AutoYaST installation" do
+      before(:each) do
+        Yast::Profile.ReadXML(ay_profile)
+        Yast::Mode.SetMode("autoinstallation")
+        subject.main()
+      end
+
+      it "check default entries" do
```
imobachgs: I would prefer a more behaviour-oriented text like _sets default values_.

imobachgs:
Codewise it looks good; just make rubocop happy.

library/general/test/report_test.rb
schubi2: Hm, not only Export but also default settings while normal installation. True, I am using Export for checking the values :-)

library/general/test/report_test.rb
```diff
@@ -121,4 +123,58 @@
...
+  describe ".Settings" do
+    DATA_PATH =  File.join(File.dirname(__FILE__), 'data')
+    let(:ay_profile) { File.join(DATA_PATH, 'ay_profile.yml') }
+    let(:default_normal) { File.join(DATA_PATH, 'default_normal_installation.yml') }
+    let(:default_ay) { File.join(DATA_PATH, 'default_ay_installation.yml') }
+    let(:result_ay) { File.join(DATA_PATH, 'ay_installation.yml') }
+
+    context "while normal installation" do
+      it "check default entries" do
+        Yast::Mode.SetMode("installation")
```
lslezak: Um, I'd avoid setting the mode directly, this changes the global state outside the tests. It might have some side effects...

Simply mocking `Mode.mode` should work I think.
```ruby
allow(Yast::Mode).to receive(:mode).and_return("installation")
```

library/general/test/report_test.rb
```diff
@@ -121,4 +123,58 @@
       expect(subject.GetMessages(0, 1, 0, 0)).to match(/Message/)
     end
   end
+
+  describe ".Settings" do
+    DATA_PATH =  File.join(File.dirname(__FILE__), 'data')
+    let(:ay_profile) { File.join(DATA_PATH, 'ay_profile.yml') }
```
lslezak: The `YAML.load_file` call is repeated quite a lot in this file, you can put it directly here instead of repeating it the tests:
```ruby
let(:ay_profile) { YAML.load_file(File.join(DATA_PATH, 'ay_profile.yml')) }
```
```ruby
expect(subject.Export()).to match(ay_profile)
```

imobachgs:
LGTM

PR #493, opened 2016-09-16
Removed build dependency on yast2-perl-bindings (bsc#999203)

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=999203

The remaining 4 pm files in this package do not participate in build
time tests.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Fri Sep 16 07:23:10 UTC 2016 - mvidner@suse.com
+
+- Removed build dependency on yast2-perl-bindings (bsc#999203)
+
+-------------------------------------------------------------------
 Fri Aug 26 10:37:45 UTC 2016 - kanderssen@suse.com
 
 - Packages: remove warning icon from package callbacks.
--- a/package/yast2.spec
+++ b/package/yast2.spec
@@ -42,7 +42,6 @@ BuildRequires:  update-desktop-files
 # Needed already in build time
 BuildRequires:  yast2-core >= 2.18.12
 BuildRequires:  yast2-devtools >= 3.1.10
-BuildRequires:  yast2-perl-bindings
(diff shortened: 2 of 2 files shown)


jreidinger:
Please check also other build requires, e.g. libyui bindings and yast2-core looks suspicious.

mvidner:
OK, I'll make a correct dependency graph from OBS data first (via `osc whatdependson`), so that we don't eliminate dependencies which are far from the critical path.

jreidinger:
@mvidner I will adapt my script to generate proper output with osc whatdependson, but still I think that for future it make sense to ensure we have correct dependencies in spec file.

mvidner:
Thanks to your script we now have [accurate dependency data](https://github.com/mvidner/rpm-build-dependencies/blob/e7143675b697dfc5bfb756a4914990b6b111e77d/yast_deps.yaml#L364).
yast2.rpm is in the 5th layer and its requirements are
```
yast2 ~ 131s:
  - libyui [0]
  - libzypp [0]
  - rubygem-cfa [0]
  - rubygem-rspec [0]
  - rubygem-rspec-core [0]
  - rubygem-rspec-expectations [0]
  - rubygem-rspec-mocks [0]
  - yast2-core [1]
  - yast2-devtools [0]
  - yast2-perl-bindings [4] # This PR removes this dependency and moves yast.rpm to layer 4
  - yast2-pkg-bindings [2]
  - yast2-ruby-bindings [3]
  - yast2-testsuite [1]
  - yast2-ycp-ui-bindings [2]
```

So the order in which we should try removing further requirements is
- yast2-ruby-bindings
- yast2-ycp-ui-bindings, yast2-pkg-bindings
- yast2-core, yast2-testsuite

I'll check that.

mvidner:
Aaand, the ruby bindings are obviously quite hard to remove , since they contain the YCP compatibility libraries.

jreidinger:
LGTM

PR #494, opened 2016-09-16
Locate returns interface names instead of types (bsc#998717)

teclator:
During the cleanup of NetworkInterfaces the behaviour of Locate method was changed by error as you can see here: 

https://github.com/yast/yast-yast2/pull/433/files#diff-9e2299f8f1f2334535a5e511b4383b4aR181

This just give the original behavior, fixing also current method tests.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,8 +1,10 @@
 -------------------------------------------------------------------
 Fri Sep 16 17:53:15 UTC 2016 - kanderssen@suse.com
 
-- Network: Locate returns interface names instead of type.
-  (bsc#998717)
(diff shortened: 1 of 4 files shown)


package/yast2.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Sep 16 17:53:15 UTC 2016 - kanderssen@suse.com
+
+- Network: Locate returns interface names instead of type.
```
mchf: as martin always says, changelog note should describe the issue from the user's POV

mchf:
Despite of changelog entry it LGTM

mchf:
LGTM

PR #498, opened 2016-09-28
save_y2logs: use canonical path (bsc#1001454)

wfeldt:

PR #499, opened 2016-09-28
save_y2logs: use canonical path (bsc#1001454)

wfeldt:
Same as https://github.com/yast/yast-yast2/pull/498 but for SLE-12-SP2 branch.


imobachgs:
LGTM

PR #500, opened 2016-10-06
Network: adapt old network configuration when needed (bsc#962824)

teclator:
This is an alternative and hope than an easiest solution to what I was trying to solve in yast/yast-network#441

Basically the problem is that some years ago the way to configure an enslaved interface was assigning to it the a static `0.0.0.0` IP.
```
BOOTPROTO="static"
IPADDR="0.0.0.0/32"
```

But since SLE10 is was not need but used until it was fixed in SLE12. So currently if you want to enlsave an interface the BOOTPROTO should be modified to none and the IPADDR, that could be something like
```
BOOTPROTO="none"
IPADDR=""
NETMASK=""
PREFIXLEN=""
```

With this PR the configuration is adapted when the files are read and in just in case that the user change something in the interface and apply the changes, only in that case the files should be written.


mchf:
As I said, I'd prefer to do any magic in yast2-network. This should just read ifcfg, any additional handling should be done in upper layers. In this case e.g. in LanItems#Read. There are other consumers of NetworkInterfaces' output, like OES scripts, and this change can corrupt them.

teclator:
I agree, just decide where should be the best place to do it and what to do. For example calling an adapt method just after `NetworkInterfaces.Read` is called in `LanItems.Read` working over the **NetworkInterfaces** should also produce the same results (well at least from the lan client POV) without polluting **NetworkInterfaces API**. I have to check when and where the NetworkInterfaces could be read again.

mchf:
lgtm

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -758,6 +758,29 @@ def generate_config(pth, values)
...
 
+    # Adapts the interface configuration used during many year for enslaved
+    # interfaces (IPADDR == 0.0.0.0 and BOOTPROTO == 'static').
+    #
+    # Sets the BOOTPROTO as none, empties the IPADDR, and also empties the
+    # NETMASK and the PREFIXLEN if exist.
+    def adapt_old_config!
+      @Devices.each do |devtype, devices|
+        devices.each do |device, config|
+          bootproto = config["BOOTPROTO"].to_s || "static"
```
mchf: NP: I think that ```config["BOOTPROTO"].to_s``` is never false so ```|| "static"``` is useless

teclator: True :+1:

PR #503, opened 2016-10-18
Add-on module: Do not escape characters like ":" in the path

schubi2:


library/types/test/url_test.rb
```diff
@@ -22,8 +22,18 @@
...
+    context "given a http(s) URL" do
+      it "returns a hash containing the token extracted from the URL" do
+        expect(subject.Parse(url)).to eq(tokens)
+      end
+
+      it "returns url with changed user" do
+        url = subject.Parse(
+          "http://name:pass@www.suse.cz:80/path/index.html?question#part"
+        )
+        url["user"] = "user@domain"
```
jreidinger: I expect that you here also verify that it correctly escape `:` which you modify. So try `user:1@domain` which shows that `:` is properly escaped. I know it is converted old testsuite, but you should extend it to ensure your change do not break it.

jreidinger:
LGTM

PR #507, opened 2016-11-03
Fix not show

jreidinger:


library/packages/src/modules/SignatureCheckDialogs.rb
```diff
@@ -1197,6 +870,350 @@ def UseFileWithUnknownDigest(filename, digest, dont_show_dialog_ident)
     publish function: :ImportGPGKeyIntoTrustedDialog, type: "boolean (map <string, any>, integer)"
     publish function: :UseFileWithWrongDigest, type: "boolean (string, string, string, string)"
     publish function: :UseFileWithUnknownDigest, type: "boolean (string, string, string)"
+
+  private
+
+    # helper to strip download path. It uses internal knowledge that download
+    # prefix ends in TmpDir.* zypp location
```
lslezak: :+1: for the valuable comment

library/packages/src/modules/SignatureCheckDialogs.rb
```diff
@@ -1197,6 +870,350 @@ def UseFileWithUnknownDigest(filename, digest, dont_show_dialog_ident)
...
+      end
+
+      splittedstring
+    end
+
+    # Returns term with message icon
+    #
+    # @param string message type "error", "warning" or "question"
+    # @return [Yast::Term] `Image(...) with margins
+    def MessageIcon(msg_type)
```
lslezak: Ken requested removing all icons from popups, maybe we should drop this when touching this anyway...

library/packages/src/modules/SignatureCheckDialogs.rb
```diff
@@ -1197,6 +870,350 @@ def UseFileWithUnknownDigest(filename, digest, dont_show_dialog_ident)
     publish function: :ImportGPGKeyIntoTrustedDialog, type: "boolean (map <string, any>, integer)"
     publish function: :UseFileWithWrongDigest, type: "boolean (string, string, string, string)"
     publish function: :UseFileWithUnknownDigest, type: "boolean (string, string, string)"
+
+  private
+
+    # helper to strip download path. It uses internal knowledge that download
+    # prefix ends in TmpDir.* zypp location
+    def strip_download_prefix(path)
+      path.sub(/^\/.*\/TmpDir\.[^\/]+\//, "")
```
lslezak: NP: a new-line char is really unlikely, but `\A` should be preferred here...

PR #509, opened 2016-11-07
Add to fstab parser only optional ending line otherwise it can cause

jreidinger:
data loss (bsc#429326#c11)

Downside of this change is that it can parse also handfully crafted
entries that have two entries on one line, but choosing between invalid
read of valid file and parsing broken file as valid, I choose to always
read properly valid entries.


imobachgs:
LGTM

library/general/test/agents_test/test_root/etc/fstab
```diff
@@ -1,7 +1,7 @@
-UUID=b66c1028-cd4a-4c08-b92d-fc4b9840845d       /       ext4    noatime,data=writeback,acl,user_xattr 1 1
-UUID=d2811ace-66e0-4ef2-9b68-9b2758359391       /home   ext4    noatime,data=writeback,acl 1 2
-192.168.1.2:/home/kv    /home/kv2       nfs     defaults 0 0
-192.168.1.2:/media/new2 /media/new2     nfs     defaults 0 0
-192.168.1.2:/media/new  /media/new      nfs     defaults 0 0
+UUID=b66c1028-cd4a-4c08-b92d-fc4b9840845d	/	ext4	noatime,data=writeback,acl,user_xattr 1 1
```
mvidner: This is the file *after* it has been written by the agent. Notice the trailing whitespace (added by <code>\`Optional(`Whitespace())</code>) and the last line *does* have a newline so it does not test your fix.

jreidinger: @mvidner damn, it is when I add comment, vim add it automatic :( will modify it

jreidinger: @mvidner and file is at it is in bug report, just with added comment line. So user probably already try it

jreidinger:
travis failed due to old yast2-core

mvidner:
OK, tests are green, but I don't understand the point of this.
Also, shouldn't we *add* a test instead of *replacing* it?

mvidner:
It seems you've copied the fixture data wrong.

PR #516, opened 2016-11-18
Restart file cleanup

lslezak:
Remove the restart file (usually `/var/lib/YaST2/restart_yast`) if it already exists before starting the module to  avoid possible infinite loop. Fixes [bug #842910](https://bugzilla.suse.com/show_bug.cgi?id=842910).

The file might be accidentally left in the system when YaST crashes or system shuts down unexpectedly before YaST removes that restart file.

How to reproduce the issue:

```
touch /var/lib/YaST2/restart_yast
yast2 lan
# now you cannot escape from the module, it will keep starting again and again
```

### DIFF
--- a/scripts/yast2
+++ b/scripts/yast2
@@ -422,6 +422,11 @@ else
 	*) REDO_FILE=/var/lib/YaST2/restart_yast ;;
     esac
     snapshot_pre $module
+
+    # remove the redo file if it already exists before starting the module,
+    # avoid possible infinite loop (bsc#842910)
+    rm -f "$REDO_FILE"
+
     #  break out on errors, #343258
     while [ $exit_code = 0 ]; do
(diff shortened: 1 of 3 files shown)


scripts/yast2
```diff
@@ -422,6 +422,11 @@ else
 	*) REDO_FILE=/var/lib/YaST2/restart_yast ;;
     esac
     snapshot_pre $module
+
+    # remove the redo file if it already exists before starting the module,
+    # avoid possible infinite loop (bsc#842910)
+    rm -f "$REDO_FILE"
+
```
jreidinger: or does it take in count only when y2base create new one? in such case it make sense

lslezak: Yes, this should fix only the case when the file already exists *before* starting the module (via `y2base`).

The condition still applies when the module started from `y2base` creates the restart file and wants to restart YaST.

jreidinger:
LGTM

PR #517, opened 2016-11-23
Fix workflow module replacement (bsc#1011869)

imobachgs:
Fix [bsc#1011869](https://bugzilla.suse.com/show_bug.cgi?id=1011869). This bug was introduced during a [rubocop update](https://github.com/yast/yast-yast2/pull/478/files#diff-ee72b5fab6634396aa4067ae6b3800afR751).


imobachgs:
The test coverage decreased, but the PR contains a new test for the changed code. It looks like the problem is that `FileUtils.rb` is now loaded (because of `WorkflowManager` test).

PR #519, opened 2016-11-30
PackagesProposal - Added API for optional resolvables

lslezak:
This API enhancement is needed by `yast2-packager` to fix blocked installation after deselecting some default patterns in openQA tests (YaST complains about missing patterns and blocks the installation).

- There are now two kinds of resolvables: *required* and *optional*, the optional ones are stored separately so the software proposal module can only complain for the missing required resolvables. User can deselect the optional resolvables without any problem.
- The API is backward compatible, the added method parameter is optional with the backward compatible default value. We do not have to change the existing code.
- Code cleanup - removed all `Builtins` and `Convert` calls
- Refactoring - more Ruby-like code
- Added unit tests for the `PackagesProposal` module (covers 100% of the module)
- Improved the code documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,8 +32,11 @@ library/*/testsuite/*.sum
 library/*/testsuite/*.exp
 library/*/testsuite/*.bak
 library/*/test/*.log
+library/*/test/*/*.log
 library/*/test/*.trs
+library/*/test/*/*.trs
 /test-driver
+coverage
 
 doc/yast.8
 
--- a/.gitignore
+++ b/.gitignore
@@ -15,11 +15,8 @@ configure.ac
 install-sh
 pluglib-bindings.ami
 */.dep
-library/*/.dep
-library/*/*/.dep
-library/*/*.ybc
-library/*/*/*.ybc
-library/*/*/*/*.ybc
+library/**/.dep
+library/**/*.ybc
 doc/autodocs
 library/*/doc/autodocs
 library/*/testsuite/config/
@@ -31,10 +28,8 @@ library/*/testsuite/*.log
 library/*/testsuite/*.sum
 library/*/testsuite/*.exp
 library/*/testsuite/*.bak
-library/*/test/*.log
-library/*/test/*/*.log
-library/*/test/*.trs
-library/*/test/*/*.trs
+library/*/test/**/*.log
+library/*/test/**/*.trs
 /test-driver
 coverage
 
--- a/library/general/src/modules/PackagesProposal.rb
+++ b/library/general/src/modules/PackagesProposal.rb
@@ -55,11 +55,7 @@ module Yast
 
     # Resets all resolvables to install (both required and optional). Use carefully.
     def ResetAll
-      if !@resolvables_to_install.empty? || !@opt_resolvables_to_install.empty?
-        log.warn("Resetting all PackagesProposal items")
-      else
-        log.info("Resetting already empty items")
-      end
+      log.info("Resetting all PackagesProposal items")
 
       @resolvables_to_install.clear
       @opt_resolvables_to_install.clear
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -32,8 +32,11 @@ library/*/testsuite/*.sum
 library/*/testsuite/*.exp
 library/*/testsuite/*.bak
 library/*/test/*.log
+library/*/test/*/*.log
```
jreidinger: what about modifying above to `library/*/test/**/*.log`

.gitignore
```diff
@@ -32,8 +32,11 @@ library/*/testsuite/*.sum
...
 library/*/test/*.trs
+library/*/test/*/*.trs
```
jreidinger: and same here

library/general/src/modules/PackagesProposal.rb
```diff
@@ -44,20 +46,23 @@ def main
...
       @supported_resolvables = [:package, :pattern]
     end
 
-    # Resets all resolvables to install. Use carefully.
+    # Resets all resolvables to install (both required and optional). Use carefully.
     def ResetAll
-      if @resolvables_to_install != {}
-        Builtins.y2warning("Reseting all PackagesProposal items")
+      if !@resolvables_to_install.empty? || !@opt_resolvables_to_install.empty?
+        log.warn("Resetting all PackagesProposal items")
```
jreidinger: why it is warning? you call method to reset it and it warn if it reset something?

library/general/src/modules/PackagesProposal.rb
```diff
@@ -73,34 +78,29 @@ def GetSupportedResolvables
 
     def IsSupportedResolvableType(type)
       if type.nil?
-        Builtins.y2error("Wrong type: %1", type)
+        log.error("Type cannot be nil")
```
jreidinger: I think below return is not needed and can be only one liner.

library/general/src/modules/PackagesProposal.rb
```diff
@@ -144,35 +142,20 @@ def CheckParams(unique_ID, type)
...
-            Ops.get(@resolvables_to_install, [unique_ID, type], []),
-            resolvables
-          ),
-          from: "list",
-          to:   "list <string>"
-        )
-      )
+      log.info("Adding #{log_label(optional)} #{resolvables} of type #{type} for #{unique_ID}")
+
+      data(optional)[unique_ID][type].concat(resolvables)
```
jreidinger: it would be nice if `CreateEmptyStructureIfMissing(unique_ID, type, optional: optional)` just return that `data(optional)[unique_ID][type]` so maybe better name would be `data_for_uid_and_type` which create it if needed.

library/general/src/modules/PackagesProposal.rb
```diff
@@ -278,35 +253,29 @@ def GetResolvables(unique_ID, type)
...
-        end
+      data(optional).each do |_unique_ID, resolvables|
+        ret.concat(resolvables[type]) if resolvables.key?(type)
       end
 
+      # sort the result and remove the duplicates
+      ret.sort!
+      ret.uniq!
+
       deep_copy(ret)
```
jreidinger: is deep copy needed? do you expect someone will modify strings?

.gitignore
lslezak: Yes, that's better, I didn't know that `**` is supported...

library/general/src/modules/PackagesProposal.rb
```diff
@@ -326,31 +295,28 @@ def GetAllResolvables(type)
...
+      GetSupportedResolvables().each do |one_type|
+        resolvables = GetAllResolvables(one_type, optional: optional)
+        ret[one_type] = resolvables if !resolvables.nil? && !resolvables.empty?
       end
 
-      deep_copy(ret)
+      ret
     end
 
     # Return whether a unique ID is already in use.
```
jreidinger: in fact this is quite confusing, even if cleared in return part of doc

jreidinger:
in proposal more removed lines then added, so LGTM for me. Just coverage decrease looks a bit suspicious.

PR #520, opened 2016-12-14
Keyboard layout widget

jreidinger:


mchf:
lgtm

library/general/src/lib/ui/widgets.rb
```diff
@@ -0,0 +1,37 @@
...
+      def initialize
+        textdomain "base"
+      end
+
+      def label
+        _("&Test Keyboard Layout")
+      end
+    end
+  end
+end
```
imobachgs: I would prefer having this class defined in a separate file (`keyboard_layout_test`). Something like `lib/ui/widgets/keyboard_layout_test`. What do you think?

imobachgs: It's fine for me.

PR #521, opened 2016-12-15
improve documentation to document that notify is not auto-passed

jreidinger:


library/cwm/src/lib/cwm/widget.rb
```diff
@@ -98,6 +98,8 @@ def self.widget_type=(type)
     #   @return [String] translated label text for the widget
 
     # @!method opt
+    #   Specifies options passed to widget. When {handle} method defined, then
```
imobachgs: What about this? When {handle} method *is* defined, is expected to be immediatelly notified.

Or, at least, replace *immeditally* for *inmediately*.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -109,11 +111,15 @@ def self.widget_type=(type)
     # @overload handle
     #   Process an event generated by this widget. This method is invoked
     #   if {#handle_all_events} is `false`.
+    #   @note defining handle itself does not pass `:notify` to widget. It has
```
imobachgs: Should it be `{handle}`?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -109,11 +111,15 @@ def self.widget_type=(type)
...
+    #     to be done with {opt} method.
     #   @return [nil,Symbol] what to return from the dialog,
     #     or `nil` to continue processing
     # @overload handle(event)
     #   Process an event generated a widget. This method is invoked
     #   if {#handle_all_events} is `true`.
+    #   @note defining handle itself does not pass `:notify` to widget. It has
```
imobachgs: Same as above.

library/cwm/src/lib/cwm/widget.rb
jreidinger: well, it not need to be always defined and e.g. custom widget do it different way. So that is reason why I use often. so what about
```
When {handle} method is defined, it often expected to be immediatelly notified from widget and for that purpose opt with `[:notify]` as return value is needed.
```

library/cwm/src/lib/cwm/widget.rb
jreidinger: yes

library/cwm/src/lib/cwm/widget.rb
imobachgs: Better. I would just add "it *is* often expected"

PR #526, opened 2017-01-05
Fix for the new yast2-core

lslezak:
- Fixes a bug revealed by the new yast2-core - a mocked Pkg implementation did not implement all called methods.

Note: The commits are in an unusual order - the first is the changelog to demonstrate the failure in Travis, second commit is the fix to show the fixed state. I'll squash them at merge to not have this issue in `master`.

### DIFF
--- a/library/packages/test/file_conflict_callbacks_test.rb
+++ b/library/packages/test/file_conflict_callbacks_test.rb
@@ -110,8 +110,7 @@ class DummyPkg
 
   MOCK_METHODS.each do |method|
     # mock empty methods with a single argument
-    send(:define_method, method) do |func|
-    end
+    define_method(method) { |arg| }
   end
 end
 
--- a/library/packages/test/file_conflict_callbacks_test.rb
+++ b/library/packages/test/file_conflict_callbacks_test.rb
@@ -24,6 +24,95 @@ class DummyPkg
   def CallbackFileConflictFinish(func)
     @fc_finish = func
   end
+
+  # run this command in "irb -ryast" to obtain the method names:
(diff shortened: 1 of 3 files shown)


library/packages/test/file_conflict_callbacks_test.rb
```diff
@@ -24,6 +24,95 @@ def CallbackFileConflictReport(func)
...
+    :CallbackStartScanDb,
+    :CallbackStopConvertDb,
+    :CallbackStopRebuildDb,
+    :CallbackTrustedKeyAdded,
+    :CallbackTrustedKeyRemoved
+  ].freeze
+
+  MOCK_METHODS.each do |method|
+    # mock empty methods with a single argument
+    send(:define_method, method) do |func|
```
jreidinger: is send really needed? why simple `define_method` do not work here? so something like `define_method(method){ |f| }`

library/packages/test/file_conflict_callbacks_test.rb
```diff
@@ -24,6 +24,95 @@ def CallbackFileConflictReport(func)
   def CallbackFileConflictFinish(func)
     @fc_finish = func
   end
+
+  # run this command in "irb -ryast" to obtain the method names:
+  # Yast.import "Pkg"; Yast::Pkg.methods.select {|m| m.to_s.start_with?("Callback")}
```
mvidner: @lslezak BTW it is even easier with `Yast::Pkg.methods.grep /^Callback/`

lslezak: Oh, a nice trick...

PR #527, opened 2017-01-05
Do a full type detection when creating cache

mchf:
It avoids ambiguous categorization of a device like in [bsc#956755](https://bugzilla.suse.com/show_bug.cgi?id=956755)
another one: https://bugzilla.suse.com/show_bug.cgi?id=1061306

### DIFF
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -789,7 +789,7 @@ module Yast
     # @param [Hash] a map with netconfig (ifcfg) configuration
     #
     def add_device(device, ifcfg)
-      devtype = GetTypeFromIfcfg(ifcfg) || GetType(device)
+      devtype = GetType(device)
       @Devices[devtype] ||= {}
       @Devices[devtype][device] = ifcfg
     end
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Jan  5 11:58:05 UTC 2017 - mfilka@suse.com
+
+- bnc#956755
+  - Use full type detection when caching network configuration to
+    solve ambiguos configurations (like bridge over wlan).
+- 3.2.12
+
+-------------------------------------------------------------------
 Thu Jan  5 11:35:23 UTC 2017 - mfilka@suse.com
 
 - bnc#1017716
--- a/package/yast2.spec
+++ b/package/yast2.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2
-Version:        3.2.11
+Version:        3.2.12
 Release:        0
 Summary:        YaST2 - Main Package
 License:        GPL-2.0


package/yast2.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Jan  5 11:58:05 UTC 2017 - mfilka@suse.com
+
+- bnc#956755
+  - Use full type detection when caching network configuration to
+    solve ambiguos configurations (like bridge over wlan).
```
mvidner: ambiguo**u**s

But this is describing implementation details and tells the user nothing about the bug this is fixing or its symptoms.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -789,7 +789,7 @@ def adapt_old_config!
     # @param [Hash] a map with netconfig (ifcfg) configuration
     #
     def add_device(device, ifcfg)
-      devtype = GetTypeFromIfcfg(ifcfg) || GetType(device)
+      devtype = GetType(device)
```
mvidner: When I want to understand the change, I look up the names in http://apitalism.herokuapp.com/?q=GetType
and get
- http://www.rubydoc.info/github/yast/yast-yast2/Yast%2FNetworkInterfacesClass%3AGetTypeFromIfcfg
  Detects device type according given ifcfg configuration
- http://www.rubydoc.info/github/yast/yast-yast2/Yast%2FNetworkInterfacesClass%3AGetType 
Detects device type according cached data

  If cached ifcfg for given device is found it is used as parameter for GetTypeFromIfcfgOrName( dev, ifcfg). Otherwise is device handled as unconfigured and result is equal to GetTypeFromIfcfgOrName( dev, nil)

I am still lost. What is the difference in the return value for this particular bug?

mvidner:
The test suite also has something to say, starting [here](https://travis-ci.org/yast/yast-yast2/builds/189156546#L1426).

library/network/src/modules/NetworkInterfaces.rb
mchf: Difference is that ```GetTypeFromIfcfgOrName``` uses ifcfg file or name for device type detection. It is just a subset of whole type detection process which is implemented in ```GetType```. ```GetType``` is the only function which should be present in public API (but it is not bcs of ycp). Particular example of collision is described in the commit message.

package/yast2.changes
mchf: This is an internal module which is used mainly by other yast parts. And the change is implementation detail bcs in perfect code internal sorting of NetworkInterfaces' cache would be really internal. Sadly we more or less expose it's organization, so any change in it is public as well.

mchf:
Regarding testsuite. It is failing only in old testsuite, I'll update "out" files.

lslezak:
@mchf what's the status of this?

mchf:
Will be replaced by an updated PR

PR #528, opened 2017-01-05
Do not cache ifcfgs with emtpy device name part

mchf:
bnc#1017716

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Thu Jan  5 11:35:23 UTC 2017 - mfilka@suse.com
+
+- bnc#1017716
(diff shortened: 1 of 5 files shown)


package/yast2.changes
```diff
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Thu Jan  5 11:35:23 UTC 2017 - mfilka@suse.com
+
+- bnc#1017716
+  - do not cache ifcfg files with empty device name part (ifcfg-).
+    Such file cannot be mappet to any existing device and providing
```
teclator: s/mappet/mapped

mchf: that's why I should not write descriptive changelogs ;-)

mchf:
Thanks for the review

PR #530, opened 2017-01-16
Declare simple_mode for make_proposal (FATE#322328)

mvidner:
https://trello.com/c/XDgl2dj7/816-13-casp-new-all-in-one-dialog
Main PR: https://github.com/yast/yast-installation/pull/487

For context, here is the [rendered documentation for `Installation::ProposalClient#make_proposal`](http://www.rubydoc.info/github/yast/yast-yast2/Installation%2FProposalClient:make_proposal) (before this PR)

### DIFF
--- a/library/general/src/lib/installation/proposal_client.rb
+++ b/library/general/src/lib/installation/proposal_client.rb
@@ -131,14 +131,14 @@ module Installation
     #     Human readable proposal preformatted in HTML. It is possible to use
     #     the {Yast::HTMLClass Yast::HTML} module for such formatting.
     #
-    #   * **`"raw_proposal"`** [Array<String>, nil]
-    #     (only used if `preformatted_proposal` is not present) ---
+    #   * **`"raw_proposal"`** [Array<String>, nil] ---
+    #     (only used if `preformatted_proposal` is not present)
     #     Human readable proposal, not formatted yet.
     #     The caller will format each item as a HTML list item (`<li>`). The
     #     proposal can contain hyperlinks with IDs listed in the list `links`.
     #
-    #   * **`"label_proposal"`** [Array<String>, nil]
-    #     (only used for `cramped_label_mode`) ---
+    #   * **`"label_proposal"`** [Array<String>, nil] ---
+    #     (only used for `cramped_label_mode`)
     #     Human readable proposal.
     #     The caller will format each item as Label.
     #
--- a/library/general/src/lib/installation/proposal_client.rb
+++ b/library/general/src/lib/installation/proposal_client.rb
@@ -116,6 +116,11 @@ module Installation
     #   submodules that translate any strings internally based on internal maps
     #   (e.g., keyboard etc.) need to take more action.
     #
+    # @option attrs [Boolean] "cramped_label_mode"
+    #   Propose in a different format,
+    #   pioneered in the CASP/CaaSP all-in-one dialog: we are short of space,
+    #   and employ labels instead of RichText.
+    #
     # @return [Hash] containing:
     #
     #   * **`"links"`** [Array<String>] ---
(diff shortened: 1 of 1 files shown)


library/general/src/lib/installation/proposal_client.rb
```diff
@@ -132,6 +137,11 @@ def run
     #     The caller will format each item as a HTML list item (`<li>`). The
     #     proposal can contain hyperlinks with IDs listed in the list `links`.
     #
+    #   * **`"label_proposal"`** [Array<String>, nil]
```
jreidinger: why introduce another key in result? I suggest to reuse "raw_proposal" which already have right format and when "cramped_label_mode" is used, then it is hard requirement to return it. What do you think? I think we should add only unnecessary complexity.
btw cramped_label_mode is not easily remembered key. what about "minimal" or "shrinked_mode" ?

mvidner: raw_proposal does *not* have the right format: it can contain rich text hyperlinks (eg. Bootloader: `<font color="red">Do not install...`)

I'll ask around for a  name that  people would consider fitting. BTW, "shrinked" is gram-not-matical: http://www.thefreedictionary.com/shrink

mvidner:
This is for https://github.com/yast/yast-installation/pull/487

library/general/src/lib/installation/proposal_client.rb
jreidinger: well, we can define, that with "cramped_label_mode" it do not contain it. I just do not like to return hash with possible bunch of keys. If we define that if this key is passed it have to return "raw_proposal" without any richtext tags, it will reduce list of possible return states.

mvidner: OK, I see your point. The result already has 10 keys (not fully clear which are optional). But IMHO reusing `raw_proposal` would complicate both the implementation and the documentation:

> `raw_proposal`: Human readable proposal, not formatted yet. For `cramped_label_mode` they will be put in Labels. Otherwise the caller will format each item as a HTML list item (`<li>`). The proposal can contain hyperlinks with IDs listed in the list links.

jreidinger: I would remove details from documentation and use it like
```
raw_proposal: Human readable proposal, not in final format. Each item is used as separate statement about module. For cramped_label_mode does not use any richtext tags. The proposal outside of cramped_label_mode can contain hyperlinks with IDs listed in the list links.
```

And to be honest I am not sure if final sentence do not change in future as it is quite useful to have one click modification.

about implementation I worry we need to have separate implementation even now due to links stuff and others.

Or just idea that comes to my mind is to not have any cramped_label_mode and simply always return label_proposal in response. Which can simplify a bit stuff, but also produce some duplicite work. What do you think?

lslezak: I'd would be probably against reusing the `raw_proposal` key. Defining yet another key does not look nice but's the only way how to know that the client really handles the `cramped_label_mode` correctly.

Think about some products handled outside the YaST team (e.g. OES, SAP...), maybe they would like to use this all-in-one dialog too. But they would like to use different proposal clients, which might not have support for this new mode...

They might try and see that it works (in almost all cases), but they miss a special error/corner case when the proposal suddenly returns rich text instead of plain text label and then it breaks...

Reusing the same key with a different meaning/semantic does not look right to me... When using a new key we can easily check the result and if the key is missing then display an error saying that the module does not support the simplified proposal mode .

library/general/src/lib/installation/proposal_client.rb
```diff
@@ -116,6 +116,11 @@ def run
     #   submodules that translate any strings internally based on internal maps
     #   (e.g., keyboard etc.) need to take more action.
     #
+    # @option attrs [Boolean] "cramped_label_mode"
```
lslezak: What about something like `simple_mode`? `cramped_label_mode` sounds a bit complicated to me...

mvidner:
I'll comment here because we're discussing the whole thing at once anyway:

**Output**: Lada has mentioned more reasons not to reuse `raw_proposal`, so let's use `label_proposal` for the return value.

**Input**: it is an interesting idea to not distinguish the input in any way, and simply include the new output always. I have a feeling that it may work out or not. The calling environment will simply be different and we will soon find out whether we need some special handling regarding interactions between modules, refreshes, error handling etc. So I think it is better to include the input flag. As for the name, `simple_mode` sounds good.

mvidner:
Jenkins failing because ci.o.o cannot see the IBS servers
Travis failing because... casp is halfway between master and sp2? Anyway, this PR is comments only.

PR #531, opened 2017-01-17
Disable buttons cwm

jreidinger:


mvidner:
Thanks! Does this disable (gray out) the buttons or remove (hide) them?

Now I have doubts: our libraries have options for *disabling* the buttons but https://trello.com/c/XDgl2dj7/816-13-casp-new-all-in-one-dialog calls for *removing* them. Again the question whether the requester wants specifically this, or "something like this, the way you usually do it".

I asked for `""` to mean removing, because then we remove any doubt regarding the label of a disabled button. No string, no button.

jreidinger:
@mvidner does wizard support it? CWM is build on top of Wizard

jreidinger:
doin quick overview I do not see it https://github.com/libyui/libyui/blob/e2c6fd3ce25abeeb367091ffaeac93ae18cfc5e3/src/YWizard.h

mvidner:
OK, let's go with *disabled*, not *removed* buttons.

I don't like `""` because it is not obvious to the reader at the call site what it means. And using the old API from ShowAndRun, `disable_buttons: ["back_button", "abort_button"]` feels clumsy.

What about `back_button: :disabled`?

kobliha:
Requirements say:

- [Help] and [Install] buttons, nothing else now ... see below
- No [Abort] button

That sounds like removing instead of disabling. UI/UX IMO did not consider showing disabled buttons, but another piece of info is: `We have stripped down everything to a bare minimum and we will show only that, nothing else` - that also sounds like removing. You can still ask PM/PjM.

jreidinger:
lets close it for now, as not needed. Lets consider again it when we have real world use case.

PR #532, opened 2017-01-18
All in one overview

jreidinger:


mvidner:
Please fix rubocop style in the example

mvidner:
.changes +=
> Added a CWM::ReplacePoint widget

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -821,4 +821,72 @@ def replace_point
...
+    end
+
+    # Replaces content with different widget. All its events are properly
+    # handled.
+    # @param widget [CWM::AbstractWidget] widget to display and process events
+    def replace(widget)
+      log.info "replacing with new widget #{widget.inspect}"
+      Yast::UI.ReplaceWidget(@id, widget_content(widget))
+      @widget = widget
+      @widget.init if @widget.respond_to?(:init)
```
mvidner: Doesn't this mean that `CWM::AbstractWidget` should rather always define an empty `init`?

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -821,4 +821,72 @@ def replace_point
...
+    end
+
+    def help
+      @widget.respond_to?(:help) ? @widget.help : ""
```
mvidner: Likewise, `CWM::AbstractWidget` should rather always define a `help` returning `""`?

mvidner: And so on for all the other methods

mvidner: Ah, now I see the reason:
https://github.com/yast/yast-yast2/blob/450edff9aeaa1dbd145e90d6175cad285a55cdd0/library/cwm/src/lib/cwm/widget.rb#L90

jreidinger: yes, there is CWM defaults and I do not want to duplicite it. and in fact, for common widgets, when there is no help method, it will create for cwm "nohelp" key instead of common "help". For this one, I do not want to overcomplicated it, even if with ruby I can dynamically define and undefine methods.

library/cwm/src/lib/cwm/widget.rb
jreidinger: well, it is caused by CWM itself, which allow no key, but I do not have reasonable method to recognize that method init is empty, so it is design decision in CWM widgets itself.

PR #535, opened 2017-01-20
Fix replace widget

jreidinger:
See also https://lists.opensuse.org/yast-devel/2017-01/msg00035.html

### DIFF
--- a/library/cwm/src/lib/cwm/widget.rb
+++ b/library/cwm/src/lib/cwm/widget.rb
@@ -836,7 +836,7 @@ module CWM
     end
 
     def contents
-      ReplacePoint(Id(widget_id.to_sym), widget_content(@widget))
+      ReplacePoint(Id(widget_id), widget_content(@widget))
     end
 
     def init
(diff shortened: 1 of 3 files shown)


mvidner:
I don't understand. What is the symptom of the bug? A piece of y2log?

Yes, a CWM widget id probably needs to be a String, but I have tested a UI ReplacePoint with an Id("foo") id and it has worked (replaced all occurrences of `:rp` with `"rp"` in https://github.com/yast/yast-ycp-ui-bindings/blob/466ee34720198f1cecd2c13884452999c84815ca/examples/ReplacePoint1.rb)

jreidinger:
@mvidner see mail to yast-devel - or see UI call - https://doc.opensuse.org/projects/YaST/openSUSE11.3/tdg/ReplaceWidget.html

mvidner:
OK, so the problem is that in all places where YUI expects a widget ID, you can pass a symbol, **or** a term whose value **can** be a non-symbol. And we always pass a term whose name is `id` (written `Id` in Ruby). But [the documentation](https://doc.opensuse.org/projects/YaST/openSUSE11.3/tdg/ReplaceWidget.html) only mentions the first of the overloaded signatures, the one with `symbol`.

PR #539, opened 2017-02-22
adapt to new y2base argument passing

jreidinger:
will fail, as it use y2base from yast2-core and not one from ruby-bindings
fail will be solved by dependencies.


mvidner:
da74e25 "WIP of rewritting DnsServerAPI to ruby" does not belong to this PR

jreidinger:
damn, you are right, it is part of pr to drop perl. I will remove it.

scripts/yast2
```diff
@@ -285,19 +285,14 @@ else
...
 
 if [ -n "$CMDLINE_HELP" ]; then
     set -- $@ help;
 fi
 if [ $# -gt 0 ]; then
     a=("$@")
-    b=("${a[@]/#/(}")
-    c=("${b[@]/%/)}")
-    set -- -S "${c[@]}"
+    set -- --arg "${a[@]}"
```
mvidner: This is wrong. `-S` starts multiple arguments but `--arg` has a single argument.

mvidner: BTW you can debug this with
```sh
printf "%q\n" "${@}"
printf "%q\n" "${a[@]}"
```

scripts/yast2
```diff
@@ -290,10 +290,11 @@ fi
...
 fi
-if [ $# -gt 0 ]; then
-    a=("$@")
-    set -- --arg "${a[@]}"
-fi
+ARGS=""
+for i in $@; do
+  ARGS="$ARGS --arg $i"
+done
+set -- $ARGS
```
mvidner: No, this breaks arguments that contain spaces. Please test that.

scripts/yast2
```diff
@@ -290,10 +290,11 @@ fi
 if [ -n "$CMDLINE_HELP" ]; then
     set -- $@ help;
 fi
-if [ $# -gt 0 ]; then
-    a=("$@")
-    set -- --arg "${a[@]}"
-fi
+ARGS=""
+for i in "$@"; do
+  ARGS="$ARGS --arg '$i'"
```
mvidner: This will still fail if the value contains a single quote

jreidinger: so any suggestion to use this ****** shell **** programming language?

jreidinger: and also what is probability that some parameter will contain it?

mvidner: We got a bug once, not for a quote but for a backslash: https://bugzilla.suse.com/show_bug.cgi?id=382883

mvidner: Bash is awful, but it does have a solution for this: `printf %q` is the proper way to quote for an `eval`:

```sh
ARGS=$(printf '%s -arg %q' "$ARGS" "$i")
```

scripts/yast2
```diff
@@ -285,20 +284,16 @@ else
...
 fi
-if [ $# -gt 0 ]; then
-    a=("$@")
-    b=("${a[@]/#/(}")
-    c=("${b[@]/%/)}")
-    set -- -S "${c[@]}"
-fi
+ARGS=""
+for i in "$@"; do
+  ARGS="$ARGS --arg `printf %q "$i"`"
```
mvidner: Well, my earlier suggestion, and this variant, are both wrong.
We need tests for this.

jreidinger: I testing it manually trying various stuff on CLI. Why is this wrong? And for tests you mean tests in bash? I hope not

mvidner:
When I take today's commits and apply them to master, the test should pass. But it still fails, and it is hard to see why:
```
[   61s] Failures:
[   61s] 
[   61s]   1) yast2 script pass properly all arguments
[   61s]      Failure/Error: Cheetah.run(TEST_DIR + "/yast2", "args_test_client", 'abc"\'\\|;&<>! ', "second", env: { "TESTING_YAST2" => "1" })
[   61s] 
[   61s]      Cheetah::ExecutionFailed:
[   61s]        Execution of "/home/abuild/rpmbuild/BUILD/yast2-3.2.14/scripts/yast2 args_test_client abc\"\'\\\|\;\&\<\>\!\  second" failed with status 1 (no error output).
[   61s]      # ./yast2_test.rb:13:in `block (2 levels) in <top (required)>'
```

mvidner:
BTW when Cheetah says " (no error output)" it is because the program does
```sh
echo fail
exit 1
```
instead of
```sh
echo fail >&2
exit 1
```
and it cannot be helped with `stdout: :capture` or `stderr: :capture`.

jreidinger:
@mvidner true, cheetah show in exception message only error ouput, but have access to both in exception - https://github.com/openSUSE/cheetah/blob/master/lib/cheetah.rb#L53

PR #540, opened 2017-03-02
[WIP] Drop perl

jreidinger:


lslezak:
@jreidinger what's the status of this? If it needs a lot of work then create a Trello card so we can plan it for the next sprint.

jreidinger:
@lslezak  basically it is unfinished hackweek project where I do not finish it for remaining perl module, which is the biggest one.

jreidinger:
lets close for now, but please keep branch

PR #542, opened 2017-03-08
Add system roles through addons

imobachgs:
- https://trello.com/c/mb2P0JdQ
- https://fate.suse.com/320772

This PR adds support to allow add-ons to add new roles.

* It promotes `system_roles` within the `ProductControl` module so you don't need to access to its definition through the `ProductControl.productControl` method.
* System roles are integrated as workflows, proposals, etc.

Here's an example of adding system roles in the *update* section of an add-on.

```xml
<update>
  <system_roles>
    <insert_system_roles config:type="list">
      <insert_system_roles>
        <!-- We could specify additional information here, like: -->
        <!-- <before>default_role</before> -->
        <!-- <position>-1</position> -->
        <!-- <after>kvm_host_role</after> -->

        <system_roles config:type="list">
          <system_role>
            <id>additional_role</id>
          </system_role>
        </system_roles>
      </insert_system_roles>
    </insert_system_roles>
  </system_roles>
</update>
```

For the records, this PR is related to https://github.com/yast/yast-installation/pull/534.


library/control/src/modules/ProductControl.rb
```diff
@@ -1591,11 +1601,29 @@ def ResetAdditionalWorkflowParams
       nil
     end
 
+    # Add new system roles
+    #
+    # For the time being, roles are inserted at the beginning of the
```
mvidner: `Array#concat` adds at the end, not at the beginning.

library/control/src/modules/WorkflowManager.rb
```diff
@@ -83,6 +83,7 @@ def main
       @wkf_initial_proposals = []
       @wkf_initial_inst_finish = []
       @wkf_initial_clone_modules = []
+      @wkf_initial_roles = []
```
mvidner: wkf_initial_system_roles

library/control/src/modules/ProductControl.rb
```diff
@@ -1591,11 +1601,29 @@ def ResetAdditionalWorkflowParams
...
+    # roles list.
+    #
+    # @example Adding a simple role
+    #   ProductControl.system_roles #=> []
+    #   ProductControl.add_system_roles([{"id" => "new_role"}]) #=> [{"id" => "new_role"}]
+    #   ProductControl.system_roles #=> [{"id" => "new_role"}]
+    #
+    # @param [Array<Hash>] new_roles Roles to add
+    # @return [Array<Hash>] All existing roles after adding the new roles
```
mvidner: But the caller does not use that, and the `publish` declaration says `void` too.

library/control/src/modules/ProductControl.rb
imobachgs: Yes, I updated the code and forgot to update the comment

imobachgs:
Thanks!

PR #543, opened 2017-03-09
Backported code from the current master

mchf:
https://bugzilla.suse.com/show_bug.cgi?id=1028081

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ before_install:
     # disable rvm, use system Ruby
     - rvm reset
     - wget https://raw.githubusercontent.com/yast/yast-devtools/master/travis-tools/travis_setup.sh
-    - sh ./travis_setup.sh -p "rake yast2-core yast2-devtools yast2-testsuite yast2-ruby-bindings yast2 yast2-pkg-bindings" -g "rspec:3.3.0 yast-rake gettext simplecov coveralls rubocop:0.29.1"
+    - sh ./travis_setup.sh -p "rake yast2-core yast2-devtools yast2-testsuite yast2-ruby-bindings yast2 yast2-pkg-bindings" -g "rspec:3.3.0 yast-rake gettext coveralls rubocop:0.29.1"
 script:
     - rake check:pot
     - rubocop
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -701,11 +701,10 @@ module Yast
       @Devices = {}
 
       # preparation
-      allfiles = SCR.Dir(path(".network.section"))
-      allfiles = [] if allfiles.nil?
-      devices = Builtins.filter(allfiles) do |file|
-        !Builtins.regexpmatch(file, "[~]")
-      end
+      devices = SCR.Dir(path(".network.section")) || []
+      devices.select! { |file| file !~ /[~]/ }
+      devices.delete_if(&:empty?)
+
       Builtins.y2debug("devices=%1", devices)
 
       # Read devices
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -29,6 +29,8 @@ module Yast
   # Categorizes the configurations according to type.
   # Presents them one ifcfg at a time through the {#Current} hash.
   class NetworkInterfacesClass < Module
+    include Logger
+
     Yast.import "String"
 
     # A single character used to separate alias id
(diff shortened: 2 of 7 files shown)


library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -681,6 +683,20 @@ def ConcealSecrets(devs)
       deep_copy(out)
     end
 
+    # Get current sysconfig configured interfaces
+    #
+    # @param devregex [String] regex to filter by
+    # @return [Array] of ifcfg names
+    def get_devices(devregex = "[~]")
```
jreidinger: why parameter is not regexp and instead only string that is later interpreted as regexp?

jreidinger: also why regexp is not just `~` ?

library/network/test/network_interfaces_helpers_test.rb
```diff
@@ -71,5 +71,47 @@ module Yast
...
+        subject.main
+        allow(subject).to receive(:Read).and_return(true)
+        allow(Yast::SCR).to receive(:Dir).with(Yast::Path.new(".network.section")).and_return(devices)
+      end
+
+      it "returns an array of configured interfaces filtered by regexp" do
+        expect(subject.get_devices("1")).not_to include "em1", "eth1", "br1"
+      end
+
+      it "filters with <[~]> by default" do
```
jreidinger: maybe i woulld name it `filters devices with "~" by default`

library/network/test/network_interfaces_helpers_test.rb
```diff
@@ -71,5 +71,47 @@ module Yast
...
+        expect(subject.get_devices).not_to include "tr~"
+      end
+
+      it "returns an empty array with <.> argument" do
+        expect(subject.get_devices(".")).to eql []
+      end
+
+      it "returns all devices filtering with <''>" do
```
jreidinger: `with empty string`

library/network/test/network_interfaces_helpers_test.rb
```diff
@@ -71,5 +71,47 @@ module Yast
...
+        expect(subject.get_devices("")).to eql devices
+      end
+
+      it "does not crash with exception" do
```
jreidinger: quite strong statement and you are never sure that it do not return expection. E.g. low memory or signal can get almost in every line of code.

library/network/test/network_interfaces_helpers_test.rb
```diff
@@ -71,5 +71,47 @@ module Yast
...
+        expect { subject.get_devices }.not_to raise_error
+      end
+
+      it "doesn't carry empty strings" do
```
jreidinger: `filters out empty strings`

library/network/src/modules/NetworkInterfaces.rb
mchf: I don't know, ask @teclator ;-)

mchf:
@jreidinger Are you fine with creating a PR with modifications based on your comments against master? This should be pure backport of code which already (more less) is in master.

jreidinger:
@mchf yes, I am fine.

library/network/src/modules/NetworkInterfaces.rb
teclator: You can improve it 👍 . It was one of my first cleanups and just use same match that was used with `Builtins.regexpmatch`

https://github.com/yast/yast-yast2/pull/433/files#diff-36f13a6d2bafc7ad7bcfa6f8a452e81dL704

PR: https://github.com/yast/yast-yast2/pull/433

teclator:
@mchf @jreidinger, I'm fine with just merge this PR with the backported code but the improvements or modifications maybe would be great to do directly in SP1 and then merge into SP2 and master

library/network/src/modules/NetworkInterfaces.rb
mchf: @teclator I know ;-) I'll do so for master not maintenance branch.

.travis.yml
```diff
@@ -5,7 +5,7 @@ before_install:
     # disable rvm, use system Ruby
     - rvm reset
     - wget https://raw.githubusercontent.com/yast/yast-devtools/master/travis-tools/travis_setup.sh
-    - sh ./travis_setup.sh -p "rake yast2-core yast2-devtools yast2-testsuite yast2-ruby-bindings yast2 yast2-pkg-bindings" -g "rspec:3.3.0 yast-rake gettext simplecov coveralls rubocop:0.29.1"
+    - sh ./travis_setup.sh -p "rake yast2-core yast2-devtools yast2-testsuite yast2-ruby-bindings yast2 yast2-pkg-bindings" -g "rspec:3.3.0 yast-rake gettext simplecov:0.10.0 coveralls rubocop:0.29.1"
```
jreidinger: better fix is to completelly remove simplecov and it pick right version

mchf: ok

PR #546, opened 2017-03-10
Remove /sbin/insserv

mchf:


jreidinger:
but please fix also simplecov dependency so travis do not fail

PR #547, opened 2017-03-10
Added PackageDownloader and PackageExtractor

lslezak:
This is refactoring of the [`UpdateRepository` class](https://github.com/yast/yast-installation/blob/04de58011068c5339c2c76b9be5954a8ea0b808c/src/lib/installation/update_repository.rb#L44) to share the package downloading and extracting code from the self-update feature with the installer extension feature ([FATE#320772](https://fate.suse.com/320772)).

- The `PackageDownloader` implements the [`fetch_package`](https://github.com/yast/yast-installation/blob/04de58011068c5339c2c76b9be5954a8ea0b808c/src/lib/installation/update_repository.rb#L286) functionality.
- The `PackageExtractor` implements the [`extract`](https://github.com/yast/yast-installation/blob/04de58011068c5339c2c76b9be5954a8ea0b808c/src/lib/installation/update_repository.rb#L311) functionality.
- Added a real small testing RPM for verifying the extractor functinality.

After merging this PR the `UpdateRepository` class should be refactored to use this new shared code.


library/packages/src/lib/packages/package_downloader.rb
```diff
@@ -0,0 +1,81 @@
...
+      @repo_id = repo_id
+      @package_name = package_name
+    end
+
+    # Download the package locally to the given path.
+    #
+    # It is responsibility of the caller to remove the downloaded package
+    # when it is not needed anymore.
+    #
+    # @param [#to_s] path path where the downloaded package will be stored
```
imobachgs: np: I usually prefer the alternative syntax to document params:

```ruby
# @param path [#to_s] path where the downloaded package will be stored.
```

library/packages/test/Makefile.am
```diff
@@ -2,14 +2,16 @@ TESTS = \
   commit_result_test.rb \
   dummy_callbacks_test.rb \
   file_conflict_callbacks_test.rb \
+  lib/product_test.rb \
+  lib/repository_test.rb \
   package_callbacks_test.rb \
+  package_downloader_test.rb \
```
imobachgs: I would move those tests to the `test/lib` directory.

library/packages/src/lib/packages/package_downloader.rb
```diff
@@ -0,0 +1,81 @@
...
+  #     downloader = PackageDownloader.new(3, "yast2")
+  #     tmp = Tempfile.new("downloaded-package-")
+  #     downloader.download(tmp.path)
+  #     # do something with the package...
+  #   ensure
+  #     tmp.close
+  #     tmp.unlink
+  #   end
+  #
+  class PackageDownloader
```
imobachgs: Why not use the `Packages` namespace? Classes defined in that directory are under the `Packages` too. Searching for another example, `library/general/src/lib/installation/` uses the `Installation` module. I would prefer being consistent.

lslezak: Oh, good point, I overlooked this... Fixed

PR #548, opened 2017-03-15
Add FileChanges.new_created_files (part of fix for bsc#1027582)

ancorgs:


library/system/src/modules/FileChanges.rb
```diff
@@ -213,16 +213,24 @@ def CheckFiles(files)
       ret == :ok
     end
 
+    # Files that are really new
+    #
+    # @param files [Array<String>] candidate files that may be new
+    # @return [Array<String>]
+    def new_created_files(files)
```
jreidinger: name is a bit misleading for me, maybe `newly_created_files`?

jreidinger: current name looks like it create new object of type CreatedFiles.

ancorgs: It's for consistency with `CheckNewCreatedFiles` (the new method is extracting part of its logic).

jreidinger: hmm, it is different naming convention and also in combination of CheckNew sounds for me a bit different. I actually notice it first time in second pull request, where it hit my eyes, but I let it up to you

PR #553, opened 2017-03-16
Sle 12 sp2 casp bnc 1012047

schubi2:


ancorgs:
Correct me if I'm wrong, but according to the relatively recent thread in yast-devel. The version number here should be 3.1.214.1. Isn't it?

PR #556, opened 2017-03-22
Cfa switched scr

jreidinger:


library/system/src/lib/yast2/target_file.rb
```diff
@@ -42,8 +40,7 @@ def self.write(path, content)
     end
 
     def self.final_path(path)
-      root = "/"
-      root = Yast::Installation.destdir if Yast::WFM.scr_chrooted?
+      root = Yast::WFM.scr_root
```
imobachgs: You could do all in just one line, but I'm fine if it's more readable for you.

jreidinger: well, if method is too long I will do it, but in this case there are just two statements, so in this case I think it is more readable this way.

PR #557, opened 2017-03-22
Extra spacing support for CWM::RadioButtons

ancorgs:
See https://github.com/yast/yast-installation/pull

### DIFF
--- a/library/cwm/src/lib/cwm/widget.rb
+++ b/library/cwm/src/lib/cwm/widget.rb
@@ -602,12 +602,30 @@ module CWM
     include ItemsSelection
     abstract_method :label
 
+    # @!method vspacing
+    #   @return [Fixnum] space between the options
+
     def value
       Yast::UI.QueryWidget(Id(widget_id), :CurrentButton)
     end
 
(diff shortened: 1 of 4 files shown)


library/cwm/src/lib/cwm/widget.rb
```diff
@@ -602,13 +602,31 @@ class RadioButtons < AbstractWidget
     include ItemsSelection
     abstract_method :label
 
+    # @!method vspacing
+    #   @return [Fixnum] space between the options
```
jreidinger: in fact radio buttons support both vspacing and hspacing. So I think it make sense to add both.

see https://github.com/yast/yast-yast2/blob/master/library/cwm/src/modules/CWM.rb#L594

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -602,13 +602,37 @@ class RadioButtons < AbstractWidget
...
+
```
jreidinger: please mention here also method hspacing, as there is usually no other way to know which methods are useful.

PR #559, opened 2017-03-24
Download addons installation.xml from RPM package

lslezak:
- Enhanced and improved the PoC by @jsrain 
- Changed the RPM `Provides` from `installerupdate()` to `installerextension()` ("installerupdate" sounds rather self-update related and might be confusing).
- Added tests

### DIFF
--- a/library/control/src/modules/WorkflowManager.rb
+++ b/library/control/src/modules/WorkflowManager.rb
@@ -451,9 +451,7 @@ module Yast
             true
           )
 
-          if use_filename.nil?
-            use_filename = GetControlFileFromPackage(src_id)
-          end
+          use_filename = GetControlFileFromPackage(src_id) if use_filename.nil?
 
           # File exists?
           return use_filename.nil? ? nil : StoreWorkflowFile(use_filename, disk_filename)
@@ -1484,7 +1482,7 @@ module Yast
     def find_control_package(release_package)
       return nil unless release_package && release_package["deps"]
 
-      release_package["deps"].each do | dep |
+      release_package["deps"].each do |dep|
         provide = dep["provides"]
         next unless provide && provide.match(/\Ainstallerextension\((.+)\)\z/)
 
@@ -1512,16 +1510,14 @@ module Yast
     end
 
     def fetch_package(repo_id, package, dir)
-      begin
-        downloader = Packages::PackageDownloader.new(repo_id, package)
-        tmp = Tempfile.new("downloaded-package-")
-        downloader.download(tmp.path)
-        extract(tmp, dir)
-      ensure
-        if tmp
-          tmp.close
-          tmp.unlink
-        end
+      downloader = Packages::PackageDownloader.new(repo_id, package)
+      tmp = Tempfile.new("downloaded-package-")
+      downloader.download(tmp.path)
+      extract(tmp, dir)
+    ensure
+      if tmp
+        tmp.close
+        tmp.unlink
       end
     end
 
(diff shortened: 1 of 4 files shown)


library/control/src/modules/WorkflowManager.rb
```diff
@@ -393,6 +395,62 @@ def StoreWorkflowFile(file_from, file_to)
...
     end
 
+    # Download and extract the control file (installation.xml) from the add-on
+    # repository.
+    #
+    # @param src_id [Fixnum] repository ID
+    # @return [String, nil] path to downloaded installation.xml file or nil
+    #   or nil when no workflow is defined or the workflow package is missing
+    # @raise [Packages::PackageDownloader::FetchError] if package download failed
+    # @raise [Packages::PackageExtractor::ExtractionFailed] if package extraction failed
```
jreidinger: this is wrong, method in fact catch both and do not raise it.

library/control/src/modules/WorkflowManager.rb
```diff
@@ -411,14 +469,18 @@ def GetCachedWorkflowFilename(type, src_id, _name)
...
           use_filename = Pkg.SourceProvideDigestedFile(
             src_id,
             1,
             "/installation.xml",
             true
           )
 
+          # The most generic way it to use the package referenced by the "installerextension()"
+          # provides, this works with all repository types, including the RPM-MD repositories.
+          use_filename = addon_control_file(src_id) if use_filename.nil?
```
jreidinger: can be written as `use_filename ||= addon_control_file(src_id)`

library/control/src/modules/WorkflowManager.rb
```diff
@@ -1427,6 +1489,93 @@ def DumpCurrentSettings
...
+    # RPM "Provides" dependency.
+    # @return [String,nil] a package name or nil if not found
+    def find_control_package(release_package)
+      return nil unless release_package && release_package["deps"]
+
+      release_package["deps"].each do |dep|
+        provide = dep["provides"]
+        next unless provide && provide.match(/\Ainstallerextension\((.+)\)\z/)
+
+        control_file_package = Regexp.last_match[1].strip
```
jreidinger: well, this is not so readable, to check where it is in fact defined.

What about 
```
next unless provide
control_file_package = provide[/\Ainstallerextension\((.+)\)\z/, 1]
next unless control_file_package
return control_file_package.strip
```

library/control/src/modules/WorkflowManager.rb
```diff
@@ -1427,6 +1489,93 @@ def DumpCurrentSettings
...
+    end
+
+    # Download and extract a package from a repository.
+    # @param repo_id [Fixnum] repository ID
+    # @param package [String] name of the package
+    # @raise [Packages::PackageDownloader::FetchError] if package download failed
+    # @raise [Packages::PackageExtractor::ExtractionFailed] if package extraction failed
+    def fetch_package(repo_id, package, dir)
+      downloader = Packages::PackageDownloader.new(repo_id, package)
+      tmp = Tempfile.new("downloaded-package-")
```
jreidinger: why not use block form of Tempfile.open ?

library/control/test/workflow_manager_test.rb
```diff
@@ -266,4 +266,154 @@
...
+      end
+    end
+
+    it "does not create the directory if it already exists" do
+      dir = subject.addon_control_dir(src_id)
+      expect(File.directory?(subject.addon_control_dir(src_id))).to be true
+      expect(FileUtils).to_not receive(:mkdir_p).with(dir)
+      subject.addon_control_dir(src_id)
+    end
+  end
```
jreidinger: :+1: for test coverage

PR #560, opened 2017-03-28
SlideShow: escape license in plain text (bsc#1028721).

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1028721

Just applied the same fix already in [ProductLicense](https://github.com/yast/yast-packager/blob/dff498280c3969db2a386e9face338af36e24a4b/src/modules/ProductLicense.rb#L179)

### Before the fix:

screenshot_sles12sp3_2017-03-28_07 28 12

### After the fix
screenshot_sled12sp3_2017-03-28_07 18 54

screenshot_sles12sp3_2017-03-28_08 10 46

### DIFF
--- a/library/packages/src/modules/SlideShow.rb
+++ b/library/packages/src/modules/SlideShow.rb
@@ -568,6 +568,7 @@ module Yast
     # @return	A term describing the widgets
     #
     def RelNotesPageWidgets(id)
+      # Release notes in plain text need to be escaped to be shown properly (bsc#1028721)
       rel_notes =
         if @_rn_tabs[id] =~ /<\/.*>/
           @_rn_tabs[id]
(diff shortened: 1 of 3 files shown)


library/packages/src/modules/SlideShow.rb
```diff
@@ -568,7 +568,14 @@ def DetailsPageWidgets
     # @return	A term describing the widgets
     #
     def RelNotesPageWidgets(id)
-      widgets = AddProgressWidgets(:relNotesPage, RichText(@_rn_tabs[id]))
+      rel_notes =
+        if @_rn_tabs[id] =~ /<\/.*>/
```
jreidinger: please add comment why we are doing this.

teclator: Added

PR #561, opened 2017-03-28
SlideShow: escape license in plain text. (backport from SP3)

teclator:
It is a backport from #560 where you can see screenshots and more details.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Mar 28 10:19:44 WEST 2017 - knut.anderssen@suse.com
+
+- backported
+  - SlideShow: Escape plain text release notes being shown properly
+    in RichText (bsc#1028721).
+- 3.1.215
+
+-------------------------------------------------------------------
 Thu Mar 16 13:48:06 CET 2017 - schubi@suse.de
 
 - UnitFileState will be used for evaluating enable state of
--- a/library/packages/src/modules/SlideShow.rb
+++ b/library/packages/src/modules/SlideShow.rb
@@ -568,7 +568,15 @@ module Yast
     # @return	A term describing the widgets
     #
     def RelNotesPageWidgets(id)
-      widgets = AddProgressWidgets(:relNotesPage, RichText(@_rn_tabs[id]))
+      # Release notes in plain text need to be escaped to be shown properly (bsc#1028721)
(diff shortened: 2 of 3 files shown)


package/yast2.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Mar 28 10:19:44 WEST 2017 - knut.anderssen@suse.com
+
+- backported
+  - SlideShow: Escape plain text release notes being shown properly
+    in RichText (bsc#1028721).
+- 3.1.215
```
jreidinger: is version correct? it looks like it collide with CASP

teclator: I will merge now against CaaSP, in CaaSP current version is 3.1.214.1 so after the merge it could be 3.1.215.1 if not backported 3.1.214.1 is that correct?

teclator: And it is a good opportunity to sync it with CASP, could we backport all the changes? I think that with the widgets will not be a problem, what about removing inserv #546 

@jreidinger @mchf could you check?, by now I can just backport this fix but was there some strong reason to only apply those changes to CASP branch?.

jreidinger: @teclator in general CASP should contain all SP2 changes

teclator: @jreidinger sorry it is not about merging changes from SP2 into CASP but from CASP back to SP2 ... I mean all the changes from `3.1.210.1` to `3.1.210.6`

teclator:
I will merge it by now but could be taking in account in next changes.

package/yast2.changes
jreidinger: @teclator no, CASP changes should not go now to SP2. otherwise it will part of next MU, which is what we really do not want.

jreidinger: @teclator I think it is generic agreement, as each change is potential risk of failure, so we add to SP2 only what is requested.

ancorgs: @jreidinger As I understood the agreement with QA, we should merge code from CaaSP into SP2 as often as possible. It means some MU will contain code that is not there to fix a real SP2 bug, but only to keep SP2 and CaaSP in sync.

That's expected and accepted by QA, as I understood, since it's less harmful that having a **huge** MU for SP2 released at the CaaSP release day and containing everything.

@kobliha, @jsrain Please correct me if I'm wrong.

PR #564, opened 2017-03-30
Fixed downloading installer extension package (FATE#320772)

lslezak:
This is an fix up for #559.

I tested the latest SP3 build with my example extension, it revealed some bugs not found by unit tests:

- A conflict between `Yast::Packages` (from `modules/Packages.rb`) and the `::Packages` name space. Fixed by full qualification.
- The `extract` method argument was a `Tempfile` object instead of a `String` - the call was mocked in the test so it was not found. Fixed and improved the test to check also the passed argument type to avoid regressions.

### DIFF
--- a/library/control/test/workflow_manager_test.rb
+++ b/library/control/test/workflow_manager_test.rb
@@ -362,9 +362,9 @@ describe Yast::WorkflowManager do
     end
 
     it "downloads and extracts the extension package" do
-      expect_any_instance_of(Packages::PackageDownloader).to receive(:download)
+      expect_any_instance_of(Packages::PackageDownloader).to receive(:download).with(instance_of(String))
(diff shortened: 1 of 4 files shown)


library/control/test/workflow_manager_test.rb
```diff
@@ -363,6 +363,7 @@
 
     it "downloads and extracts the extension package" do
       expect_any_instance_of(Packages::PackageDownloader).to receive(:download)
+      expect(Packages::PackageExtractor).to receive(:new).with(instance_of(String)).and_call_original
       expect_any_instance_of(Packages::PackageExtractor).to receive(:extract)
```
jreidinger: please add here check if arguments are correct, so your fix is covered.

PR #565, opened 2017-04-03
fix yardoc tags

jreidinger:
it does not fix all issues, but it makes sense alone and I do not want to create to big pull requests.

### DIFF
--- a/library/packages/src/modules/SignatureCheckDialogs.rb
+++ b/library/packages/src/modules/SignatureCheckDialogs.rb
@@ -70,7 +70,7 @@ module Yast
     # Functions sets whether user want's to show the dialog again
     #
     # @param [String] popup_type dialog type
-    # @param boolean show again
+    # @param [true, false] show_it show again
     # @param [String] popup_url
     def SetShowThisPopup(popup_type, show_it, popup_url)
       if popup_type.nil? || show_it.nil?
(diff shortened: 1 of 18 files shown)


library/packages/src/modules/SignatureCheckDialogs.rb
```diff
@@ -878,10 +876,10 @@ def HandleDoNotShowDialogAgain(default_return, dont_show_dialog_ident, dont_show
 
     # Function adds delimiter between after_chars characters in the string
     #
-    # @param string to be splitted
-    # @param [String] delimiter
-    # @param integer after characters
-    # @return [String] with delimiters
+    # @param whattosplit [String] text to be splitted
```
imobachgs: np: Here we're using the form "param_name [class] description". However, we're using "[class] param_name description" in [other](https://github.com/yast/yast-yast2/pull/565/files#diff-02086261e6ecff029ebc83758fdf1b6fR474) [places](https://github.com/yast/yast-yast2/pull/565/files#diff-87d364fd10564ca337c52efbb988328cR160). I prefer this one (param name first) but I'm ok with both of them.

jreidinger: I try to be consistent with rest of file, somewhere ( at start I try to modify it, but it is huge amount of places). I also prefer first form 😉

PR #570, opened 2017-04-12
[WIP] Fstab agent - handle better white space only lines (bsc#1030425)

lslezak:
The lines containing just spaces are not handled by the agent and the parser stops ignoring the rest of the file. See [bsc#1030425](https://bugzilla.suse.com/show_bug.cgi?id=1030425) for more details.

This patch defines space and tab at the beginning as a comment so a line containing only spaces is ignored.

### Disadvanatages

The drawback is that a space at the beginning of a valid line makes it a comment as well. But that's actually not a regression, the original code did not handle it as well and the rest of the file was ignored.
With this change it is still not perfect, but at least it parses the other lines so it is still an improvement.

### More Fixes Still Needed

This patch does not fix the upgrade issue completely, the same fix needs to applied at https://github.com/yast/yast-update/blob/master/src/modules/RootPart.rb#L819.

### Note

I tried a different approach using `:Or()` any_agent feature but the result then contained empty strings (in the list of Hashes!) or with `:Tuple()` it contained an empty hash. This could break some code as the `ret["file"]` would result in `nil` which might not be handled at all places.

But as this issue is there probably since the beginning of YaST2 this simple fix is good enough. For more robust fstab parsing we should switch to Augeas, it already has a fstab lense defined...

### DIFF
--- a/library/general/src/scrconf/etc_fstab.scr
+++ b/library/general/src/scrconf/etc_fstab.scr
@@ -42,6 +42,12 @@
  */
 .etc.fstab
 
+/*
+ * Note: the same agent definition is used in
+ * https://github.com/yast/yast-update/blob/e2052274034f5240e9f09bceae7c6d888b18468d/src/modules/RootPart.rb#L812
+ * apply any fixes also there.
+ */
+
 `ag_anyagent(
   `Description (
       (`File("/etc/fstab")),	// real file name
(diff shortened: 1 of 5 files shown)


jreidinger:
LGTM, test passes, so just add changelog and increase version

PR #571, opened 2017-04-19
Add CWM::WrapperWidget for old hash style CWM widgets (bsc#1032725)

jreidinger:
tested manually on added example file

[bsc#1032725](https://bugzilla.suse.com/show_bug.cgi?id=1032725)


kobliha:
I'm missing tests here ... I know it's a wrapper, but it's a new code and the rule for new code is: Unit Tests

jreidinger:
@kobliha as I write in description. I tested it manually with example that is included in pull request. Problem with testing CWM is that it heavily depend on UI and unit:tests are in this case heavily mocked, so quite useless.

jreidinger:
@kobliha btw if you checkout my branch, you can try it yourself. example is self-contained and can be easily run

jreidinger:
@kobliha I add some trivial test for most important methods.

library/cwm/src/lib/cwm/widget.rb
```diff
@@ -258,6 +258,60 @@ def validate_wrapper(_widget_id, _event)
     end
   end
 
+  # Wrapper when combination of old hash based CWM definition needed to be used
+  # with new object based one. Useful e.g. when content is provided by other
+  # module like CWMFirewallInterfaces.CreateOpenFirewallWidget
+  # @note it does not support common methods and helpers from abstract widget
```
mvidner: Ugh! So many `unsupported` exceptions. I suspect we are going to be arrested for violating the [law of Ms Liskov](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

jreidinger: Well, do you have other idea? Another option is to ignore call of that method, but I find it too error prone.

mvidner: Maybe we should add an ancestor to AbstractWidget and subclass this one from it

jreidinger: To be honest, I would like to keep object hierarchy clean. Another ancestor can be a bit confusing. What name you have in mind? It can maybe help if it have good name.

jreidinger:
@mvidner I updated it, now number of unsupported methods are much lower. Will merge master and adapt file structure when we agreed on code.

library/cwm/src/lib/cwm/wrapper_widget.rb
```diff
@@ -0,0 +1,46 @@
...
+  #
+  # @example how to initialize object from firewall interface
+  #   ::CWM::WrapperWidget.new("firewall",
+  #     CWMFirewallInterfaces.CreateOpenFirewallWidget("services" => ["service:sshd", "service:ntp"])
+  #   )
+  class WrapperWidget < AbstractWidget
+    # Creates new instance with specified id and content
+    # @param widget_id [String] name of widget used as identified, have to be unique.
+    #   It have to be same as real widget id in content, otherwise enable/disable won't work.
+    # @param content [Hash<String, Object>] CWM hash definition
```
mvidner: So is the type in fact a [WidgetHash](http://www.rubydoc.info/github/yast/yast-yast2/CWM/WidgetHash)

I hate the "content" word, it has so many meanings. That is why introduced the type, to have some clarity

library/cwm/src/lib/cwm/wrapper_widget.rb
```diff
@@ -0,0 +1,46 @@
...
+    def initialize(widget_id, content)
```
mvidner: ReplacePoint uses keyword args, perhaps we should make this consistent

library/cwm/src/lib/cwm/wrapper_widget.rb
jreidinger: In fact I prefer to have here `StringHash`. I am not sure if it will work with widgets inside and what is more important I really do no want to support.

library/cwm/src/lib/cwm/wrapper_widget.rb
jreidinger: fair point, will modify it

library/cwm/src/lib/cwm/wrapper_widget.rb
```diff
@@ -14,11 +14,12 @@ module CWM
...
   class WrapperWidget < AbstractWidget
     # Creates new instance with specified id and content
-    # @param widget_id [String] name of widget used as identified, have to be unique.
+    # @param id [String] name of widget used as identified, have to be unique.
     #   It have to be same as real widget id in content, otherwise enable/disable won't work.
-    # @param content [Hash<String, Object>] CWM hash definition
-    def initialize(widget_id, content)
-      self.widget_id = widget_id
+    #   If nil is used, it use default widget_id from class name.
+    # @param content [CWM::StringTerm] CWM hash definition
```
mvidner: No, a widget **hash** is not the same thing as a widget **term**, or a string term! I even linked to the thing so I will link again.
http://www.rubydoc.info/github/yast/yast-yast2/CWM/WidgetHash <- this is wanted here, because cwm_definition uses it
http://www.rubydoc.info/github/yast/yast-yast2/CWM/WidgetTerm
http://www.rubydoc.info/github/yast/yast-yast2/CWM/StringTerm

jreidinger: ah, that think. Naming is always hard. Will fix it.

jreidinger:
@mvidner thanks

PR #573, opened 2017-05-12
Add CWM::Tree

mvidner:
[boo#1039901](https://bugzilla.opensuse.org/show_bug.cgi?id=1039901)

The basic idea is to have a Tree on the left automatically handle content switching on the right, similarly to what Tabs do.

This is needed for https://trello.com/c/9HhmcdOP/978-8-new-storage-new-expert-partitioner-1-read-only-system-view , implemented in https://github.com/yast/yast-partitioner/pull/2


library/cwm/examples/object_api_tree.rb
```diff
@@ -0,0 +1,164 @@
...
+    include Yast::UIShortcuts
+    include Yast::Logger
+
+    def run
+      textdomain "example"
+
+      lucky_number_tab = LuckyNumberTab.new
+      true_love_tab = TrueLoveTab.new
+
+      tabs = ::CWM::Tree.new(lucky_number_tab, true_love_tab)
```
mvidner: This file is just a copied object_api_tabs.rb with `CWM::Tree` instead of `CWM::Tabs` on this line.

jreidinger:
Few comments to top level ideas of this pull request ( I do not  comment anything code specific ):

- I do not like general idea of having tree widget, that is more then tree widget, but widget with some dynamic content. Tree widget should be really only Tree widget with items. If you think that tree with dynamic content of right side is general useful ( I prefer to see it at least at two places to see requirements for API). Then I prefer name like `TreeWithDynamicContent` or `ContentBasedOnTree`

- Using Tab child for content is bad and confusing idea. If you need it, I suggest to have `DynamicContent` and `Tab` as its child ( even if it is just mark without own content ).

- TreeItem with specific arguments looks good

- splitting classes to multiple files is good

mvidner:
Thanks! I agree with your criticism. This is really a dirty prototype that I pushed out at the end of the week.

To structure the classes better I've  had a look now at how Qt does it, in particular how it uses widgets to switch among  visibility of other widgets: A [QTabWidget]( http://doc.qt.io/qt-5/qtabwidget.html) is a compound of a [QTabBar](http://doc.qt.io/qt-5/qtabbar.html) that controls a [QStackedWidget](http://doc.qt.io/qt-5/qstackedwidget.html). (Compare CWM::Tabs that uses UI::DumbTab or UI::PushButton to manage a set of CWM::Tab)

For trees, Qt does not seem to have anything controlling a QStackedWidget. Instead it concentrates on a model-view abstraction for the trees ([QTreeWidget](http://doc.qt.io/qt-5/qtreewidget.html) inherits [QTreeView](http://doc.qt.io/qt-5/qtreeview.html)).

I think we will have `CWM::Tree` - just a tree, with `CWM::TreeItem`, an item contains a member `data`: `Object` that it leaves alone. A `CWM::TreePager` will be a combination of a `Tree` whose `TreeItem#data` point to(*) a `CWM::Page` (maybe a special class will not be needed and an `AbstractWidget` will be enough, but then we do have a `CWM::Tab` class already)

(*) this will be a lambda returning the page, as for trees it mostly makes sense to have dynamic pages, as opposed to a tab set.

mvidner:
I have removed the laziness. I think we will need to have a look at that again, but for now it is a distraction.

library/cwm/src/lib/cwm/page.rb
```diff
@@ -0,0 +1,53 @@
...
+  # {Tabs} is a {Pager} and a {Tab} is its {Page}
+  # (FIXME: We haven't made Tabs a subclass of Pager yet.
+  # That will come after we have TreePager working.
+  class Page < CustomWidget
+    # @return [Boolean] is this the initially selected tab
+    attr_accessor :initial
+
+    # @return [WidgetTerm] contents of the tab, can contain {AbstractWidget}s
+    attr_reader :contents
+    # @return [String] label defines name of tab header
```
jreidinger: what it stand for Tree? I think tree doesn't have Label. Only it can have heading, but it should not be part of pager, as it is not mandatory.

library/cwm/src/lib/cwm/page.rb
```diff
@@ -0,0 +1,53 @@
...
+    attr_reader :label
+    attr_reader :widget_id
```
jreidinger: why it is needed?

jreidinger: Abstract and also Custom widget already have widget_id, so why redefine it? https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm/abstract_widget.rb#L66

library/cwm/src/lib/cwm/pager.rb
```diff
@@ -0,0 +1,146 @@
...
+    # gets id of initial page
+    # This default implementation returns first page passed to constructor
+    def initial_page_id
+      initial = @pages.find(&:initial)
+
+      (initial || @pages.first).widget_id
+    end
+
+    def contents
+      if Yast::UI.HasSpecialWidget(:DumbTab)
```
jreidinger: In general this should be only in Tabs and not in Pager. As pager should be generic.

library/cwm/src/lib/cwm/pager.rb
```diff
@@ -0,0 +1,146 @@
...
+      @pages.find { |t| t.widget_id == id }
+    end
+
+  private
+
+    def replace_point_id
+      :_cwm_page_contents_rp
+    end
+
+    def replace_point
```
jreidinger: when I implement Tabs ReplacePoint as widget does not exist. But now it exist, so it make sense to use it and its methods

library/cwm/src/lib/cwm/tree.rb
```diff
@@ -0,0 +1,62 @@
...
+    # FIXME: CurrentBranch? item id uniqueness?
+    # TODO: extract value/value= to CurrentItemBasedWidget
+    # or declare: value_property :CurrentItem
+    def value
+      Yast::UI.QueryWidget(Id(widget_id), :CurrentItem)
+    end
+
+    def value=(val)
+      Yast::UI.ChangeWidget(Id(widget_id), :CurrentItem, val)
+    end
```
jreidinger: I hope there will be also something like change_items which is also in ItemBasedSelection https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm/widget.rb#L71

Reason is when we modify anything on left side of pager, we have to redraw left side. ( Or create it completely from scratch ).

jreidinger: and one desing question. Should we return here id or complete TreeItem that was passed before?
Later can help us, if create subclass of TreeItem that have in addition widget to replace and then just can call `replace_point.replace(tree.value.widget)`

library/cwm/src/lib/cwm/tree_pager.rb
```diff
@@ -0,0 +1,50 @@
...
+    # @param children [Array<PagerTreeItem>]
+    def initialize(page, icon: nil, open: true, children: [])
+      @page = page
+      super(page.widget_id, page.label,
+            icon: icon, open: open, children: children)
+    end
+
+    def pages
+      children.values.flat_map(&:pages).unshift(@page)
+    end
```
jreidinger: what is purpose of this method?

library/cwm/src/lib/cwm/page.rb
mvidner: I used the two attrs, label and widget_id, for the [PagerTreeItem constructor](https://github.com/yast/yast-yast2/blob/e44e8a0f15015f116a8685b4ca358217101f1db4/library/cwm/src/lib/cwm/tree_pager.rb#L14).

library/cwm/src/lib/cwm/pager.rb
mvidner: Ah, good point. But I'd like to postpone it until the next sprint.

library/cwm/src/lib/cwm/tree_pager.rb
mvidner: I needed it for Pager constructor; added a comment.

library/cwm/src/lib/cwm/page.rb
mvidner: Explained in a comment

library/cwm/src/lib/cwm/pager.rb
mvidner: Duh, right! Fixed.

library/cwm/src/lib/cwm/tree.rb
mvidner: I don't understand. Are you proposing to change other widgets based on changes in this one? IMHO that doesn't belong in a plain Tree; possible to do it in a derived class.

library/cwm/src/lib/cwm/page.rb
jreidinger: It is fine for me, but `CustomWidget` already have widget_id, so it should work without it. And if you need to assign widget_id. You can do it in initialize like `self.widget_id = widget_id` even if it is still against idea of widget_id. Which is by default taken from class name. So if you have `class DisksPage < Page` it will have widget_id "DisksPage" and if it have multiple instance in one UI, then it can set it on its own.

library/cwm/src/lib/cwm/page.rb
jreidinger: ??? this comment mention tab header, but this is generic Page and as mentioned above Tab is Page, but not vice verse. So I do not see explained it anywhere in comment.

library/cwm/src/lib/cwm/tree.rb
jreidinger: no, what I propose is to have method `change_items` that allows to get list of TreeItems and change items in widget. Nothing more.

jreidinger: similar like we have in CWM::Table

library/cwm/src/lib/cwm/page.rb
mvidner: Ah! Somehow I missed the `@widget_id || ` part of `AbstractWidget#widget_id` and only thought of the `|| self.class.to_s`. So this attr_reader is superfluous :)

mvidner:
Argh, I haven't written any tests yet.

library/cwm/src/lib/cwm/page.rb
```diff
@@ -0,0 +1,56 @@
...
+  class Page < CustomWidget
+    # @return [Boolean] is this the initially selected tab
+    attr_accessor :initial
+
+    # @return [WidgetTerm] contents of the tab, can contain {AbstractWidget}s
+    attr_reader :contents
+    # @return [String] Label of {Tab} or of {PagerTreeItem}
+    attr_reader :label
+
+    def initialize(widget_id:, label:, contents:)
```
mvidner: @jreidinger if I understood you correctly a generic Page should do nothing in its constructor and have subclasses override the methods, right?

Only in this class but others too?

jreidinger: yes. In general. You can for trees have bunch of Pages. And e.g. content can be quite complex to compute. When it is method, it is computed when needed.

library/cwm/src/lib/cwm/tree.rb
mvidner: Done

mvidner: Now I finally understand: you are proposing that the type handled by `value`, and `value=` be `TreeItem` instead of the **id** of a `TreeItem`.
Well, for my current usage (` id = @tree.value if id == @tree.widget_id `) **id** is better.

library/cwm/src/lib/cwm/tabs.rb
```diff
@@ -1,180 +1,76 @@
...
+  class Tabs < Pager
+    # {Tabs} does not have instances:
+    # {Tabs.new} overrides {Class.new} and calls
+    # either {DumbTabPager.new} or {PushButtonTabPager.new}.
+    def self.new(*args)
+      if Yast::UI.HasSpecialWidget(:DumbTab)
+        DumbTabPager.new(*args)
+      else
+        PushButtonTabPager.new(*args)
       end
```
jreidinger: 👍

library/cwm/src/lib/cwm/tabs.rb
```diff
@@ -1,180 +1,76 @@
...
-    # validates current tab
-    def validate
-      Yast::CWM.validateWidgets(@current_tab.cwm_definition["widgets"], "ID" => @current_tab.widget_id)
     end
+  end
 
-    def help
-      @current_tab ? @current_tab.help : ""
-    end
+  # A {Pager} for the GUI, using the DumbTab widget
```
jreidinger: I really recommend to add here note, to not use it directly. And use Tabs instead, which smartly decide which tabs use.

library/cwm/src/lib/cwm/tabs.rb
```diff
@@ -1,180 +1,76 @@
...
-        if @current_tab
-          Yast::UI.ChangeWidget(
-            Id(@current_tab.widget_id),
-            :Label,
-            @current_tab.label
-          )
-        end
+    # @param page [Page]
+    def mark_page(page)
+      if @current_page
```
jreidinger: this method do not use at all `page` is it correct?

mvidner: It does, but the diff makes it really hard to see. Look here: https://github.com/yast/yast-yast2/blob/20d59b6d77c7e3e44f937ab1a2a45d67ae233715/library/cwm/src/lib/cwm/tabs.rb#L62

jreidinger: now I see and method make sense. Thanks

library/cwm/src/lib/cwm/pager.rb
mvidner: I tried but it did not work... and that's actually because we have violated Liskov substitution with Page#contents (as with Tab#contents earlier): it returns a WidgetTerm but there are callers that expect a UITerm instead.

jreidinger: in such case I suggest to change CustomWidget to allow to get WidgetTerm, as it is compatible with UITerm.

library/cwm/src/lib/cwm/tree.rb
```diff
@@ -0,0 +1,71 @@
...
+    # @return [String]
+    attr_reader :label
+    # @return [String] icon filename
+    attr_reader :icon
+    # @return [Boolean] is the subtree open?
+    attr_reader :open
+    # @return [Hash{id => TreeItem}]
+    attr_reader :children
+
+    def initialize(id, label, icon: nil, open: true, children: [])
```
jreidinger: why default for open is true and not false? it should be at least documented.

library/cwm/src/lib/cwm/tree_pager.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  # A {Pager} that uses a {Tree} to select the {Page}s
+  class TreePager < Pager
+    # @param items [Array<PagerTreeItem>]
+    def initialize(*items, label:)
+      @tree = Tree.new(items, label: label)
```
jreidinger: Ah, I notice it now. Again this design is a bit against CWM phylosophy To have stuff in constructor. So what is my idea how it can looks like? items should be defined in instance subclass same like we have in all items based widgets.
so
```
class MyTree < Tree
  def items
   [
      ...
   ]
  end

  def label
    _("Tree label")
  end
end
```
same way with label.

This way it really confuse me why TreePager have label and now I see that it is label for tree and not for Tree pager ( so different location of pager then I expect ).

Question is then how should look like items when tree is used in TreePager as it need to have different items. Well, answer is that tree already have to know about it, as it is tree widget designed to be used in Pager context. So it is fine for it, if it create TreePagerItems and not simple TreeItems.
That also bring me, that it can be quite easy to add helper to Tree itself to create TreeItem as item and for tree for pager, mention that it have to be pager_item.

jreidinger:
@mvidner road to NUE was a long so I have a time to thining about this API 😨 
## Short summary
what I in the end was my final idea. 

1. Page should be in fact just common CustomWidget, that just return WidgetUI. So if we modify CustomWidget, it can be just alias
2. whole concept of passing to Tabs class Tab instances is wrong and also that Tree child knows about other pages in the end looks wrong for me. My final idea is that it is pager responsibility

So how this can work? Basically difference between TreePager and TabsPager should be only that TreePager child gets tree and have difference contents. Tabs pager should not get anything in constructor.

What should have Pagers in common is method `page_for` that returns page for given id. For tabsPager it should be tab id. For treePager it should be `tree.value`. and this page_for method is used to replace replace point.

## Why I think it is better

Lets define "caller" code that use that Pager instance. Pagers can be more complex. For TabsPager it can have logic that decide which tabs to show. I feel a wrong that this logic will be in caller and not in TabsPager child itself. Fot TreePager on other hand I feel it is a bit wrong to have this mapping tree value and page in widget itself, as it should be that bussiness ( even it do not look so wrong when we discuss it during morning ).
Another aspect is who should have knowledge about Pages. I feel that correct place is pager child itself to have logic value -> page. Now for TabsPager it is Tab itself, but caller have to know which tabs should be shown. So TabsPager itself is not much reusable. For TreePager it is tree itself, which I feel limit a bit reusability of Tree itself and also it is against single purpose principle.
And last but not least, all layers know only exactly one its layer below. So calle know pager, pager knows pages ( and tree ), tree does'n know anything about pager pages. This way dependencies between classes are really minimal.

## Examples

Let me show you few examples I have in mind when I am thinking about design.

```ruby
# Tree example using new design
class DiskPage < CWM::Page
  def contens
    ...
  end

  # no label method, as Page do not have label itself. See that Page can be used in Tree or Tab
  # no problem with label with or without shortcut.
 
end

class MyDeviceTree < CWM::Tree
  def items
     [
        item_for("disk:/dev/sda", _("/dev/sda")],
     ]
  end
end

class MyDevicePager < CWM::TreePager
  def initialize(device_graph)
    super(MyTree.new(device_graph))
  end

  def page_for(id)
    case id
    when /^disk:/
      DiskPage.new(device_graph.find(id[/^disk:(.*)/))
    end
  end
end

# and caller
tabs = MyDevicePager.new(device_graph)
# and use this widget somewhere
```

```ruby
# Tabs example using new design
class Page1 < CWM::Page
  def contens
    ...
  end

  # no label method, as Page do not have label itself
 
end

class MyTabs < CWM::TabsPager
  def items
     [
        tab(:tab1, _("My cool tab")],
     ]
  end

  def page_for(id)
    return Page1.new if id == :tab1
  end
end

# and caller
tabs = MyTabs.new
# and use this widget somewhere
```

## result
I know that it will need even more time and I plan to invest it myself to it and that design changed quite lot. So if it blocks you in your work, feel free to merge it and I will later modify it and adapt all its usage.
I also know that modification of Tabs means, that we need to modify bootloader that uses it and I plan to do it. I am just not sure if it should be in SP3 or wait for SLE15. So I will give LGTM if you want to merge it in this state.

mvidner:
@jreidinger Well, your proposal is really a big change from the design until now, so I don't want to delay this PR further.
I will merge this to be able to implement more of the partitioner, and we can improve the class collaboration in the next iteration.

PR #574, opened 2017-05-15
Split off some CWM widgets to separate files.

mvidner:
For CWM::Tree (https://github.com/yast/yast-yast2/pull/573) and CWM::Table (TBD).


mvidner:
Ah, files missing from Makefile.am. Fixing.

mvidner:
@jreidinger here's the file splitting.

PR #555, opened 2017-03-20
Shared RSpec examples for CWM widgets

mvidner:
to enable stupid but easy test coverage

PR #575, opened 2017-05-15
implement cwm table

jreidinger:
[bsc#1039901](https://bugzilla.suse.com/show_bug.cgi?id=1039901)
for https://trello.com/c/9HhmcdOP

### DIFF
--- a/library/cwm/src/lib/cwm/table.rb
+++ b/library/cwm/src/lib/cwm/table.rb
@@ -56,6 +56,15 @@ module CWM
       Yast::UI.ChangeWidget(Id(widget_id), :SelectedItems, Array[id])
     end
 
+    # modifies single cell
+    # @param id [Object] id of row ( the first element in #items )
+    # @param collumn_number [Integer] index of cell in row. Index start with 0 for first cell.
+    # @param cell_content [String, Yast::Term, Object] Content of cell. Support same stuff as #items
+    # @note more efficient for bigger tables then changing everything with #change_items
+    def modify_cell(id, collumn_number, cell_content)
+      Yast::UI.ChangeWidget(Id(widget_id), Cell(id, collumn_number), cell_content)
+    end
+
   protected
 
     # helper to create icon term
@@ -66,6 +75,9 @@ module CWM
 
     # helper to create icon term
     # @param args content of cell, often used to combine icon and string
+    # @note Please see difference between `Cell` and `cell`. The first one is
+    #   used in queries to pick exact cell, but later is used for content of cell.
+    #   For reasons ask libyui authors.
     #
     # @example
     #   cell(icon("/tmp/cool_icon.png"), "Really cool!!!")
--- a/library/cwm/src/lib/cwm/table.rb
+++ b/library/cwm/src/lib/cwm/table.rb
@@ -41,6 +41,21 @@ module CWM
       Yast::UI.ChangeWidget(Id(widget_id), :Items, format_items(items_list))
     end
 
+    # gets id of selected item in table or array of ids if multiselection option is used
+    # @return [Array<Object>, Object] array if multiselection? return true
+    def value
+      val = Yast::UI.QueryWidget(Id(widget_id), :SelectedItems)
+      multiselection? ? val : val.first
+    end
+
+    # sets id of selected item(-s) in table
+    # @param id [Object, Array<Object>] selected id, if multiselection? is true
+    #   it require array of ids to select
+    def value=(id)
+      Yast::UI.ChangeWidget(Id(widget_id), :SelectedItems, Array[id])
+    end
+
+
   protected
 
     # helper to create icon term
@@ -56,6 +71,13 @@ module CWM
     #   cell(icon("/tmp/cool_icon.png"), "Really cool!!!")
     def cell(*args)
       Yast::Term.new(:cell, *args)
+    end
+
+    # helper to say if table have multiselection
+    def multiselection?
+      return false unless respond_to?(:opt, true)
+
+      return opt.include?(:multiSelection)
     end
 
   private
--- a/library/cwm/src/lib/cwm/table.rb
+++ b/library/cwm/src/lib/cwm/table.rb
@@ -65,6 +65,18 @@ module CWM
       Yast::UI.ChangeWidget(Id(widget_id), Cell(id, collumn_number), cell_content)
     end
 
+    # Resulting table as YUI term.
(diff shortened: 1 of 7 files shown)


library/cwm/src/lib/cwm/table.rb
```diff
@@ -0,0 +1,52 @@
+require "abstract_method"
+require "cwm/abstract_widget"
```
lslezak: I cannot see any reference to `AbstractWidget` or did I overlook something?

jreidinger: you are right.

library/cwm/src/lib/cwm/table.rb
```diff
@@ -0,0 +1,79 @@
+require "abstract_method"
+require "cwm/custom_widget"
+
+module CWM
+  # Represents Table widget
+  class Table < CustomWidget
+    # @method header
+    #   return array of String or Yast::Term which is used as headers for table
+    #   it can use e.g. align Left/Center/Right
+    # @note have to be overwritten
```
mvidner: it **has** to be *overridden*

library/cwm/src/lib/cwm/table.rb
```diff
@@ -0,0 +1,79 @@
+require "abstract_method"
+require "cwm/custom_widget"
+
+module CWM
+  # Represents Table widget
+  class Table < CustomWidget
+    # @method header
+    #   return array of String or Yast::Term which is used as headers for table
```
mvidner: > @return [Array<String|Yast::Term>]

library/cwm/src/lib/cwm/table.rb
```diff
@@ -0,0 +1,79 @@
...
+    #       [:first_user, "Joe", "Doe"],
+    #       [:best_user, "Chuck", "Norris"]
+    #     ]
+    #   end
+    def items
+      []
+    end
+
+    # change list on fly with argument. Useful when content of widget is changed.
+    # @arg items_list [Array<Array<Object>>] same format as {#items}
```
mvidner: Not `@arg` but `@param ...`, in several places.

mvidner:
I am missing any information how to work with the table once it is constructed. Which row is selected? How do I change or query a single cell?
Remember how we broke some modules because we confused `:cell` and `:Cell`? We may want to document it here.

jreidinger:
@mvidner you are right, `value` and `value=` is missing here. Can you be more specific what you mean with query or change single cell? Currently it just support change of all of them, but if there is good example what you want to achieve with single cell, I can add it. I just would like to see some real world usage, so it make design better.

library/cwm/src/lib/cwm/table.rb
```diff
@@ -0,0 +1,113 @@
...
+    # helper to say if table have multiselection
+    def multiselection?
+      return false unless respond_to?(:opt, true)
+
+      opt.include?(:multiSelection)
+    end
+
+  private
+
+    def contents
```
mvidner: `#contents` is part of CustomWidget API so it should not be private.

I found this while testing this with https://github.com/yast/yast-yast2/pull/555 (shared RSpec tests) (yes, I will finalize it today) so for testability we should even make `#contents` public.

jreidinger: well, it is part of API that need to be implemented, but for me something, that should not be publicly exposed. It is only private data passed to CWM. I do not see any other user for this method except tests, or do you have something in mind?
For tests you can use `send` or `respond_to?(:contents, true)`.

library/cwm/src/lib/cwm/table.rb
mvidner: (done)

mvidner:
If you merge `master`, you can add some test coverage with

```rb
#! /usr/bin/env rspec

require_relative "test_helper"

require "cwm/rspec"
require "cwm/table"

describe CWM::Table do
  class MyTable < CWM::Table
    def header
      ["English", "Deutsch"]
    end

    def items
      [
        [:one, "one", "eins"],
        [:two, "two", "zwei"]
      ]
    end
  end
  subject { MyTable.new }

  include_examples "CWM::CustomWidget"
end
```

(funny point, `def items` does not help coverage because then your default implementation is not used. I think it even makes sense to make it an abstract_method instead. I don't see use for an empty table)

jreidinger:
@mvidner idea with empty table is if you have content that is computation expensive and you want to provide early feedback. In such case keep default `items` and call `change_items` in `init`. This way it shows empty table and rest of widgets and then start filling them with proper busy mouse icon, etc.

library/cwm/src/lib/cwm/table.rb
```diff
@@ -0,0 +1,119 @@
...
+    #       [:best_user, "Chuck", "Norris"]
+    #     ]
+    #   end
+    def items
+      []
+    end
+
+    # change list on fly with argument. Useful when content of widget is changed.
+    # @param items_list [Array<Array<Object>>] same format as {#items}
+    def change_items(items_list)
```
mvidner: I am still not sure how to handle the distinction between the _initial_ `items` and the _current_ `change_items`.
... Ah! it is in fact consistent with the [ItemsSelection](http://www.rubydoc.info/github/yast/yast-yast2/CWM/ItemsSelection) mixin. Let's mention that name somewhere.

library/cwm/src/lib/cwm/table.rb
```diff
@@ -0,0 +1,119 @@
...
+    # sets id of selected item(-s) in table
+    # @param id [Object, Array<Object>] selected id, if multiselection? is true
+    #   it require array of ids to select
+    def value=(id)
+      Yast::UI.ChangeWidget(Id(widget_id), :SelectedItems, Array[id])
+    end
+
+    # modifies single cell
+    # @param id [Object] id of row ( the first element in #items )
+    # @param collumn_number [Integer] index of cell in row. Index start with 0 for first cell.
```
mvidner: spelling: column (single L)

library/cwm/src/lib/cwm/table.rb
```diff
@@ -0,0 +1,119 @@
...
+    # @param cell_content [String, Yast::Term, Object] Content of cell. Support same stuff as #items
+    # @note more efficient for bigger tables then changing everything with #change_items
+    def modify_cell(id, collumn_number, cell_content)
```
mvidner: for consistency with ItemsSelection this should be named `change_cell`

PR #576, opened 2017-05-17
Document the 3 kind of Terms used in CWM

mvidner:
I needed to understand this for implementing the CWM::TreePager switching. Even the current situation is a bit tricky.

- UITerm: can be passed as is to Yast::UI methods
- WidgetTerm: contains instances of CWM::AbstractWidget
- StringTerm: contains strings which identify the old style hash based CWM widgets

### DIFF
--- a/library/cwm/src/lib/cwm/abstract_widget.rb
+++ b/library/cwm/src/lib/cwm/abstract_widget.rb
@@ -1,6 +1,22 @@
 require "abstract_method"
 
 module CWM
+  # A Yast::Term that can be passed as is to Yast::UI methods
+  # (OpenDialog, ReplaceWidget)
+  class UITerm < Yast::Term; end
+
+  # A Yast::Term that contains instances of {CWM::AbstractWidget}.
+  # Can be passed to {Yast::CWMClass#show Yast::CWM.show}
+  class WidgetTerm < Yast::Term; end
+
+  # A Yast::Term that contains strings
+  # which identify the old style hash based CWM widgets.
+  # Can be passed to {Yast::CWMClass#ShowAndRun Yast::CWM.ShowAndRun}
+  class StringTerm < Yast::Term; end
+
+  # A {Hash{String=>Object}} that {Yast::CWMClass} knows to handle
+  class WidgetHash < Hash; end
+
   # Represent base for any widget used in CWM. It can be passed as "widget"
   # argument. For more details about usage
   # see {Yast::CWMClass#show Yast::CWM.show}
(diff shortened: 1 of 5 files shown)


library/cwm/src/lib/cwm/abstract_widget.rb
```diff
@@ -1,6 +1,22 @@
 require "abstract_method"
 
 module CWM
+  # A Yast::Term that can be passed as is to Yast::UI methods
+  # (OpenDialog, ReplaceWidget)
+  class UITerm < Yast::Term; end
```
mvidner: @jreidinger has in the past objected to this kind of "code for documentation  purposes only". So I tried to avoid this :) We are using `@!method` Yard macros elsewhere, but there is no `@!class`. You can use `@!parse class Foo; end` which does make the class appear... but the documentation is missing.

library/cwm/src/lib/cwm/abstract_widget.rb
```diff
@@ -1,6 +1,22 @@
...
+  # A Yast::Term that contains instances of {CWM::AbstractWidget}.
+  # Can be passed to {Yast::CWMClass#show Yast::CWM.show}
+  class WidgetTerm < Yast::Term; end
+
+  # A Yast::Term that contains strings
+  # which identify the old style hash based CWM widgets.
+  # Can be passed to {Yast::CWMClass#ShowAndRun Yast::CWM.ShowAndRun}
+  class StringTerm < Yast::Term; end
+
+  # A {Hash{String=>Object}} that {Yast::CWMClass} knows to handle
```
jreidinger: to be honest. This is just confusing. It make sense if you want to document what CWM actually support and knows in this hash. But it alone is just confusing, like if it is not just common Hash.

library/cwm/src/lib/cwm/abstract_widget.rb
jreidinger: well, I am fine with it. Maybe mention in those three terms, that it is three steps of processing.
`WidgetTerm -> StringTerm -> UITerm` first processing expands CWM::AbstractWidget and create StringTerm, which is later processed to UITerm by old part of CWM.

library/cwm/src/lib/cwm/abstract_widget.rb
jreidinger: and second note. I think it make sense to have it in CWMClass itself and not here. As format is not modified nor maintained here.

mvidner: Well, it `is_a? Hash`, which is IMHO expressed well enough by its name, and it is not a generic hash. It is a specific kind of hash that tries to be an object, so I think we are fully justified in documenting it in this way. 

Observe the API docs in CWM.rb where we have several kinds of hashes. I think this does improve the situation.

You are right that it would be good to document the actual members of the hash, but that is too much yak shaving for me now. 

Anyway, what would improve the situation for you and make you less confused?

library/cwm/src/lib/cwm/abstract_widget.rb
mvidner: You are right, I will add that

PR #578, opened 2017-05-26
Adds to Yast::Execute exclamation mark variants of methods

jreidinger:
lets open it first for discussion. I add these methods based on feedback from workshop on osc, where users complain, that they cannot in Yast::Execute easily check for errors, which is possible in cheetah.


jreidinger:
@ancorgs @lslezak you can be interested as you was on workshop

mvidner:
Thanks! I agree with the addition to the code.

But I have a complaint about the title of the PR being "extend_execute". WTF! Please take a few seconds to formulate meaningful titles of your pull requests. And draft the .changes entry too please.

jreidinger:
@mvidner agreed, it was just quickly capture idea on yast workshop on conference. I plan to polish it after some discussion about the change.

jreidinger:
I will create changes, when we slightly reduce number of pull requests, as there is a lot of them ready, so to avoid collisions.

jreidinger:
@mvidner thanks

PR #579, opened 2017-05-27
Cwm improvements

jreidinger:
@mvidner in general what is goal of this pull request is to

1. make clear what custom widget allows and extend examples to demonstrate it
2. fix replace point to play nicely with custom widget

that allows to significantly simplify Pager and Page itself ( goal is to not modify them in this pull request ).
I tested it manually all examples we currently have.


mvidner:
@jreidinger please have a look at the failing tests

library/cwm/src/Makefile.am
```diff
@@ -13,16 +13,20 @@ module_DATA = \
...
   lib/cwm/pager.rb \
+  lib/cwm/replace_point.rb \
   lib/cwm/table.rb \
   lib/cwm/tabs.rb \
   lib/cwm/tree.rb \
   lib/cwm/tree_pager.rb \
   lib/cwm/widget.rb
 
+ylib2dir = @ylibdir@
+ylib2_DATA = lib/cwm.rb
```
mvidner: Why not simply add `lib/cwm.rb` to `ylib_DATA`?

jreidinger: because it will be installed to /usr/share/YaST2/lib/cwm/cwm.rb ? Or I get it wrong?

mvidner: Oh, you're right. I've confused `ylibdir` and `@ylibdir@`, and currently `ylibdir = @ylibdir@/cwm`. That is a real mess. Lets make
```
ylibdir = @ylibdir@
cwmlibdir = @ylibdir@/cwm
```

jreidinger:
@mvidner it is ready now

library/cwm/src/lib/cwm/pager.rb
```diff
@@ -96,30 +74,30 @@ def switch_page(page_id)
...
       initial = @pages.find(&:initial)
 
-      (initial || @pages.first).widget_id
+      initial || @pages.first
     end
 
-    def page_for_id(id)
-      @pages.find { |t| t.widget_id == id }
+    # gets id of initial page
+    def initial_page_id
```
mvidner: Let's drop this and let the only caller use initial_page.widget_id (or even compare pages instead of their ids, see that caller)

library/cwm/src/lib/cwm/pager.rb
```diff
@@ -96,30 +74,30 @@ def switch_page(page_id)
...
-
-    def replace_point_id
-      :_cwm_page_contents_rp
+    def page_for_id(id)
+      @pages.find { |t| t.widget_id == id }
     end
 
     def replace_point
-      ReplacePoint(Id(replace_point_id), VBox(VStretch(), HStretch()))
+      @replace_point = ReplacePoint.new(id: "_cwm_page_contents_rp", widget: initial_page)
```
mvidner: `||=` !

library/cwm/src/lib/cwm/pager.rb
```diff
@@ -96,30 +74,30 @@ def switch_page(page_id)
     abstract_method :mark_page
 
     # The contents will probably include a *selector*, such as {Tabs}
-    # or {Tree} and a {ReplacePoint} where {Page}s will appear.
+    # or {Tree} and a {replace_point} where {Page}s will appear.
```
mvidner: ... or {Tree} and must include {#replace_point} where ...

library/cwm/src/lib/cwm/replace_point.rb
```diff
@@ -0,0 +1,69 @@
...
+    # @param widget [CWM::AbstractWidget] initial widget in placeholder
+    def initialize(id: "_placeholder", widget: Empty.new("_initial_placeholder"))
+      self.handle_all_events = true
+      self.widget_id = id
+      @widget = widget
+    end
+
+    # @return [UITerm]
+    def contents
+      ReplacePoint(Id(widget_id), Empty(Id("____CWM_empty")))
```
jreidinger: it make sense, this will also simplify init, as init will be done as part of CWM itself.

library/cwm/src/lib/cwm/replace_point.rb
```diff
@@ -0,0 +1,69 @@
...
+    # Switches to initial widgets
+    def init
+      replace(@widget)
+    end
+
+    # Replaces content with different widget. All its events are properly
+    # handled.
+    # @param widget [CWM::AbstractWidget] widget to display and process events
+    def replace(widget)
+      log.info "replacing with new widget #{widget.inspect}"
```
jreidinger: agreed

library/cwm/src/modules/CWM.rb
```diff
@@ -135,6 +139,9 @@ def ProcessTerm(t, widgets)
             arg = ProcessTerm(Convert.to_term(arg), widgets)
           end
         elsif Ops.is_string?(arg) # action
+          Builtins.y2error("find string #{arg} in term #{t.inspect} which is unknown.") unless widgets[arg]
```
jreidinger: OK, I want to have explicit error when there is a string and there are no widget defined with given id. CWM continue with replacing it with `VBox()` which is extremelly annoying for debug. So I want to have explicit error in log.

So message trying to say. "I found in StringTerm string which cannot be associated to any defined widget."

library/cwm/test/replace_point_test.rb
```diff
@@ -0,0 +1,117 @@
...
+      subject = described_class.new(widget: CWM::Empty.new(:initial))
+      expect(Yast::CWM).to receive(:ReplaceWidgetHelp)
+      subject.replace(widget)
+    end
+  end
+
+  describe "#handle" do
+    # rspec expect have problem with arity of expected classes
+    # so test failing
+    xit "Passes handle to CWM on active widget" do
```
jreidinger: in general it works great, only problem is rspec, which define its wrapper like `def handle(*args)` which confuse this code. My idea why to have this "magic" is that if someone is interested in event id, he can define method with it and if someone is not, just define it without.

jreidinger: in general I like this API more then having two separated methods like handle and handle_event which is more confusing for me.

library/cwm/src/lib/cwm/replace_point.rb
```diff
@@ -20,26 +20,18 @@ def initialize(id: "_placeholder", widget: Empty.new("_initial_placeholder"))
 
     # @return [UITerm]
     def contents
-      ReplacePoint(Id(widget_id), Empty(Id("____CWM_empty")))
-    end
-
-    # Switches to initial widgets
-    def init
```
mvidner: Now many tests fail.

jreidinger: yeah, in the end I think it make sense to really use init as it was before. Reason is that if it is part of content, then CWM itself handle it. So Replace point should not do it, but when it is replaced, it will be handled by replace point. Which is not so easy to see. With old approach that init do replace, then from beginning all content events and callbacks are handled in replace point.

library/cwm/src/lib/cwm/replace_point.rb
```diff
@@ -20,7 +20,16 @@ def initialize(id: "_placeholder", widget: Empty.new("_initial_placeholder"))
 
     # @return [UITerm]
     def contents
-      ReplacePoint(Id(widget_id), @widget)
+      # Use for contents empty widget, because if @widget passed to constructor
+      # will be used then CWM itself will handle events for this initial widget.
+      # This is against replace point idea that replace point due to its dynamic
+      # content will handle events itself
+      ReplacePoint(Id(widget_id), Empty(Id("___cwm_rp_empty")))
```
mvidner: OK, but now I come back to the reason why I originally suggested `@widget` here:
- I don't like the id being `"___cwm_rp_empty"` here and `"_initial_placeholder"` in the constructor.
- I don't understand why use `Empty`  here and `Empty.new` in the constructor. Well, they are a Term and AbstractWidget respectively, but I see accidental complexity of dummy widgets that are going to be replaced anyway. Can we DRY this?

library/cwm/src/lib/cwm/replace_point.rb
```diff
@@ -20,7 +20,16 @@ def initialize(id: "_placeholder", widget: Empty.new("_initial_placeholder"))
 
     # @return [UITerm]
     def contents
-      ReplacePoint(Id(widget_id), @widget)
+      # Use for contents empty widget, because if @widget passed to constructor
+      # will be used then CWM itself will handle events for this initial widget.
+      # This is against replace point idea that replace point due to its dynamic
+      # content will handle events itself
```
mvidner: Really? That was never my idea. If that is needed then you must document it in a top level class comment, not in a method implementation detail.

library/cwm/src/lib/cwm/common_widgets.rb
```diff
@@ -271,6 +271,8 @@ def cwm_definition
 
   # A selection of a value via radio buttons.
   # The {#label} method is mandatory.
+  # @note if radio buttons are modified during runtime, like with #change_items
+  #   then handle won't work correctly unless handle_all_events specified
```
mvidner: Please have a changes entry for this. "Fix handl**ing** events for radio buttons (that were changed with change_items ?)"

library/cwm/src/lib/cwm/replace_point.rb
mvidner: Ah, I think I understand now. Your English is so bad it takes me quite long to get what you mean:

```rb
# In `contents` we must use an Empty Term, otherwise CWMClass
# would see an {AbstractWidget} and handle events itself,
# bypassing our {#handle}
```

 (BTW I don't see why we need it but that's another topic)

library/cwm/src/lib/cwm/replace_point.rb
```diff
@@ -0,0 +1,70 @@
...
+
+module CWM
+  # Placeholder widget that is used to replace content on demand.
+  # The most important method is {#replace} which allows switching content
+  class ReplacePoint < CustomWidget
+    # @param id [Object] id of widget. Needed to redefine only if more than one
+    # placeholder needed to be in dialog. Parameter type is limited by component
+    # system
+    # @param widget [CWM::AbstractWidget] initial widget in placeholder
+    def initialize(id: "_placeholder", widget: Empty.new("_initial_placeholder"))
```
mvidner: Could we make the `widget` argument mandatory? IMHO most users will have a meaningful default widget, or can supply an explicit `Empty.new` themselves.

library/cwm/src/lib/cwm/replace_point.rb
jreidinger: @mvidner ok, so let me explain it. if contents returns WidgetTerm that include Widget, then this widget will be part of cwm map, so CWM will forward to it its events. But when it is replaced, it have to be handled by ReplacePoint. It is something that was not designed from start in CWM, so it need a bit additional handling on top of it.

I can dry this empty stuff.

library/cwm/src/lib/cwm/common_widgets.rb
jreidinger: no, actually it is such fix handling events. this note is that automatic id registration of buttons used in RadioButtons Group won't work if buttons was later changed, unless it is registered to get all events.

library/cwm/src/lib/cwm/replace_point.rb
jreidinger: this is backward incompatible change. But I check usage and looks like it is always used with it, so if you think that it doesn't make sense to have initial value that can be replaced later, I can remove it.

mvidner: Thank you!
You are right, the tests fail ;-)
```console
     ArgumentError:
       missing keyword: widget
     # /usr/src/app/library/cwm/src/lib/cwm/replace_point.rb:15:in `initialize'
     # ./replace_point_test.rb:48:in `new'
```

jreidinger:
@mvidner thanks

teclator:
@jreidinger thnx for the improvements!

PR #580, opened 2017-05-29
bsc#1036440: Warning messages shouldn't open UI in command-line mode

gilsonsouza:


library/general/test/report_test.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    context "while in command-line mode" do
+      before(:each) do
+        allow(Yast::Mode).to receive(:commandline).and_return(true)
+      end
+
+      it "prints the message only on console" do
+        expect(Yast::CommandLine).to receive("Print")
+          .with(/#{message}/)
+        expect(Yast::Popup).to_not receive("Warning")
```
imobachgs: Just for consistence, I would use `:Warning` symbol instead of the string.

library/general/test/report_test.rb
```diff
@@ -0,0 +1,105 @@
...
+        subject.Warning(message)
+      end
+    end
+
+    context "while in UI mode and timeout is disabled" do
+      let(:timeout) { 0 }
+
+      before(:each) do
+        subject.DisplayWarnings(show, timeout)
+        allow(Yast::Mode).to receive(:commandline).and_return(false)
```
imobachgs: If you are going to mock `Yast::Mode.commandline` in all cases, maybe you would like to move it to a before block and just define the value for each case.

```ruby
before do
  allow(Yast::Mode).to receive(:commandline).and_return(commandline?)
end

context "while in UI mode and timeout is disabled" do
  let(:timeout) { 0 }
  let(:commandline?) { false }

  it "shows a popup" do
    #...
  end
end
```

library/general/test/report_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+  end
+
+  describe ".Error" do
+    let(:show) { true }
+    let(:message) { "Message" }
+
+    context "while in command-line mode" do
+      before(:each) do
+        allow(Yast::Mode).to receive(:commandline).and_return(true)
```
imobachgs: Same than above. I would move this line to the outer before block.

package/yast2.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon May 29 08:19:46 UTC 2017 - gsouza@suse.com
+
+- bsc#1036440: Warning messages shouldn't open UI in command-line
```
imobachgs: We usually write it like this:

```
- Warning messages shouldn't open UI in command-line mode (bsc#1036440)
```

imobachgs:
Thanks for these tests!

PR #584, opened 2017-06-01
Cwm wrapper

jreidinger:


mvidner:
Well, I wonder why Travis did not catch that :-(
https://travis-ci.org/yast/yast-yast2/builds/238318455#L6005

jreidinger:
@mvidner reason is that spec file have *, so ti catch all of it. But make install, do not install it, so it cannot see that something missing.

PR #587, opened 2017-06-02
Event handler

joseivanlopez:
Add EventDispatcher#event_handler method to allow custom event management.

### DIFF
--- a/library/general/src/lib/ui/event_dispatcher.rb
+++ b/library/general/src/lib/ui/event_dispatcher.rb
@@ -37,14 +37,15 @@ module UI
 
       loop do
         input = user_input
-        event_handler(input)
+        handle_event(input)
         return @_finish_dialog_value if @_finish_dialog_flag
       end
     end
 
     # General dialog events handler.
     # Can be redefined to modify the way of managing events, for example when
-    # some events need to be delegated to a widget.
+    # some events need to be delegated to a widget. To use this method in
+    # combination with default *_handler methods, call to super.
     # @example delegate events
     #    class OKDialog
     #     include Yast::UIShortcuts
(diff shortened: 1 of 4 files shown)


library/general/src/lib/ui/event_dispatcher.rb
```diff
@@ -37,14 +37,53 @@ def event_loop
...
+    #       @widget.handler(input)
+    #       case input
+    #       when :ok
+    #         finish_dialog(:ok)
+    #       when :cancel
+    #         finish_dialog(:cancel)
+    #       end
+    #     end
+    #   end
+    def event_handler(input)
```
jreidinger: Please mention in doc, that if combination of *_handler method and own event_handler is used, then in event_handler super have to be called. ( e.g. having ok and cancer handler and then for table with `:notify` catch values.

joseivanlopez: Yes, but to avoid exceptions, `super` should be called only in proper cases: 

```
def event_handler(input)
  widget.handler(input)
  super if [:ok, :cancel].include?(input)
end
```

ancorgs: I don't think `event_handler` is a good name. What if you want to use the default handler but you have a widget with id `:event`?

It should be something not matching `xy_handler`. For example `handle_event`.

joseivanlopez: @ancorgs yes, I agree. I change it.

jreidinger: @joseivanlopez well, problem your code is that it is too error prone ( you add new handler and forgot to mention it in event_handler and you already have problem. So what I suggest is to have code like
```
if respond_to?(:"#{input}_handler")
  public_send(:"#{input}_handler")
elsif respond_to?(:"handle_event")
  public_send(:"handle_event", event)
else
  raise "Unknown action #{input}"
end
```

this way

1. keep old behavior if you do not have new generic handler
2. events with handler have precedence
3. make handle_event easier code.

e.g. you example will look now like
```
def handle_event(input)
  widget.handler(input)
end
```

Consider that handle_event will be written multiple time, but code in this class only once, so better to make it smarter here and more trivial code later, which handle only its own business.

library/general/test/event_dispatcher_test.rb
```diff
@@ -32,6 +32,14 @@ def user_input
   end
 end
 
+class DispatcherHandleEventTestDialog
+  include UI::EventDispatcher
+
+  def handle_event(_input)
+    finish_dialog(:other)
```
jreidinger: maybe just check here that input is other or simple `finish_dialog(input)` so test verify that proper input is passed.

PR #589, opened 2017-06-06
Nested replace point help

jreidinger:


library/cwm/src/lib/cwm/abstract_widget.rb
```diff
@@ -83,6 +83,15 @@ def self.widget_type=(type)
...
     #   @return [String] translated help text for the widget
 
+    # @!method dynamic_help
+    #   Dynamic help is useful when content of help text is not static and when
+    #   help is recomputed with {Yast::CWM#ReplaceWidgetHelp}, then it can
+    #   return different result. Examples where it is useful is e.g. replace point which
+    #   have dynamic content and different help text depending on what is displayed.
+    #   Another example is CWM::Table which can have dynamic header depending on conditions.
+    #   In those cases after call of {Yast::CWM#ReplaceWidgetHelp}, it will return new help text
+    #   @return [String] translated help text for the widget.
```
mvidner: This can't be right, the documented signature is the same as for `help`. maybe `@yieldreturn` ?

jreidinger: it is in fact same signature. Both are methods that return "string". Difference is that the first one is used to store its result into WidgetHash, but the second is passed as reference to WidgetHash. So in first case, later change of help content is not reflected, but the second one it is called repeatedly.

library/cwm/src/lib/cwm/abstract_widget.rb
```diff
@@ -83,6 +83,15 @@ def self.widget_type=(type)
     # @!method help
     #   @return [String] translated help text for the widget
```
mvidner:
```rb
#     ... Called only once, do not put dynamic content here.
#   @see #dynamic_help
```

library/cwm/src/lib/cwm/abstract_widget.rb
mvidner: Too many words. The docs should tell me: Do I need to implement this method? When?
```rb
#   @return [String] translated help text for the widget
#   If the help text for the widget may change over its lifetime, define `dynamic_help`
#   instead of `help`.
#   It will be called ...
```
 **uh oh**, I think if I implement this for a CustomWidget it will not help anyway
  because I would need to call ReplaceWidgetHelp myself , right?

library/cwm/src/lib/cwm/abstract_widget.rb
```diff
@@ -145,6 +154,8 @@ def cwm_definition
       res["_cwm_key"] = widget_id
       if respond_to?(:help)
         res["help"] = help
+      elsif respond_to?(:dynamic_help)
```
mvidner: I am still not decided if I want `AbstractWidget#help` to be polymorphic ("string" or "string ()") but it would be an easy change here: (plus deleting dynamic_help_method)

```rb
  if respond_to?(:help)
    h = help
    if h.respond_to? :call
      h = fun_ref(method(:help), "string ()")
    end
    res["help"] = h
  else
    res["no_help"] = ""
  end
```
```

File: library/cwm/src/lib/cwm/abstract_widget.rb (hunk 1 above)
Comment (jreidinger) [inline library/cwm/src/lib/cwm/abstract_widget.rb]: @mvidner tricky part is that it is possible that dynamic help can change as side-effect of another action. Like consider that you have partition. In some conditions, it can have RadioButtons with 4 options, but in other when e.g. filesystem changed to btrfs, then it will have 6 options, so help text will change. In this case only part you need to do is to call `ReplaceWidgetHelp` in filesystem handle and not in RadioButtons itself.

And that it is exactly what happen with replace point. As outer replace point help change as a resulf of change of inner one, but outer one do not know about it. So only inner one trigger "help can change" and it is recomputed.

File: library/cwm/src/lib/cwm/abstract_widget.rb (hunk 2 above)
Comment (mvidner) [inline library/cwm/src/lib/cwm/abstract_widget.rb]:
```rb
# Called only once by default. Also triggered by {#refresh_help}.
```

```rb
# A widget will need to call this if its {#help} text has changed,to make the change effective.
def refresh_help
```

library/cwm/src/modules/CWM.rb
```diff
@@ -230,22 +229,26 @@ def ValidateValueType(key, value, widget)
...
+        when "handle" then Ops.is(value, "symbol (string, map)")
+        when "store" then Ops.is(value, "void (string, map)")
+        when "cleanup" then Ops.is(value, "void (string)")
+        when "validate_function" then Ops.is(value, "boolean (string, map)")
+        when "items" then Ops.is(value, "list <list <string>>")
+        when "_cwm_do_validate" then Ops.is(value, "boolean (string, map <string, any>)")
+        when "help"
+          # help can be static string or dynamic callback, that allows to recompute help for
+          # widgets with dynamic contents like replace_point
+          if value.is_a?(::String)
```
mvidner: `value.is_a?(::String) || Ops.is(value, "string ()")` ;-)

library/cwm/src/modules/CWM.rb
```diff
@@ -469,13 +469,6 @@ def ValidateMaps(widgets)
         else
           ["label", "widget"]
         end
-        if !Builtins.haskey(v, "no_help")
```
mvidner: OK, we have been quietly adding no_help to AbstractWidgets anyway, so dropping that for WidgetHash is not a big problem

package/yast2.changes
```diff
@@ -1,4 +1,16 @@
 -------------------------------------------------------------------
+Tue Jun  6 10:33:22 UTC 2017 - jreidinger@suse.com
+
+- Fix showing help text when CWM::ReplacePoint contain another
+  CWM::ReplacePoint. Fix include new
+  CWM::AbstractWidget#dynamic_help functionality for widgets where
+  help text can change during its lifetime. Fix also include new
+  variant of Yast::CWM.ReplaceWidgetHelp without arguments, that
+  recompute help, which should be called if help text changed in
+  widgets with CWM::AbstractWidget#dynamic_help. (boo#1039901)
```
mvidner: (This will need an update. Remember http://tinyurl.com/this-verb-form-are-wrong ;-) )

library/cwm/src/lib/cwm/abstract_widget.rb
```diff
@@ -212,6 +221,10 @@ def handle_method
       fun_ref(method(:handle_wrapper), "symbol (string, map)")
     end
 
+    def help_method
+      fun_ref(method(:dynamic_help), "string ()")
```
mvidner: The good news is that this method is covered by tests.

jreidinger:
will merge after SP3 branch will be created

PR #592, opened 2017-06-19
CWM::Dialog; shared_examples for more widgets

mvidner:
These are needed for the Expert Partitioner, and were introduced in https://github.com/yast/yast-partitioner/pull/5 (because we had not branched SP3 back then)

- CWM::Dialog
- UI::Sequence
- RSpec.shared_examples for more CWM widgets
- UI::Greasemonkey


library/general/src/lib/ui/greasemonkey.rb
```diff
@@ -0,0 +1,225 @@
...
+# Summary:	Expert Partitioner
+# Authors:	Arvin Schnell <aschnell@suse.de>
+#
+# Lets see if this turns out to be useful.
+require "yast"
+
+module UI
+  # UI layout helpers.
+  #
+  module Greasemonkey
```
jreidinger: if you want to add GreaseMonkey to yast2, then significant improvement of code quality is needed ;) as yast2 mean that it will be used on more places and there is highed requirements. Also documentation have to be significantly improved and same for test coverage.

library/general/test/greasemonkey_test.rb
```diff
@@ -0,0 +1,151 @@
+#! /usr/bin/env rspec
+require_relative "test_helper"
+require "ui/greasemonkey"
+
+describe "UI::Greasemonkey" do
+  RSpec.shared_examples "a Greasemonkey method" do |mname|
+    it "transforms its argument properly" do
```
jreidinger: it should better document what happen like `it "transforms GreaseMonkey term to UI term"`

library/cwm/src/lib/cwm/dialog.rb
```diff
@@ -0,0 +1,97 @@
...
+  #
+  # @see UI::Dialog
+  # @see CWM::AbstractWidget
+  class Dialog
+    include Yast::Logger
+    include Yast::I18n
+    include Yast::UIShortcuts
+
+    # @return [String,nil] Set a title, or keep the existing title
+    abstract_method :title
```
jreidinger: does it make sense to have method that return nil? like `def title; nil; end`? why not having nil as default?

library/cwm/src/lib/cwm/dialog.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    def wizard_create_dialog(&block)
+      Yast::Wizard.CreateDialog
+      block.call
+    ensure
+      Yast::Wizard.CloseDialog
+    end
+
+    def run_assuming_open
```
jreidinger: I really do not like this name. Also why it is public method? and why it missing documentation if it is intention to be a public method?

I suggest something like `def show`

library/cwm/src/lib/cwm/dialog.rb
```diff
@@ -0,0 +1,97 @@
...
+    # @return [Symbol]
+    def run
+      if should_open_dialog?
+        wizard_create_dialog { run_assuming_open }
+      else
+        run_assuming_open
+      end
+    end
+
+    def wizard_create_dialog(&block)
```
jreidinger: missing documentation, also why it is public method?

library/cwm/src/lib/cwm/dialog.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    # The :next button
+    # @return [String,true,nil] button label,
+    #   `true` to use the default label, or `nil` to omit the button
+    def next_button
+      true
+    end
+
+    # @return [Array<Symbol>]
```
jreidinger: at least link to cwm documentation for it

jreidinger: and in fact, it doesn't have to be onlu symbol, it is in fact `Array<any non nil object>`

library/general/src/lib/ui/greasemonkey.rb
```diff
@@ -0,0 +1,225 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact Novell, Inc.
+#
+# To contact Novell about this file by physical or electronic mail, you may
+# find current contact information at www.novell.com.
+
+# File:	Greasemonkey.ycp
+# Package:	yast2-storage
+# Summary:	Expert Partitioner
```
jreidinger: these three lines are no longer valid.

library/general/src/lib/ui/greasemonkey.rb
```diff
@@ -0,0 +1,225 @@
...
+# Authors:	Arvin Schnell <aschnell@suse.de>
+#
+# Lets see if this turns out to be useful.
```
jreidinger: ?

library/general/src/lib/ui/greasemonkey.rb
```diff
@@ -0,0 +1,225 @@
...
+require "yast"
+
+module UI
+  # UI layout helpers.
```
jreidinger: this need really more extensive documentation

library/general/test/greasemonkey_test.rb
jreidinger: also what about adding `describe ".#{mname}" do` around it?

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -0,0 +1,71 @@
+require "yast"
+Yast.import "Sequencer"
+
+# FIXME: once the API is reviewed, move this to yast-yast2
```
jreidinger: no longer valid

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -0,0 +1,71 @@
...
+    # {Yast::SequencerClass#Run Yast::Sequencer.Run}
+    def self.run(aliases, sequence)
+      Yast::Sequencer.Run(aliases, sequence)
+    end
+
+    # A replacement for
+    # {Yast::SequencerClass#Run Yast::Sequencer.Run}
+    # but smarter:
+    # - auto :abort (see {#abortable})
+    # - *aliases* are assumed to be method names if unpecified
```
jreidinger: typo *unpecified*

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -0,0 +1,71 @@
...
+    #   (see {#from_methods})
+    def run(aliases: nil, sequence:)
+      aliases = from_methods(sequence) if aliases.nil?
```
jreidinger: can be written as `aliases ||= from_methods(sequence)`

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -0,0 +1,71 @@
...
+      self.class.run(aliases, abortable(sequence))
+    end
+
+    # Add !{abort: :abort} transitions if missing
+    # (an :abort from a dialog should :abort the whole sequence)
+    def abortable(sequence_hash)
+      sequence_hash.map do |name, destination|
+        if name == "ws_start"
```
jreidinger: this "hack" should be documented why it is needed. Also it smells for me like constant.

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -0,0 +1,71 @@
...
+      end.compact.to_h
+    end
+
+    class << self
+      # Declare that a method is skipped when going :back,
+      # useful for noninteractive steps.
+      # (also see Yast::SequencerClass#WS_special)
+      def skip_stack(method_symbol)
+        @skip_stack ||= {}
+        @skip_stack[method_symbol] = true
```
jreidinger: to be honest I do not get why it is class method and not instance one as I expect that skipped methods are single run specific, not? so what happen if two or three Sequence is used in row?

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+
+      def skip_stack?(name)
```
jreidinger: documentation

library/cwm/src/lib/cwm/dialog.rb
mvidner: 1. OK.
2. I think you are wrong about this: https://github.com/yast/yast-yast2/blob/15350bbdd7bda71215caa24b114d027cb6b47e80/library/cwm/src/modules/CWM.rb#L877 enforces a Symbol.

jreidinger: well, in fact I do not test it. But from past I know that any object stop during handle and convert to symbol looks happen only after test for skip for store https://github.com/yast/yast-yast2/blob/15350bbdd7bda71215caa24b114d027cb6b47e80/library/cwm/src/modules/CWM.rb#L874

mvidner: Parse error.
Anyway, a non-symbol means `store` may run, so the dialog "works", but `nil` is returned from `show` and I think such usage is wrong and should be fixed.

jreidinger: on tricky value is string, which is converted, so if handle return `"test"` skip for have to have `"test"`, but return value is `:test`. But I agree that we should not support it.

library/general/src/lib/ui/greasemonkey.rb
mvidner: As for test coverage, I can only offer 100% ;-)

As for the API: I *need* it to stay compatible, because that is the point: reusing pieces of the old dialogs while I concentrate on other things. Actually, in https://github.com/yast/yast-partitioner/pull/5 I used this but then in https://github.com/yast/yast-partitioner/pull/15 nothing is used as I gradually rewrote the dialogs. I expect several recurrences of this.

I will add docs.

library/general/test/greasemonkey_test.rb
mvidner: 1. OK
2. No, this is done by the `describe`s below.

library/general/src/lib/ui/greasemonkey.rb
jreidinger: well, when we add it to yast2, it is like we are suggesting to reuse it. So I think for this case functional API make sense as it is conversion of one kind of term to other, but I really like to have code that does not look like ycp :)

library/sequencer/src/lib/ui/sequence.rb
mvidner: It is not a hack, it is a standard part of the sequence hash that specifies the starting point.

Unfortunately the docs for Sequencer itself is ... bad.

jreidinger: so use constant with proper name like "INITIAL_DIALOG"?

library/sequencer/src/lib/ui/sequence.rb
mvidner: You have a point. IIUC you don't want FooSequence to affect BarSequence. I believe this will not happen, because I have used a class-level `@`attribute, not a `@@`class-attribute (which is shared across the hierarchy). I will add a test for it.

It is a class method because I need it at the class level. Actual usage:

```rb
    class AddPartition < UI::Sequence
      def preconditions
        if slots.empty?
          Yast::Popup.Error("...")
          return :back
        end
        :next
      end
      skip_stack :preconditions

      def type
        Dialogs::PartitionType.run(disk, @ptemplate, @slots)
      end

      def size
        Dialogs::PartitionSize.run(disk, @ptemplate, @slots)
      end
      ...

```

jreidinger: so it is class attribute in fact. Well, it make sense then, if it is subclass specific attribute, that is persistent accross same instances.

btw you can write it as `skip_stack def preconditions`

library/general/src/lib/ui/greasemonkey.rb
```diff
@@ -86,6 +86,12 @@ def VStackFrames(old)
     end
     module_function :VStackFrames
 
+    # @param old [Yast::Term]
+    # @return    [Yast::Term]
+    # @example
+    #   term(:FrameWithMarginBox, "Title", "arg1", "arg2")
+    #      ->
+    #   Frame("Title", MarginBox(1.45, 0.45, "arg1", "arg2"))
```
jreidinger: seen this usage, I think it make sense to mention it as old example and deprecated. I think prefered way to use should be

```
include GreaseMonkey

def contents
  VBox(
    FrameWithMarginBox("Title", "arg1", "arg2")
  end
end
```

and this should be a priority example for suggested usage. Or how do you see it?

library/sequencer/test/sequence_test.rb
```diff
@@ -65,5 +71,18 @@ def second
...
+      subj = UnrelatedTestSequence.new
+      seq = {
+        "ws_start" => "first",
+        "first"    => { next: :next }
+      }
+      wanted = {
+        "first" => [subj.method(:first), true]
+      }
+
+      expect(subj.from_methods(seq)).to eq(wanted)
```
jreidinger: I do not get this test. I expect it should test if TestSequence do not skip `"first"`, not? Here it test that class that explicitly skip it really skip it, which is not what "it" description mention

library/general/src/lib/ui/greasemonkey.rb
mvidner: That is out of scope for this PR. As I say in the module header,
>   These started out in the Expert Partitioner in yast2-storage.
The use case is reusing pieces of this legacy code in the new yast2-partitioner.
 That is why the API and the implementation look old.

jreidinger: if only use case is reusing code in yast2-partitioner, that why it is in yast2 where it should serve all modules? That is why I am more strict regarding API and other stuff, because one module scope it is ok, if there are some legacy stuff, but if we add it to yast2 itself, it means we have to be more careful in future when we want to improve it.

library/cwm/src/lib/cwm/dialog.rb
```diff
@@ -0,0 +1,104 @@
...
+    #   Events for which `store` won't be called, see {Yast::CWMClass#show}
+    def skip_store_for
```
jreidinger: question is for which actions it should be done. Currently only use case I see is :redraw. So maybe this should be redraw_event and also handle it in run? Just idea as @teclator currently solving something similar.

library/general/test/greasemonkey_test.rb
jreidinger: still missing 1. fixed. It should mention that output of greasemonkey method should be valid UI term.

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -0,0 +1,77 @@
+require "yast"
+Yast.import "Sequencer"
+
+module UI
+  # A {UI::Sequence} is an object-oriented interface for the good old
+  # {Yast::SequencerClass Yast::Sequencer}.
+  # In the simple case it runs a sequence of dialogs
+  # connected by Back and Next buttons.
```
jreidinger: as sequencer can be a bit tricky, I really welcome here one or two trivial examples of usage

PR #595, opened 2017-07-10
Improve cwm dialog

jreidinger:


package/yast2.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jul 10 08:09:15 UTC 2017 - jreidinger@suse.com
+
+- Fix ommiting value in CWM::Dialog (boo#1039901)
```
mvidner: - "omitting"
- please make it understandable to the reader: *which* value?

library/cwm/src/lib/cwm/dialog.rb
```diff
@@ -86,16 +86,18 @@ def cwm_show
...
       )
     end
 
-    def replace_true(value, replacement)
-      if value == true
-        replacement
+    def convert_value(value, default)
+      case value
+      when true then default
+      # in CWM empty string means remove button, nil means use default, which is here default value
```
mvidner: Parse error. But maybe I will understand if we actually document the underlying API. Currently it just says
>  [back_button (String) — label for dialog back button](http://www.rubydoc.info/github/yast/yast-yast2/Yast/CWMClass#show-instance_method)

package/yast2.changes
jreidinger: right...it should be "fix omitting button", is this enough understandable?

library/cwm/src/lib/cwm/dialog.rb
jreidinger: I want to explain that CWM::Dialog use nil as "omit" button, but CWM module use "nil" as default value and `""` as omit button.

mvidner: s/use/uses/g !
OK, then please add this to the API docs of CWM.show
If I had known this, I would have made the API the same. Maybe we could still change it, with all existing uses.

mvidner:
Please also remove the `true` option from library/cwm/src/lib/cwm/rspec.rb

jreidinger:
@mvidner thanks for catch. Fixed.

PR #596, opened 2017-07-12
Ruby24

jreidinger:


library/systemd/src/lib/yast2/systemctl.rb
```diff
@@ -23,7 +23,9 @@ class << self
       def execute(command)
         command = SYSTEMCTL + command
         log.debug "Executing `systemctl` command: #{command}"
-        result = timeout(TIMEOUT) { SCR.Execute(Path.new(".target.bash_output"), command) }
+        result = ::Timeout.timeout(TIMEOUT) do
```
lslezak: Do we need full qualification here? `rescue Timeout::Error` below does not use it, it should be consistent.

jreidinger: after latest failure in ruby-bindings I start to be more paranoic, so I use it almost everywhere where it make sense. So I add it also below.

PR #598, opened 2017-07-18
Extend workflow package

jreidinger:
goal of this change is to allow to have installation.xml also in RPM directly.  Previously only addon can be used where it is expected that 1 repo is 1 addon which have one product and one rpm with installation.xml, so now it is more straightforward.


library/control/src/modules/WorkflowManager.rb
```diff
@@ -452,21 +464,27 @@ def addon_control_dir(src_id, cleanup: false)
     # Returns requested control filename. Parameter 'name' is ignored
     # for Add-Ons.
     #
-    # @param [Symbol] type `addon or `pattern
+    # @param [Symbol] type :addon or :pattern or :package, pattern not implemented
```
lslezak: Remove `:pattern` from `:addon or :pattern or :package`, it's strange to name it and later write it's not supported.

Plus mention returning `nil` in that case. (Should be also added to `@return`.)

library/control/src/modules/WorkflowManager.rb
```diff
@@ -477,25 +495,24 @@ def GetCachedWorkflowFilename(type, src_id, _name)
...
           # provides, this works with all repository types, including the RPM-MD repositories.
-          use_filename ||= addon_control_file(src_id)
-
-          # File exists?
-          return use_filename.nil? ? nil : StoreWorkflowFile(use_filename, disk_filename)
+          use_filename ||= control_file(src_id)
+        when :package
+          use_filename = control_file(name)
+        else
+          raise ArgumentError, "invalid argument type #{type.inspect}"
```
lslezak: This default branch should be actually never called, there is an argument check at the beginning. So either remove it or improve the error message so it's clear that the previous check is broken.

PR #600, opened 2017-07-20
Fix object wrapper usage

jreidinger:


library/cwm/test/wrapper_widget_test.rb
```diff
@@ -8,7 +8,12 @@
   describe "#cwm_definition" do
     it "returns passed hash content" do
       content = { "test" => "test" }
-      expect(described_class.new(content).cwm_definition).to eq content
+      expect(described_class.new(content).cwm_definition["test"]).to eq "test"
+    end
+
+    it "return hash that has _cwm_key key with widget id" do
+      content = { "test" => "test" }
```
mvidner: "test -> test" is bad test data, as it can be fooled by stupid mistakes in the implementation (`map{ |k, v| [k, k] }`)
so  "key" -> "value" would be better

PR #601, opened 2017-07-21
Additional "CWM::RadioButtons" tests

lslezak:
- No version change, for using it at Travis commit to YaST:Head is enough

Originally I wanted to put the `if subject.respond_to?(method)` check before the `describe` block so RSpec does not print bogus test results when the method is not defined. But that resulted in a `#let or #subject called without a block` error, so it seems it cannot be improved in that way.


lslezak:
Oh, the coverage has decreased... Do we need to write tests for tests? :smiley:

PR #602, opened 2017-07-27
Point users of CWMServiceStart to the replacement API.

mvidner:
Lingshan just asked about CWMServiceStart and it took me a while to find the modern API, so here's a shortcut.

### DIFF
--- a/library/cwm/src/modules/CWMServiceStart.rb
+++ b/library/cwm/src/modules/CWMServiceStart.rb
@@ -21,16 +21,13 @@
 # you may find current contact information at www.novell.com
 #
 # ***************************************************************************
-# File:	modules/CWMServiceStart.ycp
-# Package:	Common widget manipulation, service start widget
-# Summary:	Routines for service start widget handling
-# Authors:	Jiri Srain <jsrain@suse.cz>
-#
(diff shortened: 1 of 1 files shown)


library/cwm/src/modules/CWMServiceStart.rb
```diff
@@ -21,16 +21,13 @@
...
-#
-# $Id$
-#
+
 require "yast"
 
 module Yast
+  # Routines for service start widget handling.
+  #
+  # NOTE: a modern API for this is {::UI::ServiceStatus}
```
jreidinger: what about marking it as deprecated and replaced by that?

mvidner: Well, I wasn't confident about that, but since that time I had to look into systemctl anyway, so I'll check again and add more docs about how to migrate to the new API.

teclator:
@mvidner @jreidinger some months have passed since last comments, so what about merging current changes until we are able to mark it as deprecated and add more doc about how to migrate?

teclator:
@jreidinger was agree on IRC. I have deleted the project from our internal Jenkins.

PR #607, opened 2017-08-04
Force using Term::ReadLine::Perl library (bsc#966042)

lslezak:
- Use `Term::ReadLine::Perl` instead of the broken `Term::ReadLine::Gnu` default to avoid ugly error message printed in command line mode.
- The bug has been reported more than one year ago upstream: https://rt.cpan.org/Public/Bug/Display.html?id=110943
- But as it has not been fixed since that, so let's apply the workaround in YaST
- See more details in https://bugzilla.suse.com/show_bug.cgi?id=966042#c4
- 3.3.8

### DIFF
--- a/scripts/yast2
+++ b/scripts/yast2
@@ -35,6 +35,10 @@ export QT_IM_MODULE=xim GTK_IM_MODULE=xim
 # used correctly (see bnc#866692 for details).
 export PX_MODULE_PATH=""
 
+# Force using Perl readline library instead of the broken default Gnu variant
+# to avoid ugly error message printed in command line mode (bsc#966042)
+export PERL_RL=Perl
+
 # allow for a different prefix
 # strip the basename off $0, which can be: (bnc#382216, bnc#458385)
 # /sbin/yast2, /sbin/yast, yast2 (sh -x yast2 ...), /sbin/yast2 (PATH=/sbin/:...)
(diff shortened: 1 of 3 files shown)


scripts/yast2
```diff
@@ -35,6 +35,10 @@ export QT_IM_MODULE=xim GTK_IM_MODULE=xim
 # used correctly (see bnc#866692 for details).
 export PX_MODULE_PATH=""
 
+# Force using Perl readline library instead of the broken default Gnu variant
+# to avoid ugly error message printed in command line mode (bsc#966042)
+export PERL_RL=Perl
```
jreidinger: is it by default there or does it need some dependency?

lslezak: Aghh, good point!

Debugging revealed that it's not installed by default and `Term::ReadLine::Stub` is silently used instead. That means there is actually no readline used - the command line works but there is no command history in the interactive mode. 

So there are these possibilities:

- Keep the broken default as it is (drop this patch)
- Modify the patch to keep the user defined `$PERL_RL` value (set it only if not defined). Users who do not like the new default (and want the history) could either override it to `Gnu` value back or install the `Term::ReadLine::Perl` library manually.

We should probably discuss it with @mvidner next week...

jreidinger: agreed to discuss it with @mvidner

mvidner: Thanks for the catch! But then your proposed ideas did not really improve the situation, so I tried to look into the underlying package, and found that updating to the latest upstream version fixes it: https://bugzilla.suse.com/show_bug.cgi?id=966042#c7

mvidner:
Found that updating perl-Term-ReadLine-Gnu will fix this instead.

PR #613, opened 2017-08-15
Speed up service manager

mvidner:
References
- https://trello.com/c/IiJYLiBk/1110-3-speedup-reading-services-state-services-manager 
- [bsc#1045658](https://bugzilla.suse.com/show_bug.cgi?id=1045658)

Changes
- add `SystemdService.find_many` for a speed-up
- `SystemdUnit` fix for units with multiple dots in name
- consistent logging of `systemctl` calls
- better Yardoc

### DIFF
--- a/library/systemd/src/lib/yast2/systemctl.rb
+++ b/library/systemd/src/lib/yast2/systemctl.rb
@@ -27,6 +27,7 @@ module Yast
       # @return [#command,#stdout,#stderr,#exit]
       # @raise [SystemctlError] if it times out
       def execute(command)
+        log.info("systemctl #{command}")
         command = SYSTEMCTL + command
         log.debug "Executing `systemctl` command: #{command}"
         result = timeout(TIMEOUT) { SCR.Execute(Path.new(".target.bash_output"), command) }
--- a/library/systemd/src/lib/yast2/systemctl.rb
+++ b/library/systemd/src/lib/yast2/systemctl.rb
@@ -4,12 +4,15 @@ require "timeout"
 module Yast
   # Exception when systemctl command failed
   class SystemctlError < StandardError
-    def initialize(struct)
-      super "Systemctl command failed: #{struct}"
+    # @param details [#to_s]
+    def initialize(details)
+      super "Systemctl command failed: #{details}"
     end
   end
 
-  # Wrapper around systemctl command
+  # Wrapper around `systemctl` command.
+  # - uses non-interactive flags
+  # - has a timeout
   module Systemctl
     include Yast::Logger
 
@@ -20,6 +23,9 @@ module Yast
     TIMEOUT         = 30 # seconds
 
     class << self
+      # @param command [String]
+      # @return [#command,#stdout,#stderr,#exit]
+      # @raise [SystemctlError] if it times out
       def execute(command)
         command = SYSTEMCTL + command
         log.debug "Executing `systemctl` command: #{command}"
@@ -29,6 +35,7 @@ module Yast
         raise SystemctlError, "Timeout #{TIMEOUT} seconds: #{command}"
       end
 
+      # @return [Array<String>] like ["a.socket", "b.socket"]
       def socket_units
         sockets_from_files = list_unit_files(type: :socket).lines.map do |line|
           first_column(line)
@@ -41,6 +48,7 @@ module Yast
         (sockets_from_files | sockets_from_units).compact
       end
 
+      # @return [Array<String>] like ["a.service", "b.service"]
       def service_units
         services_from_files = list_unit_files(type: :service).lines.map do |line|
           first_column(line)
@@ -53,6 +61,7 @@ module Yast
         (services_from_files | services_from_units).compact
       end
 
+      # @return [Array<String>] like ["a.target", "b.target"]
       def target_units
         targets_from_files = list_unit_files(type: :target).lines.map do |line|
           first_column(line)
--- a/library/systemd/src/lib/yast2/systemctl.rb
+++ b/library/systemd/src/lib/yast2/systemctl.rb
@@ -23,6 +23,8 @@ module Yast
     TIMEOUT         = 30 # seconds
 
     class << self
+      BASH_SCR_PATH = Yast::Path.new(".target.bash_output")
+
       # @param command [String]
       # @return [#command,#stdout,#stderr,#exit]
       # @raise [SystemctlError] if it times out
@@ -30,7 +32,7 @@ module Yast
         log.info("systemctl #{command}")
         command = SYSTEMCTL + command
         log.debug "Executing `systemctl` command: #{command}"
-        result = timeout(TIMEOUT) { SCR.Execute(Path.new(".target.bash_output"), command) }
+        result = timeout(TIMEOUT) { SCR.Execute(BASH_SCR_PATH, command) }
         OpenStruct.new(result.merge!(command: command))
       rescue Timeout::Error
         raise SystemctlError, "Timeout #{TIMEOUT} seconds: #{command}"
--- a/library/systemd/src/lib/yast2/systemd_unit.rb
+++ b/library/systemd/src/lib/yast2/systemd_unit.rb
@@ -83,7 +83,9 @@ module Yast
       @propmap = propmap.merge!(DEFAULT_PROPMAP)
 
       @properties = show
+      # eg "Failed to get properties: Unit name apache2@.service is not valid."
       @error = properties.error
+      # Id is not present when the unit name is not valid
       @name = id.to_s.split(".").first || unit_name
     end
 
--- a/library/systemd/src/lib/yast2/systemd_unit.rb
+++ b/library/systemd/src/lib/yast2/systemd_unit.rb
@@ -225,10 +225,8 @@ module Yast
       end
 
       def load_systemd_properties
-        properties = systemd_unit.propmap.map do |_, property_name|
-          " --property=#{property_name} "
-        end
-        systemd_unit.command("show", options: properties.join)
+        names = systemd_unit.propmap.values
+        systemd_unit.command("show", options: "--property=" + names.join(","))
       end
     end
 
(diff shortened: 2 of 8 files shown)


library/systemd/src/lib/yast2/systemctl.rb
```diff
@@ -20,15 +24,34 @@ module Systemctl
...
         log.debug "Executing `systemctl` command: #{command}"
-        result = timeout(TIMEOUT) { SCR.Execute(Path.new(".target.bash_output"), command) }
+        result = timeout(TIMEOUT) { SCR.Execute(BASH_SCR_PATH, command) }
         OpenStruct.new(result.merge!(command: command))
       rescue Timeout::Error
         raise SystemctlError, "Timeout #{TIMEOUT} seconds: #{command}"
       end
 
+      # Redefining a method is more friendly to git rebasing.
+      # FIXME: remove this before merging
```
jreidinger: :wink: and same for require

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -37,7 +37,23 @@ class SystemdUnit
...
     SUPPORTED_STATES = %w(enabled disabled).freeze
 
-    DEFAULT_PROPERTIES = {
+    # Values of {#active_state} fow which we consider a unit "active".
+    #
+    # systemctl.c:check_unit_active uses (active, reloading)
+    # For bsc#884756 we also consider "activating" to be active.
+    #
+    # (The remaining states are "deactivating", "inactive", "failed".)
+    ACTIVE_STATES = ["active", "activating", "reloading"].freeze
```
jreidinger: that is a bit problematic, as systemd can add more states and we had such problem in past, that is why we call explicit systemd command is_active that move responsibility to systemd to say if it is active.

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -52,17 +68,32 @@ class SystemdUnit
...
+      raise "Missing unit type suffix" if dot.empty?
 
       log.warn "Unsupported unit type '#{unit_type}'" unless SUPPORTED_TYPES.include?(unit_type)
-      @input_properties = properties.merge!(DEFAULT_PROPERTIES)
+      @propmap = propmap.merge!(DEFAULT_PROPMAP)
 
-      @properties = show
-      @error = self.properties.error
+      @properties = show(property_text)
+      # eg "Failed to get properties: Unit name apache2@.service is not valid."
```
jreidinger: maybe add such comment to error attr documentation?

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -72,9 +103,11 @@ def refresh!
       properties
     end
 
-    def show
+    # Run 'systemctl show' to read the unit properties
+    # @return [Properties]
```
jreidinger: document what is property_text parameter

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -137,13 +174,20 @@ def run_command!
     class Properties < OpenStruct
       include Yast::Logger
 
-      def initialize(systemd_unit)
+      # @param systemd_unit [SystemdUnit]
```
jreidinger: missing doc for param property text

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -222,7 +264,7 @@ def load_systemd_properties
     class InstallationProperties < OpenStruct
       include Yast::Logger
 
-      def initialize(systemd_unit)
+      def initialize(systemd_unit, _property_text)
```
jreidinger: so if I get it, it is ignored in installation and always readed even when cached property text is passed. It is a bit surprising behavior.

library/systemd/src/modules/systemd_service.rb
```diff
@@ -62,32 +65,64 @@ module Yast
...
-
   class SystemdServiceClass < Module
+    include Yast::Logger
+
     UNIT_SUFFIX = ".service".freeze
 
-    def find(service_name, properties = {})
+    # @param service_name [String] "foo" or "foo.service"
+    # @param propmap [SystemdUnit::PropMap]
+    # @return [Service,nil]
```
jreidinger: document when it return nil

library/systemd/src/modules/systemd_service.rb
```diff
@@ -62,32 +65,64 @@ module Yast
...
+        service
+      end
+    end
+
+    # @param service_names [Array<String>] "foo" or "foo.service"
+    # @param propmap [SystemdUnit::PropMap]
+    # @return [Array<Service,nil>]
+    def find_many(service_names, propmap = {})
+      find_many!(service_names, propmap)
+    rescue SystemdServiceNotFound
```
jreidinger: to be honest I prefer other way around like in find, where it tries without exceptions and only if it is ! variant and nil is there, it raise exception.

library/systemd/src/modules/systemd_service.rb
```diff
@@ -62,32 +65,64 @@ module Yast
...
+      snames_s = snames.join(" ")
+      pnames_s = SystemdUnit::DEFAULT_PROPMAP.merge(propmap).values.join(",")
+      out = Systemctl.execute("show  --property=#{pnames_s} #{snames_s}")
+      log.error "returned #{out.exit}, #{out.stderr}" unless out.exit.zero? && out.stderr.empty?
+      property_texts = out.stdout.split("\n\n")
+      raise(SystemdServiceNotFound, "?") unless snames.size == property_texts.size
+
+      snames.zip(property_texts).map do |service_name, property_text|
+        service = Service.new(service_name, propmap, property_text)
+        next nil if service.properties.not_found?
```
jreidinger: this contradicts documentation, which does not mention nil in array.

library/systemd/src/lib/yast2/systemd_unit.rb
mvidner: > that is why we call explicit systemd command is_active

And that is why our module was so f**king slow. Should I add a comment like this?
> 10 to 50ms per call of `systemctl is-active` * 100 to 300 services (depending on hardware and software installed, VM or not) is 1 to 15 seconds delay. That is why we try really hard to avoid many systemctl calls.

library/systemd/src/modules/systemd_service.rb
mvidner: 1. From the outside it is consistent: with the `!` it may `raise`, otherwise not
2. It is a subtle point that I need to add a comment about: we do not know WHICH service was not found. One of them made `systemctl show` return incomplete data but it is hard to match correctly.

library/systemd/src/lib/yast2/systemd_unit.rb
mvidner: Damn, it seems like I missed this and this will break during the installation :(

mvidner: Fixed

library/systemd/src/modules/systemd_service.rb
mvidner: But you're right, it is confusing, and I don't need an exception for this. Used a condition instead.

library/systemd/src/modules/systemd_service.rb
mvidner: Fixed

library/systemd/src/lib/yast2/systemd_unit.rb
mvidner: Added

library/systemd/src/lib/yast2/systemd_unit.rb
jreidinger: yes, comment would be great to explain it. Also what I have in mind is explicit check that it is known state ( written only in comment now ) and if not, at least log error ( exception is maybe too strong, but who knows ).

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -191,16 +242,14 @@ def state_name_enabled?(state)
...
         end
       end
 
       def load_systemd_properties
-        properties = systemd_unit.input_properties.map do |_, property_name|
-          " --property=#{property_name} "
-        end
-        systemd_unit.command("show", options: properties.join)
+        names = systemd_unit.propmap.values
+        systemd_unit.command("show", options: "--property=" + names.join(","))
```
jreidinger: just to be sure, how name with whitespace is handled? or it cannot happen?

library/systemd/src/modules/systemd_service.rb
```diff
@@ -62,32 +65,68 @@ module Yast
...
+    # @param propmap [SystemdUnit::PropMap]
+    # @return [Service]
+    # @raise [SystemdServiceNotFound]
+    def find!(service_name, propmap = {})
+      find(service_name, propmap) || raise(SystemdServiceNotFound, service_name)
+    end
+
+    # @param service_names [Array<String>] "foo" or "foo.service"
+    # @param propmap [SystemdUnit::PropMap]
+    # @return [Array<Service,nil>] `nil` if not found, [] if no can do
```
jreidinger: what means `[] if no can do` from code it looks like it is [] if it cannot be obtain for initial or if some property name cannot be found

library/systemd/src/modules/systemd_service.rb
```diff
@@ -62,32 +65,68 @@ module Yast
...
+    # @raise [SystemdServiceNotFound] if an unexpected problem occurs
+    def find_many_at_once(service_names, propmap = {})
```
jreidinger: btw does it make sense to have it public? or it is just private speed up method?

library/systemd/src/lib/yast2/systemd_unit.rb
mvidner: No, property names do not have whitespace.

library/systemd/src/lib/yast2/systemd_unit.rb
mvidner: Added

PR #615, opened 2017-08-25
Restore Product and Repository classes

imobachgs:
See [bsc#1055677](https://bugzilla.opensuse.org/show_bug.cgi?id=1055677).


ancorgs:
Makes sense.

PR #616, opened 2017-08-27
Fall back to smaller and/or hicolor icon if the icon can't be found.

Hacklin:
Themed YaST module icons may not be available in 64x64.
The default YaST app. icon is not in the theme but in hicolor.

Possible fix for boo#902067 and boo#909730.

### DIFF
--- a/library/wizard/src/modules/Wizard.rb
+++ b/library/wizard/src/modules/Wizard.rb
@@ -32,6 +32,10 @@ require "yast"
 
 module Yast
   class WizardClass < Module
+
+    # Default icon name to use.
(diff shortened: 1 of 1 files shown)


jreidinger:
@teclator looks like that also libyui somehow support using fallback icon path. https://github.com/libyui/libyui/blob/master/src/YIconLoader.cc#L40
So maybe we should check if it cannot be used there.

teclator:
@jreidinger good point, I'm updating the PR

teclator:
Replaced by #725 

Thnx for the contribution!, I have created the other one for moving forward with the review.

PR #619, opened 2017-08-31
Merge workflow and add support to disable buttons on CWM::Dialog

imobachgs:
This PR add support to:

* Merge product's workflow (moving the logic from [ProductSelector widget](https://github.com/yast/yast-installation/blob/master/src/lib/installation/dialogs/product_selection.rb#L45).
* Extends CWM::Dialog to support disabling buttons.


package/yast2.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 31 15:30:24 UTC 2017 - igonzalezsosa@suse.com
+
+- Add support to disable buttons on CWM::Dialog
+- Add a method WorkflowManager#merge_product_workflow which allows
+  to merge a product workflow (fate#322267)
```
ancorgs: Missing version number

PR #620, opened 2017-09-04
Snapper config

ancorgs:
Part of https://trello.com/c/ytK5Yoe4/554-5-storageng-installation-to-root-subvolume

It's needed by other PRs referenced there.

The first commit looks big but it only moves code to avoid tons of `def self.xxx` and `:private_class_method`. So better review in a commit basis.

### DIFF
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -343,7 +343,7 @@ module Yast2
 
       # Executes the fourth step of the installation-helper of Snapper.
       #
-      # Unfortunatelly the steps of the Snapper helper are not much descriptive.
+      # Unfortunately the steps of the Snapper helper are not much descriptive.
       # The step 4 must be executed in the target system after installing the
       # packages and before using snapper for the first time.
       def installation_helper_step_4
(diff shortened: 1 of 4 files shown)


library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -37,6 +38,15 @@ def initialize
     end
   end
 
+  # Represents that a Snapper configuration was attempted at the wrong time or
+  # system, since it's only possible in a fresh installation after software
+  # installation.
+  class SnapperNotConfigurable < StandardError
+    def initialize
+      super "Programming error: Snapper cannot be configured in this point."
```
shundhammer: **at** this point

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -298,6 +340,26 @@ def target_root
 
         Yast::Installation.destdir
       end
+
+      def installation_helper_step_4
```
jreidinger: is possible to use better name? we should try to abstract it a bit.

ancorgs: Well, this is literally calling the step 4 of Snapper... which doesn't have a meaningful name on its own.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -341,6 +341,11 @@ def target_root
         Yast::Installation.destdir
       end
 
+      # Executes the fourth step of the installation-helper of Snapper.
+      #
+      # Unfortunatelly the steps of the Snapper helper are not much descriptive.
```
shundhammer: Typo: Unfortunately

PR #621, opened 2017-09-06
Make docs about Travis+Docker discoverable

mvidner:
WIP, Tackling https://trello.com/c/CyuJDYsJ/338-dedicated-yast-integrator

I am breaking the usual rule of "comment the WHY but not the WHAT" because the Docker CI mechanism is not a "core competency"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ services:
   - docker
 
 before_install:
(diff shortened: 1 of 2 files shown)


mvidner:
@lslezak does this approach make sense? Would you add something?

PR #622, opened 2017-09-07
Handle GPG check during installation

imobachgs:
This simple fix should solve [bsc#1054663](https://bugzilla.opensuse.org/show_bug.cgi?id=1054663). This callback is only called when validation fails and it should be ignored if Insecure=1.

Finally, it does not share code with [AutoYaST counterpart](https://github.com/yast/yast-autoinstallation/blob/676f7f03c907d212c483731651f1e5edfed041b5/src/modules/AutoInstall.rb#L326) because requirements are totally different.

### DIFF
--- a/library/packages/src/modules/PackageCallbacks.rb
+++ b/library/packages/src/modules/PackageCallbacks.rb
@@ -2899,6 +2899,7 @@ module Yast
       Pkg.CallbackStartPackage(nil)
       Pkg.CallbackProgressPackage(nil)
       Pkg.CallbackDonePackage(nil)
+      Pkg.CallbackPkgGpgCheck(nil)
 
       nil
     end
--- a/library/packages/src/modules/PackageCallbacks.rb
+++ b/library/packages/src/modules/PackageCallbacks.rb
@@ -417,8 +417,8 @@ module Yast
     #   a blank string is returned (so no decision is made).
     def pkg_gpg_check(data)
       log.debug("pkgGpgCheck data: #{data}")
-      return "I" if Stage.initial && Linuxrc.InstallInf("Insecure") == "1"
-      ""
+      insecure = Stage.initial && Linuxrc.InstallInf("Insecure") == "1"
+      insecure ? "I" : ""
     end
 
     #  After package install.
--- a/library/packages/src/modules/PackageCallbacks.rb
+++ b/library/packages/src/modules/PackageCallbacks.rb
@@ -406,6 +406,7 @@ module Yast
     # Handle GPG check result (pkgGpgCheck)
     #
     # If insecure mode is set to '1', the check result is ignored. Otherwise, no decision is made.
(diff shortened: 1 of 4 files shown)


library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -402,6 +403,22 @@ def ProgressPackage(percent)
...
+    # @param data [Hash] Output from `pkgGpgCheck` callback.
+    # @option data [Integer] "CheckPackageResult" Check result code according to libzypp.
+    # @option data [String]  "Package" Package's name.
+    # @option data [String]  "Localpath" Path to RPM file.
+    # @option data [String]  "RepoMediaUrl" Media URL.
+    # @return [String] "I" if the package should be accepted; otherwise
+    #   a blank string is returned (so no decision is made).
+    def pkg_gpg_check(data)
+      log.debug("pkgGpgCheck data: #{data}")
+      Linuxrc.InstallInf("Insecure") == "1" ? "I" : ""
```
lslezak: To avoid an [ugly error message](https://github.com/yast/yast-yast2/blob/c84f544a93f3706a00ad5d4cf841ab68fee8550f/library/general/src/modules/Linuxrc.rb#L55) I'd use
```ruby
(Stage.initial && Linuxrc.InstallInf("Insecure") == "1")
```

condition here.

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -2664,6 +2681,9 @@ def SetProvideCallbacks
       Pkg.CallbackDonePackage(
         fun_ref(method(:DonePackage), "string (integer, string)")
       )
+      Pkg.CallbackPkgGpgCheck(
+        fun_ref(method(:pkg_gpg_check), "string(map)")
```
lslezak: Please add also the reset call to the `RestoreProvideCallbacks` method.

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -416,7 +417,8 @@ def ProgressPackage(percent)
     #   a blank string is returned (so no decision is made).
     def pkg_gpg_check(data)
       log.debug("pkgGpgCheck data: #{data}")
-      Linuxrc.InstallInf("Insecure") == "1" ? "I" : ""
+      return "I" if Stage.initial && Linuxrc.InstallInf("Insecure") == "1"
```
lslezak: Um, IMHO that's difficult to read... what about something like

```ruby
insecure = Stage.initial && Linuxrc.InstallInf("Insecure") == "1"
insecure ? "I" : ""
```
???

BTW Rubocop is OK with that :wink:

imobachgs:
[(3) [SLE 15] (P1) INSECURE=1 should work](https://trello.com/c/xr8TGP9y/1135-3-sle-15-p1-insecure1-should-work)

PR #624, opened 2017-09-14
Revert to using separate property options to keep old tests working.

mvidner:
yast2-dhcp-server and yast2-samba-server log the exact command used by
us in their tests :-(

### DIFF
--- a/library/systemd/src/lib/yast2/systemd_unit.rb
+++ b/library/systemd/src/lib/yast2/systemd_unit.rb
@@ -255,7 +255,8 @@ module Yast
 
       def load_systemd_properties
         names = systemd_unit.propmap.values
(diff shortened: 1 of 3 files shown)


mvidner:
Fixes up #613

jreidinger:
@mvidner do you measure if there is performance difference?

mvidner:
No. My educated guess is that the 50ms per call of systemctl is mostly spent on process startup and DBus communication. I HOPE that it doesn't try to query the properties 1 by 1 with separate DBus calls.

mvidner:
Well, I *do* remember that I did not see any improvement  after introducing the original change in this part. So, from memory, the performance stays the same.

PR #626, opened 2017-09-15
Fix running gnome center

jreidinger:
tested manually

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Sep 15 11:10:39 UTC 2017 - jreidinger@suse.com
+
+- fix starting gnome control center (bsc#1058376)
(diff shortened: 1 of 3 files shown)


Trucido:
Quoting my response to bug report after it was closed:

```
Thanks, I looked at the pull request too but we still may need a more future proof method. This issue also occurred in Leap. Just having this package installed broke launching yast gui from anywhere.

I still think it would be a good idea to have a yast2-control-center-gtk or something which this package could dep on, or have installed standalone for other desktops like XFCE and MATE. We'd also have a fallback if gnome breaks it again. 

One of my tumbleweed box updated to gnome 3.26, so I'll have to check if it works there with the fix. Gnome kind of removed the settings panel and replaced it with a new thing with sidebar navigation similar to gnome-tweak-tool
```

PR #628, opened 2017-09-28
[Review] Request from 'schubi2' @ 'yast/yast-yast2/review_170928_disable_remote_installation'

schubi2:
  * 560cb7d1 disable vnc, shh,... installation if it is not supported

### DIFF
--- a/library/general/src/modules/Linuxrc.rb
+++ b/library/general/src/modules/Linuxrc.rb
@@ -31,7 +31,6 @@ require "yast"
 
 module Yast
   class LinuxrcClass < Module
-
     include Yast::Logger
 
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Linuxrc.rb
```diff
@@ -239,6 +241,32 @@ def value_for(feature_key)
       ret
     end
 
+    # Reset settings for vnc, ssh,... in install.inf which have been made
+    # by linuxrc settings.
+    #
+    # @param [Array<String>] list of services which will be disabled.
+    def disable_remote(services)
```
lslezak: Could you please add an unit test for this? Thank you!

PR #630, opened 2017-10-06
SuSEFirewallClass: IsInstalled does not return true if Pkg is selected

teclator:
This is just an early release of a possible fix for https://bugzilla.novell.com/show_bug.cgi?id=1037214

- [Trello Card](https://trello.com/c/HN5J9P4x)

What I have seen by now is that SuSEFirewallIsInstalled() returns true not only if the package has been Installed but even if it has only be selected in initial Stage.

Currently Lan.Write checks if the FW is installed to proceed with fw configuration.

- https://github.com/yast/yast-network/blob/SLE-12-SP2/src/modules/Lan.rb#L603

```
## Code affected

 def Write
      ...
      fw_is_installed = SuSEFirewall4Network.IsInstalled

      if fw_is_installed
        ...
        SuSEFirewall4Network.Write
        ...
```

The network dialog also show you some FW options if it is already installed.

We could do a simple [workaround](https://github.com/yast/yast-network/pull/576) in yast2-network SuSEFirewall4Network module to explicitly check if it is just selected or really installed, but I preferred to split the current SuSEFirewall.IsInstalled method.

With the proposal fix, SuSEFirewallIsInstalled will really check that the `@FIREWALL_PACKAGE` is installed or will install and there is other method SuSEFirewallIsSelectedOrInstalled which have the old behavior.

## AutoYaST

I was checking if the firewall_stage1 client could be called in AutoYaST and how the proposal was shown but it was disabled for AutoYaST (https://bugzilla.suse.com/show_bug.cgi?id=894130) so no affected

### DIFF
--- a/library/network/src/lib/network/susefirewall.rb
+++ b/library/network/src/lib/network/susefirewall.rb
@@ -60,7 +60,7 @@ module Yast
     # @param [Boolean] start_service at Write() process
     # @see #GetStartService()
     def SetStartService(start_service)
-      if !SuSEFirewallIsInstalled()
+      if !SuSEFirewallIsSelectedOrInstalled()
         Builtins.y2warning("Cannot set SetStartService")
         return nil
       end
--- a/library/network/src/lib/network/susefirewall.rb
+++ b/library/network/src/lib/network/susefirewall.rb
@@ -537,11 +537,13 @@ module Yast
     #
     # @return [Boolean] whether the selected firewall backend is installed
     def SuSEFirewallIsSelectedOrInstalled
-      if Stage.initial
-        @needed_packages_selected = Pkg.IsSelected(@FIREWALL_PACKAGE)
-        log.info "Selected for installation -> #{@needed_packages_installed}"
+      return true if @needed_packages_installed
 
-        return true if @needed_packages_selected
+      if Stage.initial
+        packages_selected = Pkg.IsSelected(@FIREWALL_PACKAGE)
+        log.info "Selected for installation -> #{packages_selected}"
+
+        return true if packages_selected
       end
 
       SuSEFirewallIsInstalled()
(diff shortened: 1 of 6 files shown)


library/network/src/lib/network/susefirewall2.rb
```diff
@@ -2757,7 +2757,6 @@ def full_init_on_boot(new_state)
     publish function: :AddServiceSupportIntoZone, type: "void (string, string)", private: true
     publish variable: :check_and_install_package, type: "boolean", private: true
     publish function: :SetInstallPackagesIfMissing, type: "void (boolean)"
-    publish variable: :needed_packages_installed, type: "boolean", private: true
```
teclator: Published twice, see:

https://github.com/teclator/yast-yast2/blob/8bc71cd64e63699c40cabdda4c495ee791a4d32d/library/network/src/lib/network/susefirewall2.rb#L2845

library/network/src/lib/network/susefirewall.rb
```diff
@@ -536,21 +536,31 @@ def IsKnownZone(zone)
     # installation)
     #
     # @return [Boolean] whether the selected firewall backend is installed
-    def SuSEFirewallIsInstalled
-      # Always recheck the status in inst-sys, user/solver might have change
-      # the list of packages selected for installation
-      # bnc#892935: in inst_finish, the package is already installed
+    def SuSEFirewallIsSelectedOrInstalled
+      return true if @needed_packages_installed
```
mvidner: Why do we need a cache and why doe it only cache positive results?

teclator: This method does not cache anything but return true if the package is alrady installed. I suppose it was cached because IsInstalled is called multiple times, so once installed does not make sense to task PackageSystem each time.

In case of the selection, the user can change the package selection so not make sense.

library/network/src/lib/network/susefirewall.rb
```diff
@@ -536,21 +536,31 @@ def IsKnownZone(zone)
...
+
       if Stage.initial
-        @needed_packages_installed = Pkg.IsSelected(@FIREWALL_PACKAGE) || PackageSystem.Installed(@FIREWALL_PACKAGE)
-        log.info "Selected for installation/installed -> #{@needed_packages_installed}"
-      elsif @needed_packages_installed.nil?
```
teclator: It was already cached in the past.

PR #632, opened 2017-10-19
Fixing "disable_remote" function in Linuxrc

kobliha:
- bsc#1055279
- Arguments "vnc", "ssh" were expected lower-cased in the function, but the code uses them ALSO upper-cased
- I've fixed it in a way that it just does not matter
- Plus if the argument is unknown, error is raised
- Plus it's covered by unit tests

### DIFF
--- a/library/general/src/modules/Linuxrc.rb
+++ b/library/general/src/modules/Linuxrc.rb
@@ -244,12 +244,16 @@ module Yast
     # Reset settings for vnc, ssh,... in install.inf which have been made
     # by linuxrc settings.
     #
-    # @param [Array<String>] list of services which will be disabled.
+    # @param [Array<String>] list of remote-management services that will be disabled.
     def disable_remote(services)
       return if !services || services.empty?
-      log.warn "Disabling #{services} due missing packages."
+
+      log.warn "Disabling #{services} due to missing packages."
       services.each do |service|
-        case service
+        # Service IDs are also used in another context in the code
+        # Making sure we always compare apples with apples
+        case polish(service.dup)
+
         when "vnc"
           SCR.Write(path(".etc.install_inf.VNC"), 0)
           SCR.Write(path(".etc.install_inf.VNCPassword"), "")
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Linuxrc.rb
```diff
@@ -244,23 +244,28 @@ def value_for(feature_key)
...
     def disable_remote(services)
       return if !services || services.empty?
-      log.warn "Disabling #{services} due missing packages."
+
+      log.warn "Disabling #{services} due to missing packages."
       services.each do |service|
-        case service
+        # Service IDs are also used in another context in the code
+        # Making sure we always compare apples with apples
+        case polish(service.dup)
```
lslezak: The `dup` is not needed, the `downcase` in `key.downcase.tr("-_\\.", "")` (in the `polish` methos) already creates a copy of the string. See the documentation: https://ruby-doc.org/core-2.4.0/String.html#method-i-downcase

library/general/test/linuxrc_test.rb
```diff
@@ -237,31 +237,41 @@ def load_install_inf(defaults_replacement = {})
         .with(path(".etc.install_inf"), nil)
     end
 
-    context "when vnc will be disabled" do
+    context "when vnc should be disabled" do
       it "updates install.inf" do
         expect(Yast::SCR).to receive(:Write)
-          .with(path(".etc.install_inf.VNC"), 0)
+          .with(path(".etc.install_inf.VNC"), 0).twice
```
lslezak: It took me a while to realize why it is written twice, it looked wrong at the first sight.

Please add a comment that we are testing two services ("vnc" and "VNC"), the same below.

library/general/src/modules/Linuxrc.rb
jreidinger: this is actually not true. If service does not contain uppercase it is not copy. Try it yourself with

```
irb(main):001:0> a = "a"
=> "a"
irb(main):002:0> a === a.downcase
=> true
irb(main):003:0> a === a.upcase
=> false
```

lslezak: Oh, I see, that's quite tricky...

library/general/test/linuxrc_test.rb
kobliha: Okaj

PR #633, opened 2017-10-23
log calculated proposal in yaml format into separate files (fate#318196)

wfeldt:
Looking at y2log alone is not helpful. Log the devicegraph into
human-readable files.


teclator:
@wfeldt what is the state of this PR? still valid? or not make sense anymore?

PR #637, opened 2017-11-08
Small improvements

jreidinger:


library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -6,6 +6,73 @@ module UI
...
+  #       "step1" => {
+  #         next: "step2",
+  #       },
+  #       "step2" => {
+  #         next: :finish
+  #       }
+  #     }
+  #
+  #     def step1
+  #       :next
```
mvidner: What about adding `# ... show UI of this step ...` above the return value?

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -6,6 +6,73 @@ module UI
...
+  #       super(sequence: SEQUENCE_HASH)
+  #     end
+  #   end
+  #
+  # @example complex sequence with branching and auto steps that are skipped when going back
+  #   class ComplexSequence < UI::Sequence
+  #     SEQUENCE_HASH = {
+  #       START   => "step1",
+  #       "step1" => {
+  #         next: "auto_step",
```
mvidner: For readability I would collapse the next-only entries to a single line:
```rb
 "step1" => { next: "auto_step" },
```

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -24,13 +91,38 @@ def self.run(aliases, sequence)
     # - auto :abort (see {#abortable})
     # - *aliases* are assumed to be method names if unspecified
     #   (see {#from_methods})
+    #
+    # @example sequence with branching with child having methods step1 and step2, where step 1 can finish sequence or run step 2
+    #   run({
```
mvidner:
```rb
run(sequence: {
```

library/sequencer/src/lib/ui/sequence.rb
```diff
@@ -24,13 +91,38 @@ def self.run(aliases, sequence)
...
+    #     START => "step1",
+    #     "step1" => { next: "step2" },
+    #     "step2" => { next: :finish }
+    #   }
+    #   output = {
+    #     START => "step1",
+    #     "step1" => { next: "step2", abort: :abort },
+    #     "step2" => { next: :finish, abort: :abort }
+    #   }
+    #   input == output # => true
```
mvidner:
```rb
abortable(input) == output
```

library/sequencer/src/lib/ui/sequence.rb
jreidinger: well, when I thinking about this it does not need to be UI. In fact it is just some kind of StateMachine with defined transitions between states.

PR #638, opened 2017-11-09
Firewalld: Added change_interface api method and tests for firewalld 

teclator:
Better method naming and tests (also a litle issue calling Service.Enabled)

### DIFF
--- a/library/network/test/y2firewall/firewalld_test.rb
+++ b/library/network/test/y2firewall/firewalld_test.rb
@@ -30,15 +30,30 @@ describe Y2Firewall::Firewalld do
   let(:firewalld) { described_class.instance }
 
   describe "#installed?" do
+    it "returns false it the firewalld is not installed" do
+      allow(Yast::PackageSystem).to receive("Installed")
+        .with(described_class::PACKAGE).and_return(false)
+
+      expect(firewalld.installed?).to eq(false)
+    end
+
     it "returns true it the firewalld is installed" do
       allow(Yast::PackageSystem).to receive("Installed")
         .with(described_class::PACKAGE).and_return true
+
       expect(firewalld.installed?).to eq(true)
     end
   end
 
   describe "#enabled?" do
     it "returns true if the firewalld service is enable" do
+      allow(Yast::Service).to receive("Enabled")
+        .with(described_class::SERVICE).and_return(true)
+
+      expect(firewalld.enabled?).to eq(true)
+    end
+
+    it "returns false if the firewalld service is disable" do
       allow(Yast::Service).to receive("Enabled")
         .with(described_class::SERVICE).and_return(false)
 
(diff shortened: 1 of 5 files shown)


library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -0,0 +1,120 @@
...
+require "y2firewall/firewalld"
+
+Yast.import "PackageSystem"
+Yast.import "Service"
+
+describe Y2Firewall::Firewalld do
+  let(:firewalld) { described_class.instance }
+
+  describe "#installed?" do
+    it "returns true it the firewalld is installed" do
```
imobachgs: "true *if* ~the~ firewalld"

imobachgs: Just a suggestion for the future: I know that using too many contexts can be cumbersome, but what about this?

```ruby
describe "#installed?" do
  before do # this is common code
    allow(Yast::Service).to receive("Enabled").with(described_class::SERVICE).and_return(installed)
  end

  context "when firewalld is installed" do
    let(:installed) { true }

    it "returns true" do
      expect(firewalld).to be_enabled # or expect(firewalld.enabled?).to eq(true)
    end
  end

  context "when firewalld is not installed" do
    let(:installed) { false }

    it "returns false" do
      expect(firewalld).to_not be_enabled # or expect(firewalld.enabled?).to eq(false)
    end
  end
end
```

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -0,0 +1,120 @@
...
+require_relative "../test_helper"
+require "y2firewall/firewalld"
+
+Yast.import "PackageSystem"
+Yast.import "Service"
+
+describe Y2Firewall::Firewalld do
+  let(:firewalld) { described_class.instance }
+
+  describe "#installed?" do
```
imobachgs: I miss here a test in case firewalld is not installed.

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -0,0 +1,120 @@
...
+    it "returns true it the firewalld is installed" do
+      allow(Yast::PackageSystem).to receive("Installed")
+        .with(described_class::PACKAGE).and_return true
+      expect(firewalld.installed?).to eq(true)
+    end
+  end
+
+  describe "#enabled?" do
+    it "returns true if the firewalld service is enable" do
```
imobachgs: *enabled*

imobachgs: I miss negative test too.

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  describe "#stop" do
+    before do
+      allow(firewalld).to receive("installed?").and_return true
+      allow(firewalld).to receive("running?").and_return true
+    end
+
+    it "returns false if the service is not installed" do
+      allow(firewalld).to receive("installed?").and_return false
```
imobachgs: np: I think that it would be better if you use contexts as explained above.

library/network/test/y2firewall/firewalld_test.rb
teclator: As it was a context with only one test I didn't use it, but it probably make the test more readable.

library/network/test/y2firewall/firewalld_test.rb
teclator: only tested the positive case

PR #639, opened 2017-11-10
new generation of popup library

jreidinger:
New generation of Popup that have simplified and consistent API. It also heavily benefit from ruby and its code is significantly smaller then old Yast::Popup and easier to read.

For demonstration see popup_serie.rb in version showed below and in short gif how it looks like:

```ruby
require "yast"

require "yast2/popup"

Yast2::Popup.show("Simple text")

Yast2::Popup.show("Long text\n" * 50)

Yast2::Popup.show("Simple text with details", details: "More details here")

Yast2::Popup.show("Simple text with timeout", timeout: 10)

Yast2::Popup.show("Continue/Cancel buttons", buttons: :continue_cancel)

Yast2::Popup.show("Yes/No buttons", buttons: :yes_no)

Yast2::Popup.show("Yes/No buttons with No focused", buttons: :yes_no, focus: :no)

Yast2::Popup.show("Yes/No buttons with timeout returning focused item", buttons: :yes_no, focus: :no, timeout: 10)

Yast2::Popup.show("Own buttons", buttons: { button1: "button 1", button2: "button 2" }, focus: :button2)

Yast2::Popup.show("Richtext is set to <b>false</b>", richtext: false)

Yast2::Popup.show("Richtext is set to <b>:auto</b>", richtext: :auto)

Yast2::Popup.show("Long text. Richtext is set to <b>:auto</b>\n" * 50, richtext: :auto)

Yast2::Popup.show("Richtext is set to <b>true</b>", richtext: true)

Yast2::Popup.show("Long text. Richtext is set to <b>true</b><br>" * 50, richtext: true)

Yast2::Popup.show("Long text with newlines. Richtext is set to <b>true</b>\n" * 50, richtext: true)

Yast2::Popup.show(":notice style", style: :notice)

Yast2::Popup.show(":important style", style: :important)

Yast2::Popup.show(":warning style", style: :warning)

Yast2::Popup.show("Headline set", headline: "Headline")

Yast2::Popup.show("Headline set to :error", headline: :error)

Yast2::Popup.show(
  "All options",
  headline: "Headline",
  details: "details",
  timeout: 10,
  buttons: :yes_no,
  focus: :no,
  richtext: false,
  style: :important
)

Yast2::Popup.feedback("feedback for time consuming operation", headline: "Syncing...") { sleep(5) }
```

and yast client in action:
popup


library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+
+      # Number of lines for richtext: :auto to switch to richtext widget
+      LINES_THRESHOLD = 20
+
+      RICHTEXT_WIDTH = 60
+      RICHTEXT_HEIGHT = 10
+
+      # Shows popup and return which symbol of pressed button
+      # @param message [String] message to show
+      # @param details [String] hidden details that can be shown, if empty then it is nil shown
```
lslezak: Parse error: `if empty then it is nil show`

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+        end
+      end
+
+    private
+
+      include Yast::UIShortcuts
+
+      def check_arguments!(message, details, timeout, focus, buttons)
+        if !message.is_a?(::String)
+          raise ArgumentError, "Invalid value #{message.inspect} of parameter message"
```
lslezak: NP: I think it would be nice to include the expected type in the error messages, like for the buttons.

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+        if headline.empty?
+          [Empty()]
+        else
+          [Left(Heading(headline)), VSpacing(0.2)]
+        end
+      end
+
+      def timeout_widget(timeout)
+        if timeout > 0
+          Label(Id(:label), timeout.to_s)
```
lslezak: NP: maybe something like `:timeout_label` would be better here, `:label` is too generic IMHO...

library/general/src/lib/yast2/popup.rb
jreidinger: ah, it should be `if empty then nothing is shown`. Probably I should be more explicit about behavior here. Like something like:
```
# @param details [String] if it is not empty, it adds button to show details, which after clicking open addition popup with given details. If string is empty, nothing is shown
```

lslezak:
Just a note to the `details` option - opening a popup for a popup does not look nice. I'd display the details at the bottom of the current popup and toggle the button to `Hide details`.

Or some YaST dialogs have `Show details` checkbox, that could also be an option...

BTW how does a dialog with details and timeout work? What if you show the details and leave the time run out? With a single popup the behavior should be easy to understand for users.

I know this implements the old behavior but as we introduce the new API we do not need to exactly follow the look&feel of the old library...

lslezak:
Anyway, this is a big step forward regarding the code usability! :+1: :+1:

library/general/src/lib/yast2/popup.rb
jreidinger: buttons does not have it specified. Only focus prints list of buttons that can be focused. I don't want to make exception too long and print mainly the most important data, if user see it, so it is less prone for error when reporting or screenshooting it.

library/general/src/lib/yast2/popup.rb
jreidinger: right, I keep same ID as old Popup and it use both, so longer will be better.

jreidinger:
@lslezak well, problem with showing details in some kind of replace point is problem with dialog size, as it can easily exceed size of screen ( especially in ncurses ).

Regarding timeout and details together. it stop timer if you click on details and it continue counting when you get back.

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+      # @param message [String] message to show
+      # @param details [String] hidden details that can be shown, if empty then it is nil shown
+      # @param headline [String, :error, :warning] sets popup headline. When String is passed it is shown as it is.
+      #   When empty string is passed no headline is shown.
+      #   If Symbol is passed it have to be from predefined set of symbols.
+      #   Note: Symbol it is just predefined strings, does not modify style of popup.
+      # @param timeout [Integer] how long wait till autoclose dialog. 0 means wait forever.
+      # @param buttons [Hash<Symbol, String>, Symbol] specified which buttons popup will have.
+      #   it can be hash in format button_id => button_text, showed in same order as in hash.
+      #   Beware that :details and :stop id is reserved.
```
ancorgs: Wouldn't it be better to use other ids with lower chances of collision for those special buttons? Something like `:__details` or `:_stop_` or any other trick? Even those values are documented to be reserved is too easy to overlook that and introduce errors.

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+      # Shows feedback till given block does not finish.
+      # @param message [String] message to show
+      # @param headline [String] sets popup headline. String is shown.
+      #   If empty string is passed no headline is shown.
+      # @yield [] operation that needs to show feedback
+      # @return result of block
+      def feedback(message, headline: "", &block)
+        headline = generate_headline(headline)
+        if !message.is_a?(::String)
+          raise ArgumentError, "Invalid value #{message.inspect} of parameter message"
```
mvidner: Confident Ruby says: thou shall check arguments as the first thing in a method. So, before `generate_headline`

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+
+        # relax sanity check as there can be situation where is
+        # e.g. OK and details with `show(text, details: text2)`
+        ButtonBox(Opt(:relaxSanityCheck), *push_buttons)
+      end
+
+      def event_loop(content, focus, timeout, details, style)
+        res = Yast::UI.OpenDialog(dialog_options(style), content)
+        remaining_time = timeout
+        raise "Failed to open dialog, see logs." unless res
```
mvidner: before `remaining_time =`, please

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+      #   Yast2::Popup.show(message, richtext: true, timeout: timeout_seconds)
+      def show(message, details: "", headline: "", timeout: 0, focus: nil, buttons: :ok, richtext: :auto, style: :notice)
+        textdomain "base"
+        buttons = generate_buttons(buttons)
+        headline = generate_headline(headline)
+        # add default focus button before adding details, as details should not be focussed
+        focus = buttons.keys.first if focus.nil?
+        buttons = add_details_button(buttons) unless details.empty?
+        buttons = add_stop_button(buttons) if timeout > 0
+        check_arguments!(message, details, timeout, focus, buttons)
```
mvidner: Confident Ruby says: thou shall check arguments as the first thing in a method. 
Hmm, I see it interacts with the `buttons` and `focus` above. I would make it handle the simple args only for clarity: `check_arguments!(message, details, timeout)`

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+        event_loop(content_res, focus, timeout, details, style)
+      end
+
+      # Shows feedback till given block does not finish.
+      # @param message [String] message to show
+      # @param headline [String] sets popup headline. String is shown.
+      #   If empty string is passed no headline is shown.
+      # @yield [] operation that needs to show feedback
+      # @return result of block
+      def feedback(message, headline: "", &block)
```
ancorgs: The block syntax is the most convenient and ruby-like one. But I think it's also good to offer the explicit open and close methods. They are useful in some situations and even the Ruby standard lib offers explicit methods to close a file (so `open` can be used without block) and so. Just like we do in https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/rspec/scr.rb#L8

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+        end
+      end
+
+      def add_details_button(buttons)
+        # use this way merge to have details as first place button
+        { details: _("&Details...") }.merge(buttons)
+      end
+
+      def add_stop_button(buttons)
+        # use this way merge to have details as first place button
```
ancorgs: Comment copy-paste.

And, in fact, to have `:stop` at the end you only need to do an straightforward `buttons[:stop] = Yast::Label.StopButton`

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+      #
+      #   Yast::Popup.TimedError(text, seconds)
+      #   Yast2::Popup.show(text, headline: :error, timeout: seconds)
+      #
+      #   Yast::Popup.TimedErrorAnyQuestion(headline, message, yes_button_message, no_button_message, focus, timeout_seconds)
+      #   Yast2::Popup.show(message, headline: headline, timeout: timeout_seconds, buttons: { yes: yes_button_message, no: no_button_message), focus: :yes)
+
+      #   Yast::Popup.TimedLongNotify(message, timeout_seconds)
+      #   Yast2::Popup.show(message, richtext: true, timeout: timeout_seconds)
+      def show(message, details: "", headline: "", timeout: 0, focus: nil, buttons: :ok, richtext: :auto, style: :notice)
```
ancorgs: Supporting `nil` for some arguments like `:details`, `:headline`, or `:timeout` (in addition to zero and empty strings) would not be much work and would be quite rubist (or maybe railsty). In any case, I'm sure it will not hurt.

library/general/src/lib/yast2/popup_rspec.rb
```diff
@@ -0,0 +1,20 @@
+require "yast"
+
+Yast.import "UI"
+
+require "yast2/popup"
+
+def expect_to_show_popup_which_return(output)
```
ancorgs: I understand the intention of these helpers, but it simply doesn't "feel right"(TM). First of all, there should be yardoc documentation here explaining what are they good for (mocking the dialog but still failing if the arguments are wrong). And even though... I don't like them. Can I finish my own HackWeek stuff before elaborating further?

library/general/src/lib/yast2/popup.rb
jreidinger: makes sense for me, I will modify it.

ancorgs:
I know this is UI code, but I believe some RSpec tests would be possible and useful (currently there are only tests for the RSpec helpers, but not for Popup itself).

library/general/src/lib/yast2/popup.rb
jreidinger: make sense and I plan to modify even that details as there is a bug in ButtonBox

library/general/src/lib/yast2/popup.rb
jreidinger: well, I explicitelly do not want to use nil. I would like to avoid it where possible as it is for me a bit error prone and require explicit type looks better for me and also `headline: ""` looks better for me then `headline: nil`

library/general/src/lib/yast2/popup_rspec.rb
jreidinger: yes, no problem

jreidinger:
@ancorgs in fact that rspec helper test tests popup itself as a nice bonus. as even coveralls demonstrate :)

ancorgs:
@jreidinger smoke coverage != proper behavior-oriented tests :)

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+      end
+
+      def handle_event(res, details, remaining_time, focus)
+        case res
+        when :details
+          show(details)
+          nil
+        when :timeout
+          if remaining_time <= 0
+            :focus
```
mvidner: Hmmm

jreidinger: ah, good catch, it should be `focus` not symbol

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+      #   `continue_cancel` ->   `{ continue: Label.ContinueButton, cancel: Label.CancelButton }`
+      #   `:yes_no` -> `{ yes: Label.YesButton, no: Label.NoButton }`
+      # @param focus [Symbol, nil] what button focus.
+      #   Also it is button which is returned if timeout exceed.
+      #   if it is nil, then choose the first button. See buttons parameter.
+      # @param richtext [Boolean, :auto] if use richtext widget.
+      #   Useful when report contain richtext tags or is long, so it have to be scrollable.
+      #   If :auto is used, then it detect too long text and use richtext for it and! escape
+      #   richtext tags, so if text is really richtext then use true.
+      # @param style [:notice, :important, :warning] popup dialog styling. :notice is common one,
```
mvidner: Huh, it still does not tell me the color :-( Can we say, "important is green, warning is red"?

jreidinger: it is defined in style sheet. To be honest I do not know and I think it is different in SLE installer and in runtime.

library/general/src/lib/yast2/popup.rb
```diff
@@ -0,0 +1,318 @@
...
+      #   Beware that :details and :stop id is reserved.
+      #   The second option is symbol that specify one of predefined set of buttons.
+      #   Current set of predefined buttons are
+      #   `:ok` -> `{ ok: Label.OKButton}`
+      #   `continue_cancel` ->   `{ continue: Label.ContinueButton, cancel: Label.CancelButton }`
+      #   `:yes_no` -> `{ yes: Label.YesButton, no: Label.NoButton }`
+      # @param focus [Symbol, nil] what button focus.
+      #   Also it is button which is returned if timeout exceed.
+      #   if it is nil, then choose the first button. See buttons parameter.
+      # @param richtext [Boolean, :auto] if use richtext widget.
```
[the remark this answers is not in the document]
jreidinger: well, in old API it was as LongText ( that take widget :open_mouth:  ) [1], there was some for richtext with tags[2] and then some that decide that automatic[3] and some that always use simple[4].

[1] https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Popup.rb#L633
[2] https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Popup.rb#L674
[3] https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Popup.rb#L1322
[4] https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Popup.rb#L1450

use case for auto is to have scrollable text when you display something that you do not know how long it will be, but you do not want to interpret richtext tags.

so use cases for this param:

- explicit label, so never use richtext widget ( to be honest I don't know use case for this ).
- auto, which is simple text that is decided if shown scrollable or not
- explicit richtext, when you want to interpret richtext tags


so maybe if we strip first one, for which I do not see good use case, we can simplify it to `tags: true|false`. What do you think?

library/general/src/lib/yast2/popup.rb
jreidinger: so you think to split simple and with buttons/focus check?

mvidner: I was just nitpicking. It's not a big deal and I found more important things.

library/general/src/lib/yast2/popup_rspec.rb
ancorgs: Too busy. Don't block the pull request because of me. :-)

library/general/src/Makefile.am
```diff
@@ -93,7 +93,13 @@ ylib2_DATA = \
   lib/ui/text_helpers.rb \
   lib/ui/widgets.rb
 
-EXTRA_DIST = $(module_DATA) $(client_DATA) $(scrconf_DATA) $(agent_SCRIPTS) $(ydata_DATA) $(fillup_DATA) $(ylib_DATA) $(ylib2_DATA)
+ylib3dir = "${yast2dir}/lib/yast2"
```
lslezak: NP: I do not like the numbers in the identifier much, `ylib2dir`, `ylib3dir`,... that's not much descriptive and is a bit confusing.

I'd prefer names like `ylibyast2dir`, `ylibinstallationdir`, `ylibuidir` if possible. (Or maybe with an underscore like `ylib_yast2dir` or `ylib_yast2_dir`, but I'm not sure what automake actually accepts here...)

test/test_helper.rb
```diff
@@ -10,6 +10,9 @@
 inc_dirs.concat(ADDITIONAL_Y2DIRS)
 ENV["Y2DIR"] = inc_dirs.join(":")
 
+ENV["LC_ALL"] = "C"
```
lslezak: Um, I'm not sure if the plain `C` is the best option, I usually rather prefer `en_US.UTF-8` just to possibly avoid potentional UTF-8 issues. (I do not have any real example where it would make troubles, but my feeling is that using the UTF-8 locale is more safe.)

jreidinger: ok, will change it.

test/test_helper.rb
```diff
@@ -10,8 +10,8 @@
 inc_dirs.concat(ADDITIONAL_Y2DIRS)
 ENV["Y2DIR"] = inc_dirs.join(":")
 
-ENV["LC_ALL"] = "C"
-ENV["LANG"] = "C"
+ENV["LC_ALL"] = "C.UTF-8"
+ENV["LANG"] = "C.UTF-8"
```
lslezak: I did not know this trick, hopefully it does not break ;-)

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Nov 28 11:44:30 UTC 2017 - jreidinger@suse.com
+
+- Add new Yast2::Popup class providing nicer API for invoking
+  popups (hackweek project)
```
lslezak: Nice to mention the hackweek :+1:

PR #642, opened 2017-11-23
Add yast-rake dependency

imobachgs:
Add missing yast-rake dependency. It should help with some of the packages that are now failing to build.


package/yast2.spec
```diff
@@ -52,6 +52,8 @@ BuildRequires:  yast2-ycp-ui-bindings >= 3.2.0
 # these should be installed in the default build anyway
 BuildRequires:  rpm
 BuildRequires:  cpio
+# YaST rake tasks
+BuildRequires:  rubygem(%{rb_default_ruby_abi}:yast-rake)
```
lslezak: NP: I'd just move the line up so the `BuildRequires:  rubygem(%{rb_default_ruby_abi}:X)` are together.

jreidinger:
why it is needed in yast2? yast2 does not use rake install, so it is not needed

PR #645, opened 2017-11-24
Use full type detection when building device cache

mchf:
Replaces https://github.com/yast/yast-yast2/pull/527

The fix replaces ifcfg only type detection by full type detection when building device config cache. type detection based on ifcfg can differ from the reality mainly in situations when the config file contains some unexpected option. E.g. when user leaves `ETHERDEVICE=eth0` in eth device's config file then ifcfg based type detection can report type `vlan` but correct one is `eth`.


mchf:
thanks for the review

PR #646, opened 2017-11-28
Merged SLE-12-SP2 into SLE-12-SP3

mchf:


mchf:
Thanks for the review

PR #647, opened 2017-11-28
Merged SLE-12-SP3 into master

mchf:


mchf:
thanks for the review

PR #650, opened 2017-11-30
document logger switch

jreidinger:


library/system/src/lib/yast2/execute.rb
```diff
@@ -28,6 +28,7 @@ module Yast
   # (not prone to shell quoting bugs).
   # It uses {http://www.rubydoc.info/github/openSUSE/cheetah/ Cheetah}
   # as the backend, but adds support for chrooting during the installation.
+  # It also switches default logger to Yast logger ( see Y2Logger ).
```
mvidner: Thanks! I suggest
- better wording
- a resovable link (optional)
```
  # It also globally switches the default Cheetah logger to
  # {http://www.rubydoc.info/github/yast/yast-ruby-bindings/Yast%2FLogger Y2Logger}.
```

jreidinger:
failed for non-relevant fillup change

PR #652, opened 2017-12-04
API improvement

mchf:
bnc#1061306

### DIFF
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -829,6 +829,8 @@ module Yast
     # Hash map is direct maping of sysconfig file into hash.
     # Keys are sysconfig options (e.g. { 'IPADDR' => '1.1.1.1' }
     #
+    # @param [String] name is device name as provided by the
+    #                 system (e.g. eth0)
     # @return [Hash] device configuration or nil in case of error
     def devmap(name)
       Devices().fetch(GetType(name), {})[name]
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -29,6 +29,8 @@ module Yast
   # Categorizes the configurations according to type.
   # Presents them one ifcfg at a time through the {#Current} hash.
   class NetworkInterfacesClass < Module
+    attr_reader :Devices
+
     include Logger
 
     Yast.import "String"
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -822,6 +822,16 @@ module Yast
       Read()
     end
 
+    # Returns a hash with configuration for particular device
+    #
+    # Hash map is direct maping of sysconfig file into hash.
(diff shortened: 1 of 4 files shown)


library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -822,6 +824,16 @@ def CleanCacheRead
...
     end
 
+    # Returns a hash with configuration for particular device
+    #
+    # Hash map is direct maping of sysconfig file into hash.
+    # Keys are sysconfig options (e.g. { 'IPADDR' => '1.1.1.1' }
+    #
+    # @return [Hash] device configuration or nil in case of error
+    def devmap(name)
+      Devices().fetch(GetType(name), {})[name]
```
teclator: Wasn't this the cause of some bugs in the past? I mean, why use the GetType to obtain the ifcfg map instead of look based on ifcfg-name ?

teclator:
The bug was created against SP2, the API improvements in master branch are intentional?

teclator:
After checked previous PRs it seems that for SP2 and SP3 the fixes are in #645 and #646 and merged into master in #647

library/network/src/modules/NetworkInterfaces.rb
mchf: no the issue was caused by the fact that the cache was constructed using type detection based on ifcfg only. Then it could happen (and happened) that searching for device configuration failed bcs according to ifcfg it was recognized as (e.g.) vlan but in fact according to hwinfo (/sys) it was (e.g.) eth. This was fixed by PRs you refer bellow. Now is cache constructed according to same rules which ```GetType``` uses for type recognition (look into /sys first, ifcfg as a backup only). 

Anyway, this PR only moves code from y2-network (```GetDeviceMap```) into yast2 package (```devname```). I think that this code do not belong to y2-network bcs it looks into internal structure of NetworkInterface's cache. I use the bug reference only due to BS politics. The fix itself is in PRs referenced bellow.

mchf:
yes api improvements are intentionally in master only.  Fix for SLE-12-SP2 is here https://github.com/yast/yast-yast2/pull/645 In fact this is one small piece from LanItems::Items removal patch.

library/network/src/modules/NetworkInterfaces.rb
teclator: Yep, if now both rely on same type detection there should not be inconsistencies anymore, and that was the main reason why in `GetDeviceMap` the map was search just matching the ifcfg name and not relying in the ifcfg type anymore.

From performance POV it is better to use the type index, just hope it does not break anymore because of that...

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -822,6 +824,16 @@ def CleanCacheRead
       Read()
     end
 
+    # Returns a hash with configuration for particular device
+    #
+    # Hash map is direct maping of sysconfig file into hash.
+    # Keys are sysconfig options (e.g. { 'IPADDR' => '1.1.1.1' }
+    #
+    # @return [Hash] device configuration or nil in case of error
+    def devmap(name)
```
teclator: Add the `@param name [String]` doc please

mchf:
thanks for the review

PR #653, opened 2017-12-04
Recognize CTC devices named like slc600

mvidner:
[bsc#1058227](https://bugzilla.suse.com/show_bug.cgi?id=1058227)
https://trello.com/c/SLjwpn09

~~The actual bug fix is the commit 40fca35e6748b08369244d8dca18c7525e19c7f6: it lets YaST "see" a network card config even if its name is "not normal". It is more of a workaround than a fix. (A real fix may involve big refactoring)~~

In the end a more conservative solution, 846c3f03f07a884eb90b4a7811ccb7c4f21d8a8c.


mvidner:
@mchf can you please review this?

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -113,39 +110,18 @@ def main
...
-        "modem"   => Ops.get(@CardRegex, "modem", ""),
-        "isdn"    => Ops.add(
-          Ops.get(@CardRegex, "isdn", ""),
-          HotplugRegex(["isdn", "ippp"])
-        ),
-        "dsl"     => Ops.get(@CardRegex, "dsl", ""),
+        "netcard" => @CardRegex["netcard"],
+        "modem"   => @CardRegex["modem"],
+        "isdn"    => @CardRegex["isdn"],
+        "dsl"     => @CardRegex["dsl"],
```
mchf: modem, isdn and dsl support was removed a long time ago. I think this can be dropped.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -113,39 +110,18 @@ def main
...
-          Ops.add(
-            Ops.add(
-              Ops.add(Ops.get(@CardRegex, "modem", ""), "|"),
-              Ops.get(@CardRegex, "dsl", "")
-            ),
-            "|"
-          ),
-          Ops.get(@CardRegex, "isdn", "")
-        )
+        "dialup"  => @CardRegex["modem"] + "|" + @CardRegex["dsl"] + "|" + @CardRegex["isdn"]
```
mchf: see above

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -205,22 +181,8 @@ def main
...
-          ret += "|#{dev}-#{hot}|#{dev}-#{hot}-"
-        end
-      end
-      ret
-    end
-
     def IsEmpty(value)
-      value = deep_copy(value)
-      TypeRepository.IsEmpty(value)
+      value.nil? ? true : value.empty?
```
mchf: finally someone did it ;-) Thanks

library/network/src/modules/NetworkInterfaces.rb
mvidner: OK. I was not sure how far I should go in this bugfix. Should we do it here?

library/network/src/modules/NetworkInterfaces.rb
mchf: the ```TypeRepository::IsEmpty``` should be removed or replaced to avoid confusion.

library/network/src/modules/NetworkInterfaces.rb
mchf: well you touch it anyway, so why not.

library/network/test/network_interfaces_test.rb
```diff
@@ -91,7 +91,7 @@
   describe "#FilterDevices" do
     let(:data_dir) { File.join(File.dirname(__FILE__), "data") }
     # Defined in test/data/etc/sysconfig/ifcfg-*
-    let(:netcard_devices) { ["arc", "bond", "br", "em", "eth", "vlan"] }
+    let(:netcard_devices) { ["arc", "bond", "br", "em", "eth", "ppp", "vlan"] }
```
mchf: ```ppp``` was not handled as netdevice before but modem. So, why was it added there?

mvidner: That's a consequence of 40fca35: we really don't know how an interface may be named, so we basically abolish the name based filtering. So *all* names get through (well, except "lo"). Even ppp. I think the result is less wrong.

mchf: well that is more less true. But the ```Devices``` cache is organized into "categories" based on device type. This types were then categorized as ("netcard", "modem", ...) and so on. This change means that if you have a device of type ```ppp``` which used to be classified as ```modem``` it will be classified as a ```netcard``` after this update. This might affect e.g. ```yast2 lan``` overview dialog.

Device name is type independent for some time already, so you can have device of ```eth``` type with ```ppp``` name even now.

PR #654, opened 2017-12-05
Make #current_page public

joseivanlopez:
Necessary for PBI: https://trello.com/c/kE0SSbi5/698-0-partitioner-plain-partitioning-hard-disks-overview-add-edit-buttons

### DIFF
--- a/library/cwm/src/lib/cwm/pager.rb
+++ b/library/cwm/src/lib/cwm/pager.rb
@@ -13,6 +13,9 @@ module CWM
   #
   # @see examples/object_api_tabs.rb
   class Pager < CustomWidget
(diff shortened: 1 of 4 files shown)


library/cwm/src/lib/cwm/pager.rb
```diff
@@ -13,6 +13,9 @@ module CWM
   #
   # @see examples/object_api_tabs.rb
   class Pager < CustomWidget
+    # @return [CWM::Page] currently selected page
```
ancorgs: strictly speaking, this can also return nil, worth documenting

ancorgs: hmm, only when called before `#init`. Still worth documenting? I'm not sure :-)

PR #657, opened 2017-12-11
Infrastructure for preserving comments in config files

shundhammer:
Imported classes from my upstream repository:

https://github.com/shundhammer/ruby-commented-config-file

That repo also contains some more examples (and general-purpose diff classes that I didn't import here and now) and of course some introductory comments.

Those classes are now used for the SLE-12-SP3 yast-nfs-client if it runs in standalone mode (i.e. outside the expert partitioner); see also the related PR against yast-nfs-client here: https://github.com/yast/yast-nfs-client/pull/56

### DIFF
--- a/library/system/src/lib/yast2/column_config_file.rb
+++ b/library/system/src/lib/yast2/column_config_file.rb
@@ -116,6 +116,8 @@ class ColumnConfigFile < CommentedConfigFile
   #
   # @return [Array<Fixnum>] column widths
   #
+  # rubocop:disable Style/For
+  #
   def calc_column_widths
     return [] unless @pad_columns
     @column_widths = []
@@ -136,7 +138,7 @@ class ColumnConfigFile < CommentedConfigFile
     ColumnConfigFile::Entry.new(self)
   end
 
-  protected
+protected
 
   # Count the maximum number of columns amont all entries.
   #
@@ -254,7 +256,7 @@ class ColumnConfigFile < CommentedConfigFile
       format
     end
 
-    protected
+  protected
 
     # Return the column delimiter used for input (parsing).
     #
--- a/library/system/src/lib/yast2/column_config_file.rb
+++ b/library/system/src/lib/yast2/column_config_file.rb
@@ -1,0 +1,290 @@
+#!/usr/bin/env ruby
+#
+# CommentedConfigFile class
+#
+# (c) 2017 Stefan Hundhammer <Stefan.Hundhammer@gmx.de>
+#     Donated to the YaST project
+#
+# Original project: https://github.com/shundhammer/ruby-commented-config-file
+#
+# License: GPL V2
+#
+
+require "yast2/commented_config_file"
+
+# Utility class to read and write column-oriented config files that might
+# contain comments that should be preserved when writing the file.
+#
+# One example for this would be /etc/fstab:
+#
+#    # /etc/fstab
+#    #
+#    # <file system>	<mount point>  <type> <options>	    <dump> <pass>
+#
+#    /dev/disk/by-label/swap	 none	swap  sw		 0  0
+#    /dev/disk/by-label/Ubuntu	 /	ext4  errors=remount-ro	 0  1
+#    /dev/disk/by-label/work	 /work	ext4  defaults		 0  2
+#
+# There are 6 columns, separated by whitespace. This class is a refinement of
+# the more generic CommentedConfigFile class to handle such cases.
+#
+class ColumnConfigFile < CommentedConfigFile
+  DEFAULT_MAX_COLUMN_WIDTH = 40
+  DEFAULT_INPUT_DELIMITER = /\s+/
+  DEFAULT_OUTPUT_DELIMITER = "  ".freeze
+
+  # @return [Fixnum] the fallback value for the maximum column width if no
+  # per-column value is specified for a column.
+  #
+  attr_accessor :fallback_max_column_width
+
+  # @return [Array<Fixnum>] Per-column maximum width for each column. If not
+  # specified for a column, 'fallback_max_column_width' is used.
+  #
+  attr_accessor :max_column_widths
+
+  # @return [Boolean] Flag: Pad the columns to a common width (up to each
+  # column's maximum width) or not?
+  #
+  attr_accessor :pad_columns
+
+  # @return [Regexp] Input column delimiter, used for parsing content lines.
+  # This is usually one or more whitespace characters, but it might also be
+  # something completely different like one colon for /etc/passwd.
+  #
+  # If this is non-blank, pad_columns should probably also set to 'false'
+  # because it will always pad with blanks which would probably not be
+  # appropriate for that file format.
+  attr_accessor :input_delimiter
+
+  # @return [String] Output column delimiter; this is used when formatting
+  # columns. The default is " " (two blanks).
+  attr_accessor :output_delimiter
+
+  # @return [Array<Fixnum>] The last column widths calculated.
+  #
+  attr_reader :column_widths
+
+  def initialize
+    super
+    @max_column_widths = []
+    @fallback_max_column_width = DEFAULT_MAX_COLUMN_WIDTH
+    @pad_columns = true
+    @input_delimiter = DEFAULT_INPUT_DELIMITER
+    @output_delimiter = DEFAULT_OUTPUT_DELIMITER
+    @column_widths = []
+  end
+
+  # Format only the entries without header or footer comments, but with
+  # comments before each entry and with the line comments.
+  #
+  # Reimplemented from CommentedConfigFile.
+  #
+  # @return [Array<String>] formatted entries
+  #
+  def format_entries
+    populate_columns
+    calc_column_widths
+    super
+  end
+
+  # Get the column with for one column.
+  # If padding is not enabled, this returns 0.
+  #
+  # @param column_no [Fixnum] number of the column (from 0)
+  #
+  # @return [Fixnum] column width
+  #
+  def get_column_width(column_no)
+    return 0 unless @pad_columns
+    calc_column_widths if @column_widths.empty?
+    @column_widths[column_no] || 0
+  end
+
+  # Get the maximum column with for a column.
+  #
+  # @param column_no [Fixnum] number of the column (from 0)
+  #
+  # @return [Fixnum] column width
+  #
+  def get_max_column_width(column_no)
+    @max_column_widths[column_no] || @fallback_max_column_width
+  end
+
+  # If @pad_columns is set, calculate the best column widths for all columns
+  # and store the result in @column_widths.
+  #
+  # @return [Array<Fixnum>] column widths
+  #
+  def calc_column_widths
+    return [] unless @pad_columns
+    @column_widths = []
+
+    for col in 0...count_max_columns
+      @column_widths[col] = calc_column_width(col)
+    end
+    @column_widths
+  end
+
+  # Create a new entry.
+  #
+  # Reimplemented from CommentedConfigFile.
+  #
+  # @return [ColumnConfigFile::Entry] new entry
+  #
+  def create_entry
+    ColumnConfigFile::Entry.new(self)
+  end
+
+  protected
+
+  # Count the maximum number of columns amont all entries.
+  #
+  # @return [Fixnum]
+  #
+  def count_max_columns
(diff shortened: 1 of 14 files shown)


jreidinger:
Well, it have to be fun to write it and here is my general comment without any looking at code quality or API quality. So why I do not like this change?

1. it is a lot of code we have to maintain ( 300 + 450 + 750 lines ) =~ 1500 lines in src
2. it is a lot of code that can ( and will contain bugs )
3. it is against NIH. I think we are not the first one that tries to parse configuration file with comments.
4. it is missing features like switched SCR or more complex syntax
5. we already have two existing infrastructure in code - old SCR which sucks in this case and new CFA which does not have issues mentioned above and what is better is uses augeas for parsing, so it have already syntax for cca 150 types of configuration files and also we are not alone to maintain it, to answer question and its bus factor is very high. To demonstrate it, I can write write etc_fstab file using CFA with same API and I think it will be even shorter.

shundhammer:
Josef, this is too long and too compex to answer here.

shundhammer:
Some single points:

> it is a lot of code we have to maintain ( 300 + 450 + 750 lines ) =~ 1500 lines in src

Excluding comment and blank lines, it's 488 lines of source code and 756 lines of unit tests.

lslezak:
My opinion: Josef's points are valid, but as this is (AFAIK) only for SP3 to fix an L3 I'd be more pragmatic here.

library/system/src/lib/yast2/etc_fstab.rb
```diff
@@ -0,0 +1,566 @@
...
+    # format).
+    #
+    # @param opts [String] mount options field
+    #
+    def parse_mount_opts(opts)
+      if opts.nil?
+        @mount_opts = []
+        return
+      end
+      @mount_opts = opts.split(/,/)
```
lslezak: NP: I think simple `split(",")` should be enough here.

library/system/test/column_config_file_test.rb
```diff
@@ -0,0 +1,114 @@
...
+          expect(entry.columns[4]).to eq "0"
+          expect(entry.columns[5]).to eq "0"
+        end
+      end
+    end
+  end
+
+  context "Formatter" do
+    describe("#format_lines") do
+      def read_twice(filename)
```
lslezak: The method name is a bit confusing to me, I expected doing something twice. But it actually reads the file, parses it  and reformats back. So something like `read_and_reformat` would be better. That would also correspond to the two returned values (`read` and `reformatted`).

library/system/test/commented_config_file_test.rb
```diff
@@ -0,0 +1,314 @@
...
+#
+
+require_relative "test_helper"
+require "yast2/commented_config_file"
+
+describe CommentedConfigFile do
+  # rubocop:disable Lint/AmbiguousRegexpLiteral
+
+  context "when created empty" do
+    subject { described_class.new }
```
lslezak: IIRC `subject { described_class.new }` is the default, this line can be omitted. (See [Implicitly defined subject](https://relishapp.com/rspec/rspec-core/v/3-6/docs/subject/implicitly-defined-subject) documentation.)

library/system/test/commented_config_file_test.rb
```diff
@@ -0,0 +1,314 @@
...
+        expect(subject.empty_line?(" \n")).to eq true
+        expect(subject.empty_line?("\t")).to eq true
+        expect(subject.empty_line?("\t\n  \t\n")).to eq true
+      end
+
+      it "Rejects non-empty lines" do
+        expect(subject.empty_line?("x")).to eq false
+        expect(subject.empty_line?("  x")).to eq false
+        expect(subject.empty_line?("  x  ")).to eq false
+        expect(subject.empty_line?("  \nx  ")).to eq false
```
lslezak: `"\nx"` or `"x\n"` might be useful to test as well...

library/system/test/commented_config_file_test.rb
```diff
@@ -0,0 +1,314 @@
...
+          entry = subject.last
+          expect(entry.content).to match /fritz.box.fritz.nas.*cifs.*forcegid/
+          expect(entry.comment_before.first).to match /^\s*$/
+        end
+      end
+
+      context "Demo /etc/fstab without header and footer comments" do
+        before(:all) do
+          @file = described_class.new
+          puts("TEST_DATA: #{TEST_DATA}")
```
lslezak: Is this need? This might break the `rspec` output, esp. if somebody uses the JSON or HTML output format. (See `rspec --help` for the output formatter options.)

Or make it optional, e.g. `if ENV["DEBUG"]`.

library/system/test/commented_config_file_test.rb
```diff
@@ -0,0 +1,314 @@
...
+          expect(entry.content).to match /fritz.box.fritz.nas.*cifs.*forcegid/
+          expect(entry.comment_before.first).to match /^\s*$/
+        end
+      end
+    end
+  end
+
+  context "Formatter" do
+    describe("#format_lines") do
+      def read_twice(filename)
```
lslezak: Again, `read_twice` is a bit strange...

library/system/test/etc_fstab_entry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    it "can be created with named parameters" do
+      entry = described_class.new(device: "/dev/sda4", mount_point: "/data", fs_type: "ext4")
+      expect(entry).not_to be_nil
+      expect(entry.to_a).to eq ["/dev/sda4", "/data", "ext4", [], 0, 0]
+    end
+  end
+
+  describe "#parse" do
+    subject { EtcFstab::Entry.new }
```
lslezak: This is the default, no need for this line... Check also the other places.

library/system/src/lib/yast2/commented_config_file.rb
```diff
@@ -0,0 +1,462 @@
...
+  # @return [Boolean] true if empty, false otherwise
+  #
+  def empty_line?(line)
+    # /m : multi-line mode
+    # \z : end of string; different from $ (end of line) in multi-line strings.
+    line =~ /^\s*\z/m ? true : false
+  end
+
+  # Split a content line into the real content and any potential line comment:
+  # "foo = bar   # baz" -> ["foo=bar", "# baz"]
```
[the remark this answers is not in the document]
shundhammer: This class will not actively add any end-of-line comments, but it also will not remove them if they are there.

library/system/src/lib/yast2/etc_fstab.rb
```diff
@@ -0,0 +1,566 @@
...
+  def fs_types
+    map(&:fs_type)
+  end
+
+  # Find the (first) entry with the specified mount point. Return nil if there
+  # is no such entry.
+  #
+  # @param mount_point [String]
+  #
+  # @return [EtcFstab::Entry] or nil if not found
```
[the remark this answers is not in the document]
shundhammer: OK

library/system/test/commented_config_file_test.rb
shundhammer: Duh - that puts is a leftover of debugging those path settings. Thanks for noticing!

library/system/src/lib/yast2/etc_fstab.rb
```diff
@@ -0,0 +1,566 @@
...
+        #
+        # When two or more entries have the same mount point (which is illegal,
+        # but somebody might write such an fstab manuallly), there is no
+        # correct mount order; this algorithm would get into an endless loop if
+        # we now checked the same index again. But by just proceeding with the
+        # next one (and silently assuming that the one we just changed is well
+        # and truly fixed), we can avoid that endless loop.
+        #
+        # The fstab is of course still incorrect, but there is nothing we can
+        # do about that at this point.
```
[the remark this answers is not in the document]
shundhammer: Hm - or maybe auto-add a comment highlighting the problem. Let me look into that.

library/system/src/lib/yast2/commented_config_file.rb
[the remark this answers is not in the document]
shundhammer: Here a user has a literal '#' inside the mount options; so that would mean that in that case it should clearly not be regarded as the start of a end-of-line comment:

https://www.linuxquestions.org/questions/linux-general-1/fstab-character-but-not-comment-219468/

OTOH some posts there suggest that this appears to work on some servers and not on some others, so this might be grey territory. But for the purposes of our EtcFstab this would mean to not do the line split in this case to be safe.

_Notice that our C++ -based EtcFstab parser for libstorage-ng has the exact same problem, so it need fixing there as well._

library/system/src/lib/yast2/etc_fstab.rb
```diff
@@ -0,0 +1,566 @@
...
+  # to do it.
+  #
+  # This class and the corresponding entry class handles this transparently, so
+  # this should never be necessary to use from the outside.
+  #
+  # @param unencoded [String] String with possible space characters
+  # @return [String] String with space characters replaced by \040
+  #
+  def self.fstab_encode(unencoded)
+    unencoded.gsub(" ", '\\\\040')
```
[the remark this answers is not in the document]
shundhammer: I don't think this will work: `String::tr()` translates each character that appears in the first string to exactly one from the second string; AFAICS you cannot replace one from the first string to several from the second. What /etc/fstab needs here, however, is exactly that: Each blank is to be replaced by the sequence `\040`, i.e. a literal backslash followed by three digits specifying the octal code.

library/system/src/lib/yast2/etc_fstab.rb
shundhammer: From the Ruby `String::split()` reference documentation this should be the same. In what way do you think using a string rather than a regexp makes a difference here?

library/system/test/column_config_file_test.rb
shundhammer: I now split that up into two separate methods.

library/system/test/etc_fstab_entry_test.rb
shundhammer: Removed.

library/system/test/commented_config_file_test.rb
shundhammer: Done.

shundhammer:
Those changes should take care of the issues @lslezak discovered except for that one that unfixable mount problems remain unfixable, but that is something the user introduced before, not anything in our responsibility. I will look into that as well, but this is irrelevant for the L3 this PR is all about.

shundhammer:
BTW I did the changes primarily in the upstream repo and then applied them here as well. Upstream now has a `yast` branch that corresponds to this repo.

library/system/src/lib/yast2/column_config_file.rb
```diff
@@ -0,0 +1,292 @@
...
+  #
+  # @return [Fixnum] column width
+  #
+  def get_column_width(column_no)
+    return 0 unless @pad_columns
+    calc_column_widths if @column_widths.empty?
+    @column_widths[column_no] || 0
+  end
+
+  # Get the maximum column with for a column.
```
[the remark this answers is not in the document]
shundhammer: Fixed.

library/system/src/lib/yast2/etc_fstab.rb
lslezak: Ah, no functional difference at all, just whenever I a see a regexp I always ask myself if it is actually needed.

The point is that using a simple string is more efficient (CPU and RAM) than a regexp. I do not know the Ruby internals, maybe there is some optimization done in background, but I try to avoid regexps whenever possible.

E.g. instead of `foo ~= /bar\z/` (`foo ~= /bar$/` in some cases) or you can use simple `foo.ends_with?("bar")`. No regexp compiling, etc... just a simple string matching.

The real runtime difference will be probably negligible in this case, that's why I used a "NP" (nitpicking) comment.

You can leave it as it is.

library/system/src/lib/yast2/etc_fstab.rb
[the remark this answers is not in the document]
lslezak: Oh, good point, I did not realize that the replacement is longer...

lslezak:
> ... except for that one that unfixable mount problems remain unfixable, but that is something the user introduced before, not anything in our responsibility.

Yes, it's an user fault. My point was if we can detect this situation and at least log a warning in that case. It can make debugging much easier and faster, esp. for the y2maintainer when checking a bug log...

PR #659, opened 2017-12-14
Firewalld API: Convenient methods for query commands (bsc#1070559)

teclator:
Added convenient methods to not fail with query methods ([bsc#1070559](https://bugzilla.suse.com/show_bug.cgi?id=1070559)):

firewallcmd

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld/api.rb
+++ b/library/network/src/lib/y2firewall/firewalld/api.rb
@@ -393,7 +393,9 @@ module Y2Firewall
       # @see #command
       # @see Yast::Execute
       # @param args [Array<String>] list of command optional arguments
-      # @param permanent [Boolean] if true it ads the --permanent option the
+      # @param permanent [Boolean] if true it adds the --permanent option the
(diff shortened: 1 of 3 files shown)


library/network/src/lib/y2firewall/firewalld/api.rb
```diff
@@ -391,13 +391,42 @@ def command
       # Executes the command for the current mode with the given arguments.
       #
       # @see #command
-      # @return [String] stdout result of the command executed
-      def run_command(*args, permanent: false)
+      # @see Yast::Execute
+      # @param args [Array<String>] list of command optional arguments
+      # @param permanent [Boolean] if true it ads the --permanent option the
```
imobachgs: ads -> adds

library/network/src/lib/y2firewall/firewalld/api.rb
```diff
@@ -391,13 +391,42 @@ def command
...
+      # command to be executed
```
imobachgs: Please, document `allowed_existstatus`.

PR #662, opened 2017-12-21
Added missing files to Makefile.am

shundhammer:
Debugging the failing tests for yast2-nfs-client showed that the relevant new files were missing in this yast2 package; it turned out that that package is still using the AutoTools, not simply collecting all *.rb files like most of our other Ruby packages.

This commit now collects those files.


shundhammer:
Wrong target branch - closing

PR #664, opened 2018-01-08
 Ignore backup suffixes when loading ifcfgs

mchf:
Bnc1073727


mvidner:
Please don't forget the .changes file

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -744,6 +744,15 @@ def add_device(device, ifcfg)
       @Devices[devtype][device] = ifcfg
     end
 
+    # Device configuration files are matched against this regex
+    #
+    # The regex defines files which should not be parsed (e.g. ifcfg-eth0.bak)
+    #
+    # @return [Regexp] regexp describing ignored configurations
+    def ignore_confs_regex
+      /(.bak|.orig|.rpmnew|.rpmorig|.rpmsave|-range|~|.old|.scpmbackup)$/
```
mvidner: But *why* use these and not others? Should it be synced with another project (wicked, ...)? A URL link to the other project's definition would be nice.

`man ifcfg` on 42.3 tells me
> Configuration filenames that contain the following blacklisted extensions, will be ignored by wicked:
>       ~ .old .bak .orig .scpmbackup .rpmnew .rpmsave .rpmorig

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -752,7 +761,7 @@ def Read
       @Devices = {}
 
       # preparation
-      devices = get_devices
+      devices = get_devices(ignore_confs_regex)
```
mvidner: Apparently this is the only use of that method, which now has a default argument of `"[~]"`. Let's update it to accept a Regexp and remove the default for the argument.
Return `Array<String>`, not `Array`.
Make it private?

teclator:
@mchf @mvidner 

Isn't there a way to just skip it from the scr agent?

mchf:
@teclator:
no

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1723,6 +1709,43 @@ def ListDevicesExcept(dev)
...
+    #
+    # Moreover configuration filenames that contain the following
+    # blacklisted extensions, will be ignored by wicked:
+    # ~ .old .bak .orig .scpmbackup .rpmnew .rpmsave .rpmorig
+    #
+    # For details see bnc#1073727
+    #
+    # @return [Regexp] regexp describing ignored configurations
+    def ignore_confs_regex
+      /(.bak|.orig|.rpmnew|.rpmorig|.rpmsave|-range|~|.old|.scpmbackup)$/
```
mvidner: If I have `ifcfg-hot` and `ifcfg-cold`, the latter is ignored.

mchf:
Thanks for the review

PR #665, opened 2018-01-08
Firewalld: Allow to modify the default zone

teclator:
Trello Card: https://trello.com/c/jL8fhEKw/1091-5-firewall-firewalld-is-the-new-susefirewall-autoyast-support

### DIFF
--- a/library/network/test/y2firewall/firewalld_test.rb
+++ b/library/network/test/y2firewall/firewalld_test.rb
@@ -220,14 +220,14 @@ describe Y2Firewall::Firewalld do
       firewalld.read
 
       external = firewalld.find_zone("external")
-      expect(external.ports).to eql(["5901/tcp", "5901/udp"])
+      expect(external.ports).to eq(["5901/tcp", "5901/udp"])
     end
 
     it "initializes global options with the current firewalld config" do
       firewalld.read
 
-      expect(firewalld.log_denied_packets).to eql(false)
-      expect(firewalld.default_zone).to eql("dmz")
+      expect(firewalld.log_denied_packets).to eq(false)
+      expect(firewalld.default_zone).to eq("dmz")
     end
   end
 
@@ -315,7 +315,7 @@ describe Y2Firewall::Firewalld do
     end
 
     it "returns true" do
-      expect(firewalld.write).to eql(true)
+      expect(firewalld.write).to eq(true)
     end
   end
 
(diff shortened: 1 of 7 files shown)


library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -179,4 +181,142 @@
...
+      allow(firewalld).to receive("api").and_return api
+    end
+
+    it "returns false if firewalld is not installed" do
+      allow(firewalld).to receive(:installed?).and_return(false)
+
+      expect(firewalld.read).to eq(false)
+    end
+
+    it "initializes @zones parsing the firewalld zones list" do
```
imobachgs: I would replace *@zones* by something like *the list of zones* (being a instance variable is an implementation detail).

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -179,4 +181,142 @@
...
+      firewalld.read
+
+      external = firewalld.find_zone("external")
+      expect(external.ports).to eql(["5901/tcp", "5901/udp"])
+    end
+
+    it "initializes the @default_zone and @log_denied_packets with the firewalld config" do
```
imobachgs: Same than above.

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -179,4 +181,142 @@
...
+        expect(firewalld.modified?).to eq(true)
+      end
+    end
+
+    context "when no attribute has been modifiede since read" do
+      it "returns false" do
+        firewalld.default_zone = "public"
+        expect(firewalld.modified?).to eq(false)
+      end
+
```
imobachgs: rubocop: blank line :)

teclator:
@imobachgs fixed conflicts, updated based on CR and added export logic

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -179,4 +181,183 @@
...
+      firewalld.read
+    end
+
+    it "returns a hash with the current firewalld config" do
+      config = firewalld.export
+      external = config["zones"].find { |z| z["name"] == "external" }
+
+      expect(config).to be_a(Hash)
+      expect(config["log_denied_packets"]).to eq(true)
+      expect(config["default_zone"]).to eq("work")
```
imobachgs: np: Why do you use `eq` here and `eql` in the line below?

teclator: Probably fast typing makes me omit the "l" and then yanked and pasted, was not intentional to use eq in this case so, will modify it, Thnx ;)

PR #666, opened 2018-01-09
[Review] Request from 'schubi2' @ 'yast/yast-yast2/review_180109_double_entries'

schubi2:
  * 3ca0f009 packaging
  * ceeff81d removing double entries

### DIFF
--- a/library/control/src/modules/WorkflowManager.rb
+++ b/library/control/src/modules/WorkflowManager.rb
@@ -896,16 +896,12 @@ module Yast
       # Merging - appending steps, settings
       appends = Ops.get_list(addon, "append_modules", [])
 
-      if Ops.greater_than(Builtins.size(appends), 0)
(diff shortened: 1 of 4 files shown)


library/control/src/modules/WorkflowManager.rb
```diff
@@ -896,16 +896,12 @@ def MergeWorkflow(base, addon, _prod_name, domain)
...
-          Ops.set(
-            base,
-            "modules",
-            Builtins.add(Ops.get_list(base, "modules", []), new)
-          )
-        end
+      unless appends.empty?
+        log.info("Append: #{appends}")
+        base["modules"] << appends
+        # Removing double entries in order not to run modules again.
```
jreidinger: is it always true? what if some addon want to run some module again?

schubi2:
We are trying to solve it via the control files.

PR #670, opened 2018-01-16
Replace SuSEFirewall2 by firewalld in CWMFirewallInterfaces

teclator:
- [Trello Card](https://trello.com/c/99OkErr6/1093-8-firewall-firewalld-is-the-new-susefirewall-cwm)

The CWMFirewallInterfaces have been maintained and just adapted to work with firewalld, the only big dfiference is that CWMFirewallInterfaces.CreateOpenFirewallWidget will return a empty VBox() if some service is not supported preventing modifications:

notsupportedservices

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld.rb
+++ b/library/network/src/lib/y2firewall/firewalld.rb
@@ -23,6 +23,7 @@
 # ***************************************************************************
 
 require "y2firewall/firewalld/api"
+require "y2firewall/firewalld/service"
 require "y2firewall/firewalld/zone"
 require "y2firewall/firewalld/zone_parser"
 require "singleton"
@@ -80,6 +81,17 @@ module Y2Firewall
       zones.find { |z| z.name == name }
     end
 
+    def find_service(name)
+      services.find { |s| s.name == name } || create_service(name)
+    end
+
+    def create_service(name)
+      service = Y2Firewall::Firewalld::Service.new(name: name)
+      service.create! if !service.supported?
+      service.read
+      service
+    end
+
     # Return true if the logging config or any of the zones where modified
     # since read
     #
@@ -92,11 +104,19 @@ module Y2Firewall
 
     # Apply the changes to the modified zones and sets the logging option
     def write
+      only_write
+      reload
+    end
+
+    # Apply the changes to the modified zones and sets the logging option
+    def only_write
       zones.map { |z| z.apply_changes! if z.modified? }
       api.log_denied_packets = log_denied_packets
       api.default_zone       = default_zone
       true
     end
+
+
 
     # Return a map with current firewalld settings.
     #
--- a/library/network/src/lib/y2firewall/firewalld.rb
+++ b/library/network/src/lib/y2firewall/firewalld.rb
@@ -60,6 +60,8 @@ module Y2Firewall
     # Constructor
     def initialize
       @api = Api.new
+      @zones = []
+      @services = []
     end
 
     # Read the current firewalld configuration initializing the zones and other
@@ -73,7 +75,6 @@ module Y2Firewall
       @default_zone       = api.default_zone
       # The list of services is not read or initialized because takes time and
       # affects to the performance and also the services are rarely touched.
-      @services           = []
       true
     end
 
--- a/library/network/src/lib/y2firewall/firewalld.rb
+++ b/library/network/src/lib/y2firewall/firewalld.rb
@@ -121,12 +121,12 @@ module Y2Firewall
 
     # Apply the changes to the modified zones and sets the logging option
     def write
-      only_write
+      write_only
       reload
     end
 
     # Apply the changes to the modified zones and sets the logging option
-    def only_write
+    def write_only
       zones.map { |z| z.apply_changes! if z.modified? }
       api.log_denied_packets = log_denied_packets
       api.default_zone       = default_zone
--- a/library/network/src/lib/y2firewall/firewalld.rb
+++ b/library/network/src/lib/y2firewall/firewalld.rb
@@ -103,7 +103,8 @@ module Y2Firewall
     # @return [Y2Firewall::Firewalld::Service] the recently added service
     def read_service(name)
       service = Y2Firewall::Firewalld::Service.new(name: name)
-      service.create! if !service.supported?
+      raise(Service::NotFound, name) if !service.supported?
+
       service.read
       @services << service
       service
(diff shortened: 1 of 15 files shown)


library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -68,18 +73,42 @@ def read
...
+    end
+
+    # It reads the configuration of the given service or create it from scratch
+    # if not exist. After read adds it to the list of touched services.
+    #
+    # @param name [String] the service name
+    # @return [Y2Firewall::Firewalld::Service] the recently added service
+    def read_service(name)
+      service = Y2Firewall::Firewalld::Service.new(name: name)
+      service.create! if !service.supported?
```
mchf: this looks a bit mysterious ... create a service when not supported?

library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -92,6 +121,12 @@ def modified?
 
     # Apply the changes to the modified zones and sets the logging option
     def write
+      write_only
+      reload
```
mchf: i know that ```write_only``` returns always true ... but may be that ```write_only && reload``` would be safer for the future

library/network/src/lib/y2firewall/firewalld/api/services.rb
```diff
@@ -0,0 +1,100 @@
+# encoding: utf-8
+#
+# ***************************************************************************
+#
+# Copyright (c) 2017 SUSE LLC.
```
mchf: 2018 ;-)

library/network/src/lib/y2firewall/firewalld/api/services.rb
```diff
@@ -0,0 +1,100 @@
...
+#
+# ***************************************************************************
+
+module Y2Firewall
+  class Firewalld
+    class Api
+      # This module contains specific api methods for handling services
+      # definition and configuration.
+      module Services
+        # @param service [String] The firewall service
```
mchf: undocumented optional parameter

library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -42,7 +42,8 @@ module Relations
       #   zone.apply_services_changes!
       #
       # @param args [Array<Symbol] relation or attribute names
-      def has_many(*relations) # rubocop:disable Style/PredicateName
+      def has_many(*relations, scope: nil) # rubocop:disable Style/PredicateName
```
mchf: note for myself. This is at the edge of my ruby knowledge -> study a bit more

library/network/src/lib/y2firewall/firewalld.rb
teclator: I was not sure about creating it or not. Basically not supported means the service is not defined, which implies you have to create it or deploy via a RPM.

So probably is better to raise an exception and report the error as it was in the past.

library/network/src/lib/y2firewall/firewalld/api/services.rb
teclator: :)

library/network/src/lib/y2firewall/firewalld/service.rb
```diff
@@ -0,0 +1,92 @@
...
+      attr_accessor :name
+      attr_accessor :short
+      attr_accessor :description
+
+      has_many :ports, :protocols, scope: "service"
+
+      # Constructor
+      #
+      # @param name [String] zone name
+      def initialize(name: nil)
```
mchf: hm, does it make sense to have a service with nil name?

library/network/src/lib/y2firewall/firewalld/service.rb
```diff
@@ -0,0 +1,92 @@
...
+  class Firewalld
+    # Class to work with Firewalld services
+    class Service
+      extend Relations
+      include Yast::I18n
+      extend Yast::I18n
+
+      attr_accessor :name
+      attr_accessor :short
+      attr_accessor :description
```
mchf: just readers should be enough ...

library/network/test/cwm_firewall_interfaces_test.rb
```diff
@@ -0,0 +1,191 @@
...
+
+      it "initializes the list of network interfaces" do
+        expect(subject).to receive(:InitAllInterfacesList)
+        subject.OpenFirewallInit(widget_settings, "")
+      end
+
+      it "initializes the list of allowed interfaces" do
+        expect(subject).to receive(:InitAllowedInterfaces)
+        subject.OpenFirewallInit(widget_settings, "")
+      end
```
mchf: double checking? the same test twice ...

library/network/src/lib/y2firewall/firewalld/api/services.rb
```diff
@@ -29,6 +29,9 @@ class Api
       # definition and configuration.
       module Services
         # @param service [String] The firewall service
+        # @param permanent [
+        # @param permanent [Boolean] if true it adds the --permanent option the
+        # command to be executed
```
mchf: now it looks a bit messy - a copy paste issue?

library/network/src/lib/y2firewall/firewalld/api/zones.rb
```diff
@@ -93,6 +93,7 @@ def interface_enabled?(zone, interface)
 
         # @param zone [String] The firewall zone
         # @param interface [String] The network interface
+        # @param permanent [Boolean] if true it adds the --permanent option the
```
mchf: missing half of sentense

library/network/src/lib/y2firewall/firewalld/service.rb
teclator: Not really, main idea was in case we allowed to instantiate a new Service and then establish the name, short and description, like allowing a Popup to create it. But in that case the Service should be instantiated with the given name, so does not make sense at all.

library/network/src/lib/y2firewall/firewalld/service.rb
teclator: If we are not going to allow to define them then sure, just read is enough

library/network/src/lib/y2firewall/firewalld/api/services.rb
teclator: ouch, yep, that issue

library/network/test/cwm_firewall_interfaces_test.rb
teclator: It's defined twice? one is for all the interfaces and the other for the allowed ones.

mchf: ah yes. sorry

PR #671, opened 2018-01-16
Fix overlay

jreidinger:
quoting first commit message:

### Fix applying too old backup for overlays (bsc#1070726)

In general problem is that system role selection clears overlay before
creating dialog, but it does it when going back and also when going
forward. So result is that when going back it is properly cleared and
then when going forward when different extensions is chosen, it is again
cleared, but with invalid backup. So this change in the end apply backup
only when going back.

### DIFF
--- a/library/control/src/modules/ProductFeatures.rb
+++ b/library/control/src/modules/ProductFeatures.rb
@@ -324,7 +324,9 @@ module Yast
     # @param features [Hash{String => Hash{String => Object}}] in the same
     #   format as `@features` or `@defaults`
     # @return void
+    # @raise RuntimeError if called twice without {ClearOverlay}
     def SetOverlay(features)
+      raise "SetOverlay called when old overlay was not cleared" if @backup_features
       @backup_features = deep_copy(@features)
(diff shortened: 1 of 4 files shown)


kobliha:
I'm missing a test case

mvidner:
"No description provided." please provide context for the PR.

WHY are we calling ClearOverlay a second time without an intervening SetOverlay? That is not supposed to happen.

mvidner:
So we are at SelectSystemRole (where ClearOverlay is called) and we go Next Back Back Next, right?

jreidinger:
@mvidner yes, see mentioned bug. In fact problematic workflow is:

1. register system
2. no extensions
3. select any role ( so first backup is stored )
4. go back to extension selection ( so first ClearOverlay is called )
5. select basesystem extension
6. select role ( here second ClearOverlay is called which removes extensions and then SetOverlay is called, but with Features that no longer have new extensions )
7. go back ( so now ClearOverlay is called with wrong backup ) and you see broken texts in roles

PR #674, opened 2018-01-24
Firewalld: Added methods for opening ports by YaPI modules.

teclator:
- Needed by https://github.com/yast/http-server
- Also solve a small issue.

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld/zone.rb
+++ b/library/network/src/lib/y2firewall/firewalld/zone.rb
@@ -142,12 +142,16 @@ module Y2Firewall
 
       # Override relation method to be more defensive. An interface can only
       # belong to one zone and the change method remove it before add.
+      #
+      # @param interface [String] interface name
       def add_interface!(interface)
         api.change_interface(name, interface)
       end
 
       # Override relation method to be more defensive. A source can only belong
       # to one zone and the change method remove it before add.
+      #
+      # @param source [String] source address
       def add_source!(source)
         api.change_source(name, source)
       end
--- a/library/network/src/lib/y2firewall/firewalld/zone.rb
+++ b/library/network/src/lib/y2firewall/firewalld/zone.rb
@@ -140,6 +140,18 @@ module Y2Firewall
         }
       end
 
+      # Override relation method to be more defensive. An interface can only
+      # belong to one zone and the change method remove it before add.
+      def add_interface!(interface)
+        api.change_interface(name, interface)
+      end
+
+      # Override relation method to be more defensive. A source can only belong
+      # to one zone and the change method remove it before add.
+      def add_source!(source)
+        api.change_source(name, source)
+      end
+
     private
 
       # Convenience method which return an instance of Y2Firewall::Firewalld
--- a/library/network/src/modules/firewalld_wrapper.rb
+++ b/library/network/src/modules/firewalld_wrapper.rb
@@ -30,29 +30,70 @@ module Yast
   # mainly a firewalld wrapper. It is inteded to be used mostly by YaST
   # modules written in Perl like yast-dns-server.
   class FirewalldWrapperClass < Module
+    include Logger
+
+    VALID_PROTOCOLS = ["udp", "tcp"].freeze
+
+    def initialize
+      Yast.import "PortAliases"
+      Yast.import "PortRanges"
+    end
+
+    # Convenience method for calling firewalld.read
     def read
       firewalld.read
     end
 
+    # Convenience method for calling firewalld.write
     def write
       firewalld.write
     end
 
+    # Convenience method for calling firewalld.write_only
     def write_only
       firewalld.write_only
     end
 
-    def add_port(port_range, protocol, interface)
-      zone = firewalld.zones.find { |z| z.interfaces.include?(interface) }
-      return unless zone
-      port = "#{port_range.sub(":", "-")}/#{protocol.downcase}"
+    # Add the port or range of ports with the given protocol to the zone the
+    # interface belongs to.
+    #
+    # @example
+    #   FirewalldWrapper.add_port("80", "TCP", "eth0")
+    #   FirewalldWrapper.add_port("8080:8090", "TCP", "eth0")
+    #   FirewalldWrapper.add_port("nameserver", "UDP", "eth0")
+    #
+    # @param port_or_range [String] port or range of ports to be added to the zone
+    # @param protocol [String] port protocol
+    # @param interface [String] interface name
+    def add_port(port_or_range, protocol, interface)
+      return false unless validate_port(port_or_range)
+      return false unless supported_protocol?(protocol)
+
+      zone = interface_zone(interface)
+      return false unless zone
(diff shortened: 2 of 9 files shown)


library/network/src/lib/y2firewall/firewalld/zone.rb
```diff
@@ -140,6 +140,18 @@ def export
         }
       end
 
+      # Override relation method to be more defensive. An interface can only
+      # belong to one zone and the change method remove it before add.
+      def add_interface!(interface)
+        api.change_interface(name, interface)
+      end
```
mchf: docu for ```interface```

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -42,9 +42,25 @@ def write_only
       firewalld.write_only
     end
 
-    publish function: :read, type: "boolean (string)"
-    publish function: :write, type: "boolean (string)"
-    publish function: :write_only, type: "boolean (string)"
+    def add_port(port_range, protocol, interface)
```
mchf: public methods deserve documentation

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -42,9 +42,25 @@ def write_only
...
+      zone = firewalld.zones.find { |z| z.interfaces.include?(interface) }
+      return unless zone
+      port = "#{port_range.sub(":", "-")}/#{protocol.downcase}"
```
mchf: According to 1st parameter name, port range is expected. I'd like to see an argument check to be present. I mean something like ```/[0-9]+:[0-9]+/ =~ port_range```

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jan 24 14:13:13 UTC 2018 - knut.anderssen@suse.com
+
+- Firewalld: Added methods to the wrapper class for opening ports
+  by YaPI modules (fate#323460 bsc#1071548 bsc#1076837)
```
mchf: is it really needed? See https://fate.suse.com/323734. @kobliha ?

library/network/src/modules/firewalld_wrapper.rb
teclator: I will document it with examples, just added a quick one to work with http-server. It can be really a por or port_range so it could be 80 or 8080:8090

package/yast2.changes
teclator: That is exactly why added a very very simple wrapper, just a temporal one until it is definitely dropped.

library/network/src/lib/y2firewall/firewalld/zone.rb
teclator: done

library/network/src/modules/firewalld_wrapper.rb
teclator: done

package/yast2.changes
mchf: ok, discussed offline. YaPI is going to be marked as obsolete, but functionall, so we need to keep it working

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -30,29 +30,70 @@ module Yast
   # mainly a firewalld wrapper. It is inteded to be used mostly by YaST
   # modules written in Perl like yast-dns-server.
   class FirewalldWrapperClass < Module
+    include Logger
+
+    VALID_PROTOCOLS = ["udp", "tcp"].freeze
```
mchf: do we need to stick with these protocols only? (note: I came from telecomunications where we used e.g. SCTP quite often ;-)

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -30,29 +30,70 @@ module Yast
...
+    #   FirewalldWrapper.add_port("80", "TCP", "eth0")
+    #   FirewalldWrapper.add_port("8080:8090", "TCP", "eth0")
+    #   FirewalldWrapper.add_port("nameserver", "UDP", "eth0")
+    #
+    # @param port_or_range [String] port or range of ports to be added to the zone
+    # @param protocol [String] port protocol
+    # @param interface [String] interface name
+    def add_port(port_or_range, protocol, interface)
+      return false unless validate_port(port_or_range)
+      return false unless supported_protocol?(protocol)
```
mchf: really NP - a bit inconsistent naming for checkers (```validate_port``` vs ```supported_protocol?```)

library/network/src/modules/firewalld_wrapper.rb
teclator: probably we will support also sctp & dccp but current needs are TPC and UDP ...

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -30,29 +30,70 @@ module Yast
...
+    # Add the port or range of ports with the given protocol to the zone the
+    # interface belongs to.
+    #
+    # @example
+    #   FirewalldWrapper.add_port("80", "TCP", "eth0")
+    #   FirewalldWrapper.add_port("8080:8090", "TCP", "eth0")
+    #   FirewalldWrapper.add_port("nameserver", "UDP", "eth0")
+    #
+    # @param port_or_range [String] port or range of ports to be added to the zone
+    # @param protocol [String] port protocol
```
mchf: number or common port / protocol name (like ssh in the example bellow)

library/network/src/modules/firewalld_wrapper.rb
teclator: added both

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -105,21 +109,29 @@ def remove_port(port_or_range, protocol, interface)
...
+    # Convenience method which returns a singleton instance of
+    # Y2Firewall::Firewalld
+    #
+    # @return [Y2Firewall::Firewalld] singleton instance
     def firewalld
       Y2Firewall::Firewalld.instance
     end
 
-    # Return whether the given port of range of ports is valid
+    # Return whether the given port of range of ports is valid. The port can be
```
mchf: port OR range of ports

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -32,7 +32,7 @@ module Yast
   class FirewalldWrapperClass < Module
     include Logger
 
-    VALID_PROTOCOLS = ["udp", "tcp"].freeze
+    VALID_PROTOCOLS = ["udp", "tcp", "sctp", "dccp"].freeze
```
mchf: well I'd really be careful here. See ```/etc/protocols```

teclator: This four are what firewalld support, this is just a wrapper we really need to support more?

teclator: https://github.com/yast/yast-http-server/blob/f4b0fbd8b20e4f58b3d199473d3d6fc28dbf0b91/src/modules/YaPI/HTTPD.pm#L1276

PR #676, opened 2018-01-24
Cache modified

teclator:
- Goes after #674 and add some improvements for modification comparisons and also add permanent to almost all the api options to avoid mistakes between running & persistent config.

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld/relations.rb
+++ b/library/network/src/lib/y2firewall/firewalld/relations.rb
@@ -44,18 +44,38 @@ module  Y2Firewall
       # @param args [Array<Symbol] relation or attribute names
       def has_many(*relations, scope: nil) # rubocop:disable Style/PredicateName
         scope = "#{scope}_" if scope
+
+        define_method "relations" do
+          relations
+        end
+
+        define_method "apply_all_relations_changes!" do
+          relations.each { |r| public_send("apply_#{r}_changes!") }
+          true
+        end
+
         relations.each do |relation|
           relation_singularized = relation.to_s.sub(/s$/, "")
-          class_eval("attr_accessor :#{relation}")
+          class_eval("attr_reader :#{relation}")
+
+          define_method "#{relation}=" do |item|
+            instance_variable_set("@#{relation}", item)
+
+            @modified << relation unless modified.include?(relation)
+          end
 
           define_method "add_#{relation_singularized}" do |item|
             return public_send(relation) if public_send(relation).include?(item)
 
+            @modified << relation unless modified.include?(relation)
             public_send(relation) << item
           end
 
           define_method "remove_#{relation_singularized}" do |item|
-            return public_send(relation) if public_send(relation).delete(item)
+            if public_send(relation).delete(item)
+              @modified << relation unless modified.include?(relation)
+              return public_send(relation)
+            end
 
             public_send(relation)
           end
@@ -93,8 +113,11 @@ module  Y2Firewall
           end
 
           define_method "apply_#{relation}_changes!" do
+            return unless (modified || []).include?(relation)
             public_send("remove_#{relation}!")
             public_send("add_#{relation}!")
+
+            modified.delete(relation)
           end
         end
       end
--- a/library/network/src/lib/y2firewall/firewalld/relations.rb
+++ b/library/network/src/lib/y2firewall/firewalld/relations.rb
@@ -109,18 +109,18 @@ module  Y2Firewall
           define_method "#{relation}=" do |item|
             instance_variable_set("@#{relation}", item)
 
-            @modified << relation if cache && !modified?(relation)
+            modified!(relation) if cache
           end
 
           define_method "add_#{relation_singularized}" do |item|
             return public_send(relation) if public_send(relation).include?(item)
-            @modified << relation if cache && !modified?(relation)
+            modified!(relation) if cache
             public_send(relation) << item
           end
 
           define_method "remove_#{relation_singularized}" do |item|
             if public_send(relation).delete(item)
-              @modified << relation if cache && !modified?(relation)
+              modified!(relation) if cache
               return public_send(relation)
             end
 
--- a/library/network/src/lib/y2firewall/firewalld/relations.rb
+++ b/library/network/src/lib/y2firewall/firewalld/relations.rb
@@ -18,6 +18,10 @@ module  Y2Firewall
       #
       #   zone = Zone.new
       #
+      #   # Return all the declared relations
+      #   zone.relations #=> [:services]
+      #   # Read all the relations initializing the object
+      #   zone.read_relations
       #   # Adds the "ssh" service into the zone object if not present
       #   zone.add_service("ssh")
       #   # Removes the "ssh" service from the zone object
@@ -40,6 +44,8 @@ module  Y2Firewall
       #   zone.services_to_remove
       #   # Apply the changes (remove_services! && add_services!)
       #   zone.apply_services_changes!
+      #   # Apply all the relations changes
+      #   zone.apply_relations_changes!
       #
       # @param args [Array<Symbol] relation or attribute names
       def has_many(*relations, scope: nil) # rubocop:disable Style/PredicateName
--- a/library/network/src/lib/y2firewall/firewalld/relations.rb
+++ b/library/network/src/lib/y2firewall/firewalld/relations.rb
@@ -49,8 +49,13 @@ module  Y2Firewall
           relations
         end
 
-        define_method "apply_all_relations_changes!" do
+        define_method "apply_relations_changes!" do
           relations.each { |r| public_send("apply_#{r}_changes!") }
+          true
+        end
+
+        define_method "read_relations" do
+          relations.each { |r| instance_variable_set("@#{r}", public_send("current_#{r}")) }
           true
         end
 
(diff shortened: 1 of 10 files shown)


mchf:
a test is failing

library/network/src/lib/y2firewall/firewalld/service.rb
```diff
@@ -86,6 +89,8 @@ def self.modify_ports(name:, tcp_ports: [], udp_ports: [])
       # @param name [String] zone name
       def initialize(name:)
         @name = name
+
+        @modified = []
```
mchf: um, is it ever used for service?

library/network/src/lib/y2firewall/firewalld/zone.rb
```diff
@@ -63,12 +66,26 @@ class Zone
...
         @name = name || api.default_zone
+        @modified = []
       end
 
       def self.known_zones
         KNOWN_ZONES
       end
 
+      # Setter method for masquerade and also add it into modified array if no
+      # present.
```
mchf: ```@param```

library/network/src/lib/y2firewall/firewalld/zone.rb
```diff
@@ -77,24 +94,20 @@ def full_name
...
       def apply_changes!
-        apply_interfaces_changes!
-        apply_services_changes!
-        apply_ports_changes!
-        apply_protocols_changes!
-        apply_sources_changes!
+        return true unless modified?
+
+        apply_relations_changes!
+        if modified.include?(:masquerade)
```
mchf: do you expect that modified will contain something else than :modified ever?

library/network/src/lib/y2firewall/firewalld/zone_parser.rb
```diff
@@ -70,6 +70,8 @@ def parse
           end
         end
 
+        zones.map { |z| z.modified = [] }
```
mchf: this breaks encapsulation a bit. Something like ```z.reset``` looks better to me

library/network/src/lib/y2firewall/firewalld/service.rb
teclator: Yes as service is also using relations.

library/network/src/lib/y2firewall/firewalld/zone.rb
teclator: what do you mean?, in modified you will have something like [:services, :interfaces] so the rest of the attributes and relations don't have to be touched avoiding api calls to check what was changed or avoiding calls to change them.

library/network/src/lib/y2firewall/firewalld/zone_parser.rb
teclator: MM, not sure if reset at all.. because maybe reset could mean other logic. Basically the thing is that when are calling the setter method we mark the zone as modified, but in this case it shouldn't as it is basically a complete read.

mchf: well I don't care about name much. I'm just against setting ```z.modified = []```

teclator: Added untouched! method to reset modifications

library/network/src/lib/y2firewall/firewalld/zone.rb
```diff
@@ -63,12 +66,27 @@ class Zone
...
       def self.known_zones
         KNOWN_ZONES
       end
 
+      # Setter method for enabling masquerading.
+      #
+      # @param enabled [Boolean] true for enable; false for disable
+      # @return [Boolean] whether it is enabled or not
+      def masquerade=(enable)
+        @modified << :masquerade unless @modified.include?(:masquerade)
```
mchf: same or very similar code is at multiple places, so it might deserve own method

library/network/src/lib/y2firewall/firewalld/service.rb
```diff
@@ -56,6 +56,9 @@ def initialize(name)
       # @return description [String] service long description
       attr_reader :description
 
+      # @return modified [Array<Symbol>] the list of relations modified
+      attr_reader :modified
+
       has_many :ports, :protocols, scope: "service"
```
mchf: just a note about the model for me
this should cover that a service represented by an object of this class (http server)
can run on multiple ports (say 80, 8080)
and support multiple protocols (say http, https)

library/network/src/lib/y2firewall/firewalld/service.rb
mchf: yes. I got into that metaprogramming stuff a bit an see now.

mchf: just a note, 
shouldn't  the stuff around "modified" be part of that metaprogramming block? It is used mainly there anyway.

library/network/src/lib/y2firewall/firewalld/zone.rb
teclator: Ok, added the cache option directly into the `Relations` module. It can be enable directly when the relations are declared just passing the argument `use_cache: true`

```ruby
 class Zone
   extend Relations
   has_many :interfaces, use_cache: true

   def initialize
     @modified = []
   end
 end
```

library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -3,21 +3,61 @@ class Firewalld
     # Extends the base class with metaprogramming methods which defines some
     # attributes common logic.
     module Relations
+      def enable_modifications_cache
+        class_eval("attr_accessor :modified")
```
imobachgs: Are you sure you need a `class_eval` call here? I guess that calling to `attr_accessor` should work.

library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -3,21 +3,61 @@ class Firewalld
...
+      # case it is important to initialize objects with the instance variable
+      # @modified as an empty array.
+      #
       # @example
       #
       #   class Zone
       #     extend Relations
       #
-      #     has_many :services
+      #     has_many :services, use_cache: true
```
imobachgs: np: Why not just `cache`? `cache: true` looks quite expressive.

library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -40,22 +80,56 @@ module Relations
...
+
+        define_method "read_relations" do
+          relations.each { |r| instance_variable_set("@#{r}", public_send("current_#{r}")) }
+          true
+        end
+
         relations.each do |relation|
           relation_singularized = relation.to_s.sub(/s$/, "")
-          class_eval("attr_accessor :#{relation}")
+          class_eval("attr_reader :#{relation}")
```
imobachgs: sure about `class_eval`?

library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -40,22 +80,56 @@ module Relations
...
+
+          define_method "#{relation}=" do |item|
+            instance_variable_set("@#{relation}", item)
+
+            if use_cache
```
imobachgs: I guess you could combine both conditions.

library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -40,22 +80,56 @@ module Relations
...
+              @modified << relation unless modified.include?(relation)
+            end
             public_send(relation) << item
           end
 
           define_method "remove_#{relation_singularized}" do |item|
-            return public_send(relation) if public_send(relation).delete(item)
+            if public_send(relation).delete(item)
+              if use_cache
```
imobachgs: Same than above.

library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -93,8 +167,13 @@ def has_many(*relations, scope: nil) # rubocop:disable Style/PredicateName
           end
 
           define_method "apply_#{relation}_changes!" do
+            if use_cache
```
imobachgs: Same than above.

library/network/src/lib/y2firewall/firewalld/service.rb
```diff
@@ -86,6 +86,7 @@ def self.modify_ports(name:, tcp_ports: [], udp_ports: [])
       # @param name [String] zone name
       def initialize(name:)
         @name = name
+        @modified = []
```
imobachgs: Strictly speaking, that's not needed anymore, right?

library/network/src/lib/y2firewall/firewalld/relations.rb
teclator: Not needed

library/network/src/lib/y2firewall/firewalld/relations.rb
teclator: Not needed

library/network/src/lib/y2firewall/firewalld/service.rb
teclator: It was a defensive initialization, just want to avoid exceptions with `<<` or `+=` calls if modified was not called.

mchf:
lgtm

PR #677, opened 2018-01-25
Checking if firewalld has been installed before calling it

schubi2:
The problem is that many modules are using CWMFirewallInterfaces. This will be often already done
by including it by starting the different module. At that point it could be that the package has not been installed at all.

e.g.: 

 https://github.com/yast/yast-ftp-server/blob/master/src/clients/ftp-server_auto.rb#L30

  * 2d3b25c9 checking if the package is available


teclator:
It should be fixed by #674 

That is what will be shown:

## When no installed (Avoid initialization or handling of anything)

firewallnotconfigurable

## When some service is not supported

notfoundservice

## If all is fine

summaryofchannges

schubi2:
This is not needed anymore.

PR #680, opened 2018-01-30
Fix double render in TreePager (bsc#1078212)

ancorgs:


library/cwm/src/lib/cwm/pager.rb
```diff
@@ -35,7 +35,8 @@ def handle(event)
 
       return nil unless page
 
-      return nil if @current_page.widget_id == new_id
+      # Don't alter this line without taking a look to bsc#1078212
+      return nil if @current_page.widget_id == page.widget_id
```
jreidinger: to be honest I still do not get why this is wrong? new_id should be id of page as page_for_id search for page widget.id

jreidinger: now I read bug and it is clear for me, but we really should document it here and not just in bug

PR #686, opened 2018-02-01
Y2packager integration

lslezak:
- Move some code from `yast2-packager` to `yast2` - it is required in `Product.rb` module
- It cannot depend on `yast2-packager` because it would create `yast2` <-> `yast2-packager` circular dependency
- Note: I'm keeping the `Y2Packager` namespace as the `Product` class is currently defined  in `Packages::Product` and `Y2Packager::Product`. These should be unified/merged later.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -2,7 +2,7 @@
 Thu Feb  1 08:09:28 UTC 2018 - lslezak@suse.cz
 
 - Move some Y2Packager classes from yast2-packager here to avoid
-  circular dependency (related to fate#1076513)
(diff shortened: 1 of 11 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Feb  1 08:09:28 UTC 2018 - lslezak@suse.cz
+
+- Move some Y2Packager classes from yast2-packager here to avoid
+  circular dependency (related to fate#1076513)
```
gilsonsouza: Please, check fate number.

PR #687, opened 2018-02-01
Import y2packager stuff

imobachgs:
After moving some classes from yast2-packager (#686) the build started to fail. We needed to move some more classes and stub `Language` and `Proxy` modules.


imobachgs:
Thanks!

PR #691, opened 2018-02-08
Turned Relation#relations into class method

mchf:
bcs #has_many is used as class method (e.g.) in Zone it makes sense to have possibility to query defined relations at the same level

For an example see https://github.com/yast/yast-firewall/pull/48

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld/relations.rb
+++ b/library/network/src/lib/y2firewall/firewalld/relations.rb
@@ -88,7 +88,7 @@ module  Y2Firewall
         scope = "#{scope}_" if scope
         enable_modifications_cache if cache
 
-        define_method "relations" do
+        define_singleton_method "relations" do
           relations
         end
 
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Feb  8 09:21:19 UTC 2018 - mfilka@suse.com
+
+- fate#323460
+  - slightly improved consistency of new Y2Firewall API
+- 4.0.50 
+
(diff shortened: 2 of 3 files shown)


library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -88,7 +88,7 @@ def has_many(*relations, scope: nil, cache: false) # rubocop:disable Style/Predi
         scope = "#{scope}_" if scope
         enable_modifications_cache if cache
 
-        define_method "relations" do
+        define_singleton_method "relations" do
```
teclator: Well, in case we need it, I would prefer to have both.

```ruby
f = Y2Firewall::Firewalld.instance
external = f.find_zone("external")
external.relations
=> [:services, :interfaces, :protocols, :ports, :sources]
Y2Firewall::Firewalld::Zone.relations
=> [:services, :interfaces, :protocols, :ports, :sources]
```

mchf: Hm, I was thinking of it from the opposite side. When we have (conceptually)

```
class Zone
  def self.has_many( *relations, ....)
end
```
then it would be nice to query these relations back when analyzing connections between classes. So I've modified that to
```
class Zone
  def self.has_many( *relations, ....)
  def self.relations
end
```
Simple use case is in the pointed yast2-firewall PR (I know that that is not crucial one, but i wanted to learn / play with meta programming a bit ;-)

So, just for curiosity do you have a use case for your case in your mind or do you know / expect any? In my POV it can be easily implemented using delegator - something like ```def_delegator self.class, relations``` What do you think?

PR #703, opened 2018-02-27
do not use gettext on examples

jreidinger:


library/cwm/examples/object_api.rb
```diff
@@ -17,7 +17,7 @@ def initialize
   end
 
   def label
-    _("Lucky number")
+    "Lucky number"
```
lslezak: Maybe should at least add a FIXME comment saying that this text should be marked for translation when used in production code. Developers usually do just copy&paste :wink:, this could avoid some mistakes in that case.

library/cwm/examples/object_api.rb
```diff
@@ -47,8 +47,6 @@ class ExampleDialog
     include Yast::I18n
     include Yast::UIShortcuts
     def run
-      textdomain "example"
```
lslezak: The same here, something like `# FIXME: to have translations set a text domain here`

PR #704, opened 2018-02-27
Firewalld: Verify that the configuration has been read before writing

schubi2:
- https://bugzilla.suse.com/show_bug.cgi?id=1082827

  * 1ee398d5 packaging
  * 29c57748 taking correct value for --set-log-denied

### DIFF
--- a/library/network/test/y2firewall/firewalld_test.rb
+++ b/library/network/test/y2firewall/firewalld_test.rb
@@ -262,7 +262,7 @@ describe Y2Firewall::Firewalld do
         allow(zone).to receive(:modified?).and_return(modified_zone)
       end
       firewalld.zones = empty_zones
-      firewalld.log_denied_packets = "on"
+      firewalld.log_denied_packets = "all"
     end
 
     context "when some of the attributes have been modified since read" do
@@ -358,7 +358,7 @@ describe Y2Firewall::Firewalld do
 
     let(:api) do
       instance_double(Y2Firewall::Firewalld::Api,
-        log_denied_packets: "on",
+        log_denied_packets: "all",
         default_zone:       "work",
         list_all_zones:     zones_definition,
         zones:              known_zones)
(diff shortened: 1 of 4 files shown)


library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -262,7 +262,7 @@
         allow(zone).to receive(:modified?).and_return(modified_zone)
       end
       firewalld.zones = empty_zones
-      firewalld.log_denied_packets = true
+      firewalld.log_denied_packets = "on"
```
teclator: In that case it should be some of: all, unicast, broadcast, multicast

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -357,7 +358,7 @@
 
     let(:api) do
       instance_double(Y2Firewall::Firewalld::Api,
-        log_denied_packets: true,
+        log_denied_packets: "on",
```
teclator: some as above

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -378,7 +379,7 @@
       expect(config).to be_a(Hash)
       expect(config["enable_firewall"]).to eq(false)
       expect(config["start_firewall"]).to eq(true)
-      expect(config["log_denied_packets"]).to eq(true)
+      expect(config["log_denied_packets"]).to eq("on")
```
teclator: same as above

teclator:
@schubi2 the bug in the description helps to identify the problem, could you attach it?. Also, the issue if I'm not wrong is because the value has not been initialized at all, isn't it?

kobliha:
Just a question: what about the schema? Do we need to adapt it too? IMO we might need so. And the documentation as well, but that's a different PR.

teclator:
@kobliha the problems arise when there is no firewall section.

schubi2:
The docu is in well shape and the rnc file is checking for text only because the values are depending on the return value of the firewall-offline-cmd call.

teclator:
@schubi2 I think that calling read during writing unless read? should be enough

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Feb 27 15:19:25 CET 2018 - schubi@suse.de
+
+- Nis configuration fails while executing firewall-offline-cmd.
```
imobachgs: np: NIS :)

PR #707, opened 2018-03-06
Find correct release notes URL during upgrade

imobachgs:


library/packages/src/lib/y2packager/product.rb
```diff
@@ -243,6 +243,18 @@ def release_notes(user_lang, format = :txt)
...
+    # Return release notes URL
+    #
+    # Release notes might not be defined in libzypp and this method returns the URL
+    # to get release notes from.
+    #
+    # @return [String,nil] Release notes URL or nil if it is not defined.
+    def relnotes_url
+      return nil unless resolvable_properties
+      url = resolvable_properties["relnotes_url"]
+      url.is_a?(String) && url.empty? ? nil : url
```
jreidinger: this is a bit hard to read and construct all cases like url nil, url empty string and url real url, so what about more readable:
```
case url
when nil, "" then nil
when String then url
else raise "Unexpected value from libzypp #{url.inspect}"
end
```

imobachgs: To be honest, it looks overkill to me (I would not expect anything than nil or a string). What about this?

```ruby
return nil if url.nil? || url.empty?
url
```

Does it look better?

library/packages/src/lib/y2packager/product.rb
```diff
@@ -256,5 +268,18 @@ def status?(*statuses)
         statuses.include?(res["status"])
       end
     end
+
+    # Return product's resolvable properties
+    #
+    # Only the 'name' and 'version' will be used to find out whether the
+    # product properties, ignoring the architecture, vendor or any other
```
jreidinger: I do not get this sentence. What means that "whether the product properties" I miss something there

jreidinger:
looks good in general, just two notes.

library/packages/src/lib/y2packager/product.rb
lslezak: Even `nil` should not be returned: https://github.com/yast/yast-pkg-bindings/blob/3d396421f16edd6375f6b26047c97499c847787e/src/Resolvable_Properties.cc#L341, there is `YCPString()` value.

But that `nil` check is still acceptable, just in case...

imobachgs: Then I would even make it simpler.

PR #709, opened 2018-03-09
fix parsing values with multiple signs (bsc#1081353)

jreidinger:


library/system/test/kernel_test.rb
```diff
@@ -147,4 +149,79 @@
...
+    context "when command line contain newline" do
+      let(:cmdline) { "splash=verbose silent\n" }
+
+      it "deletes this newline from CmdLine" do
+        subject.ParseInstallationKernelCmdline
+        expect(subject.GetCmdLine).to eq " splash=verbose silent"
+      end
+    end
+
+    context "when command line contain newline" do
```
joseivanlopez: Duplicated tests.

mvidner:
With multiple signs? as in `- - -1 == -1` ??

jreidinger:
@mvidner see tests. I add test explicitelly for customer test case.

mvidner:
"see tests" is absolutely not an acceptable commit message. Also, there is no .changes entry. I have to insist on a meaningful entry that a person different from you can understand.

mvidner:
Now it makes sense, thanks!
Since you added good test coverage, IMHO it would be great to remove the remaining zombies in the affected code, but I don't insist on that.

jreidinger:
@mvidner well, it is L3 patch, so for SLE12 I prefer to keep it. For SLE15 I plan to do more improvements.

PR #710, opened 2018-03-09
Fix cmdline parsing master

jreidinger:


mvidner:
This is a merge of #709

library/system/src/modules/Kernel.rb
```diff
@@ -161,92 +161,6 @@ def AddCmdLine(name, arg)
...
-              AddCmdLine(key, value)
-            end
-          end
-        end
-      end
-
-      nil
-    end
-
     def ParseInstallationKernelCmdline
```
mvidner:
```rb
    # Parse the installation-time kernel command line
    # - the `vga` parameter is separated, see {#GetVgaType}
    # - some specific parameters are ignored
    # - the rest is passed on to @cmdLine for which {#GetCmdLine} is a reader
    # @return [void]
```

library/system/src/modules/Kernel.rb
```diff
@@ -661,6 +575,31 @@ def InformAboutKernelChange
       @inform_about_kernel_change
     end
 
+    publish function: :AddCmdLine, type: "void (string, string)"
+    publish function: :GetVgaType, type: "string ()"
+    publish function: :SetVgaType, type: "void (string)"
```
mvidner: "FIXME: is heer because of bootloader module, should be removed" and it is not used anymore!

library/system/src/modules/Kernel.rb
```diff
@@ -731,30 +670,93 @@ def read_modules_to_load
...
+
+      discardlist
+    end
+
+    # @param	[String] cmdline to parse
+    #
+    # @return	[void]
+    # Filters out yast2 specific boot parameters and sets
+    # Parameters to the important cmdline parts.
+    def ExtractCmdlineParameters(line)
```
mvidner: As this is (not yet, but it should) a private method too, I would much prefer a cleaner calling sequence. But it is a NTH, an encapsulated implementation detail.

```rb
def ParseInstallationKernelCmdline
  cmd_string = if Mode... ...; end
  log.info ... HidePasswords(cmd_string)
  cmd_list = tokenize(cmd_string)            # renamed list_of_params
  to_reject = params_to_discard(cmd_list)
  cmd_list.reject! { |c| params_to_discard.include?(c) }
  assign_params(cmd_list)                    # renamed rest of ExtractCmdlineParameters
end

File: library/system/src/modules/Kernel.rb [hunk 4]
```diff
@@ -731,30 +670,93 @@ def read_modules_to_load
...
+      discardlist = []
+      # some systems (pseries) can autodetect the serial console
+      if cmdlist.include?("AUTOCONSOLE")
+        discardlist << "console"
+        discardlist << "AUTOCONSOLE"
+      end
+
+      # add special key filtering for s390
+      # bnc#462276 Extraneous parameters in /etc/zipl.conf from the installer
+      discardlist << "User" << "init" << "ramdisk_size" if Arch.s390
```
mvidner: These would be better as constant arrays IMHO

library/system/src/modules/Kernel.rb
```diff
@@ -731,30 +670,93 @@ def read_modules_to_load
...
-    publish function: :reset_modules_to_load, type: "void ()"
-    publish function: :modules_to_load, type: "map <string, list> ()"
+    # constructs list of keys to discard from command line
+    # @param [Array<String>] list of command line entries
+    # @return [Array<String>] list of keys to discard
+    def params_to_discard(cmdlist)
+      discardlist = []
+      # some systems (pseries) can autodetect the serial console
+      if cmdlist.include?("AUTOCONSOLE")
+        discardlist << "console"
```
mvidner:
```rb
# Note:  `console` is the only value that depends on the input argument.
```

mvidner:
@jreidinger Thanks for the cleanup! The main question is: is the cmdline parsing specified somewhere. I mean the quoting.

jreidinger:
@mvidner I try to find it, but does not find any reasonable resource. Maybe my google query is just bad.

jreidinger:
@mvidner well, it does not hurt now and we also pick up our own parameters where can be relaxed demands. I am not fan of removing such functionality in RC2 just to find that something obscure really use it.

library/system/src/modules/Kernel.rb
```diff
@@ -731,30 +670,93 @@ def read_modules_to_load
...
+      while parse_index < line.size
+        current_char = line[parse_index]
+        in_quotes = !in_quotes if current_char == "\"" && !after_backslash
+        if current_char == " " && !in_quotes
+          cmdlist << current_param
+          current_param = ""
+        else
+          current_param << current_char
+        end
+        # does kernel cmdline can have double backslash?
```
mvidner: For the in-kernel parser, a backslash is a regular character.
For this parser, it is a "stupid escape": the first backslash does not escape the second one.

library/system/test/kernel_test.rb
```diff
@@ -147,4 +149,75 @@
...
+      end
+
+      it "sets vgaType" do
+        subject.ParseInstallationKernelCmdline
+        expect(subject.GetVgaType).to eq "ask"
+      end
+    end
+
+    context "when command line value in quotes" do
+      let(:cmdline) { " splash=\"verbose\\\", silent\" text=\\\"abc\n" }
```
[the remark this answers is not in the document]
mvidner: Please remove this test, we don't want to keep this. Only test that a space in quotes does not split params, like this:
```rb
  let(:cmdline) { "confuse=\"foo vga=ask bar\"" }
...
  expect(subject.GetCmdLine).to eq "confuse=\"foo vga=ask bar\""
```

library/system/src/modules/Kernel.rb
mvidner: If you removed SetCmdLine then  SetVgaType should be removed too.

library/system/src/modules/Kernel.rb
mvidner: Please replace the comment by the above, to ease the future cleanup

library/system/test/kernel_test.rb
```diff
@@ -147,4 +149,85 @@
...
+    context "when command line value including quotes" do
+      let(:cmdline) { " splash=\"verbose, silent\" text=abc\n" }
+
+      it "it adds it to CmdLine correctly" do
+        subject.ParseInstallationKernelCmdline
+        expect(subject.GetCmdLine).to eq " splash=\"verbose, silent\" text=abc"
+      end
+    end
+
+    # FIXME: keep old functionality, but probably kernel does not allow it.
```
mvidner:
```rb
# FIXME: we do backslash escapes but in fact
# the kernel does not care about any backslashes, only about double quotes
```

PR #714, opened 2018-03-23
Fix timed error any question

jreidinger:
tested manually. example can be also used to test old behavior if used without Y2DIR.


package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Mar 23 09:17:01 UTC 2018 - jreidinger@suse.com
+
+- fix behavior of showing timed error popup ( found during
```
joseivanlopez: NP: space between "( found"

PR #716, opened 2018-03-27
Make possible to use Yast2::Popup from Report

ancorgs:
Part of https://trello.com/c/T5rxuHfg/300-2-sles15-p1-1082542-build4731storage-ng-delete-lvmlv-failed-when-reinstall-on-encrypted-lvm

See it in action in a yast-storage-ng callback:
report_yesno

The same popup using the timeout from `Report`
report_yesno_timeout

### DIFF
--- a/library/general/src/modules/Report.rb
+++ b/library/general/src/modules/Report.rb
@@ -40,6 +40,8 @@ module Yast
   # @TODO not all methods respect all environment, feel free to open issue with
   #   method that doesn't respect it.
   class ReportClass < Module
+    include Yast::Logger
+
     def main
       textdomain "base"
 
@@ -436,6 +438,8 @@ module Yast
     #   - The argument :timeout will be ignored, the timeout fixed in the Report
     #   module will always be used instead (again, see {#DisplayYesNoMessages}).
     #   - The button ids must be :yes and :no, to honor the Report API.
+    #   - Due to the previous point, if no :buttons argument is provided, the
+    #   value :yes_no will be used for it.
     #
     # Like any other method used to present yesno_messages, false is always
     # returned if the system is configured to not display messages, no matter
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Report.rb
```diff
@@ -422,6 +423,44 @@ def ErrorAnyQuestion(headline, message, yes_button_message, no_button_message, f
...
+    # delegates drawing the pop-up to {Yast2::Popup.show}, in case the message
+    # must be presented to the user (which can be configured via
+    # {#DisplayYesNoMessages}).
+    #
+    # All the arguments are forwarded to #{Yast2::Popup.show} almost as-is, but
+    # some aspects must be observed:
+    #
+    #   - The argument :timeout will be ignored, the timeout fixed in the Report
+    #   module will always be used instead (again, see {#DisplayYesNoMessages}).
+    #   - The button ids must be :yes and :no, to honor the Report API.
```
jreidinger: to be honest, I expect that here we should use by default `buttons: :yes_no` so if you omit this param, you still get yes_no popup. Default `:ok` looks wrong for me. So in other words, when you call `if Report.yesno_popup("text")` it will always return false, as it return `:ok` and not `:yes`

So here should be only note that if buttons was overwritten, only `:yes` symbol is considered for bootlean return value.

jreidinger:
can you please add screenshot when timeout is used? Just to check how it looks like.

library/general/src/modules/Report.rb
```diff
@@ -422,6 +423,44 @@ def ErrorAnyQuestion(headline, message, yes_button_message, no_button_message, f
...
+    # what was selected as the focused default answer.
+    #
+    # @param [String] message Popup message, forwarded to {Yast2::Popup.show}
+    # @param [Hash] extra_args Extra options to be forwarded to {Yast2::Popup.show},
+    #   see description for some considerations
+    # @return [Boolean] True if :yes is pressed, otherwise false
+    def yesno_popup(message, extra_args = {})
+      # Use exactly the same y2milestone call than other yesno methods
+      Builtins.y2milestone(1, "%1", message) if @log_yesno_messages
+
```
jreidinger: maybe log here warning if extra_args include timeout?

ancorgs: done

library/general/src/modules/Report.rb
ancorgs: Done. `:yes_no` is now used by default.

About `:yes` being considered, that's already documented in the `@return`.

jreidinger:
@ancorgs thanks for patch. Funny part is that I few days ago fix fir storage another Report method which is broken and now it also use new Popup - https://github.com/yast/yast-yast2/pull/714

PR #717, opened 2018-03-28
[WIP] Added ProductUpgrade class (bsc#1086259)

lslezak:
- **Still WIP**
- It finds the base product for upgrade.


library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -120,6 +132,38 @@ def available_products
...
+    # @return [Array<Hash>] pkg-bindings data structures
+    def available_products
+      # remove e.g. installed products
+      zypp_products.select { |p| p["status"] == :available || p["status"] == :selected }
+    end
+
+    # read the installed products
+    # @return [Array<Hash>] pkg-bindings data structures
+    def installed_products
+      # remove e.g. available or removed products
```
jreidinger: I do not get why removed when it is selected?

library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -120,6 +132,38 @@ def available_products
...
+      # and because a symlink can point only to one file there can be only one base product.
+      # The "status" conditition is actually not required because that symlink is created
+      # only for installed products. (Just make sure it still works in case the libzypp
+      # internal implementation is changed.)
+      base = products.find do |p|
+        p["type"] == "base" && (p["status"] == :installed || p["status"] == :removed)
+      end
+
+      log.info("Found installed base product: #{base}")
+      base
```
jreidinger: why not simple `installed_products.find { |p| p.type == "base" }` ?

library/packages/src/lib/y2packager/product_upgrade.rb
```diff
@@ -0,0 +1,171 @@
...
+        product
+      end
+
+      # find the upgrade product from the fallback mapping
+      # @param available [Array<Y2Packager::Product>] the available base products
+      # @return [Y2Packager::Product,nil] the new upgraded product
+      def find_by_mapping(available)
+        installed = Y2Packager::Product.installed_products
+
+        # sort the keys by length, try more products first
```
jreidinger: it is good that you document what code did, but I miss important info, WHY? why more products have to be done earlier?

library/packages/src/lib/y2packager/product_upgrade.rb
```diff
@@ -0,0 +1,171 @@
...
+        upgrade = MAPPING.keys.sort_by(&:size).find do |keys|
+          keys.all? { |name| installed.any? { |p| p.name == name } }
+        end
+
+        log.info("Found fallback upgrade for products: #{upgrade.inspect}")
+        return nil unless upgrade
```
jreidinger: should not be those two lines switched? log says Found, but it can be nil.

library/packages/src/lib/y2packager/product_upgrade.rb
```diff
@@ -0,0 +1,171 @@
...
+        # store the current resolvable states
+        Yast::Pkg.SaveState
+
+        # run the solver in the upgrade mode
+        Yast::Pkg.PkgUpdateAll({})
+        log_products
+
+        product = Y2Packager::Product.selected_base
+        # save the solver test case for easier debugging if no product upgrade was found
+        Yast::Pkg.CreateSolverTestCase("/var/log/YaST2/solver-product-upgrade") unless product
```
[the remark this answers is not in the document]
lslezak: Actually it's a directory, libzypp saves some XML files with data dump. It's the same as using `Extras` -> `Generate Dependency Resolver Test Case` menu in the Qt package manager or using the `--debug-solver` zypper option.

library/packages/src/lib/y2packager/product_reader.rb
lslezak: Because `Y2Packager::Product` does not have `type` attribute... :worried:

jreidinger: but installed_products is from `zypp_products` which is `Pkg` products, not `Y2Packager::Product`

jreidinger: I think you are hit by your own method names, because here is two installled_products, one in reader and second in Product class and both return different kind of object, which is a bit confusing.

lslezak: Oh, you are right, it's `Product.installed_products` vs. `ProductReader#installed_products`... :worried:

library/packages/src/lib/y2packager/product_upgrade.rb
lslezak: I have added a comment with the reason.

PR #718, opened 2018-04-04
License API refactoring

imobachgs:
This PR simplifies the License API. Cleaning up and offering a way to add arbitrary licenses are the missing pieces.

/cc @teclator


library/packages/src/lib/y2packager/product_license.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+
+      private def cache
+        @cache ||= {}
+      end
+    end
+
+    # Constructor
+    #
+    # @param product_name [String] Product name to get licenses for
```
teclator: `@param license`

library/packages/src/lib/y2packager/license.rb
```diff
@@ -12,40 +12,99 @@
...
 
+    # @return [Y2Packager::LicensesFetchers::Base] License fetcher object
+    attr_reader :fetcher
+
     alias_method :accepted?, :accepted
 
+    class << self
+      # Find a license for a given product
+      #
+      # @param product_name [String]    Product's name
```
jreidinger: well, please be specific which name you mean. Is it pretty name, short name or id ? it is often confusing which one to use.

library/packages/src/lib/y2packager/license.rb
```diff
@@ -12,40 +12,99 @@
...
+      # @param source       [:rpm,:url] Source to get the license from. For the time being,
+      #   only :rpm is really supported.
```
jreidinger: missing documentation of possible options.

library/packages/src/lib/y2packager/license.rb
```diff
@@ -12,40 +12,99 @@
...
+    # @return [Y2Packager::LicensesFetchers::Base] License fetcher object
+    attr_reader :fetcher
+
     alias_method :accepted?, :accepted
 
+    class << self
+      # Find a license for a given product
+      #
+      # @param product_name [String]    Product's name
+      # @param source       [:rpm,:url] Source to get the license from. For the time being,
```
jreidinger: well, I am not fun of this kind of hidden polymorphysm. Why this method need to know what is source of license? It should find license independently where it came from.

library/packages/src/lib/y2packager/license.rb
```diff
@@ -12,40 +12,99 @@
...
+      #   only :rpm is really supported.
+      # @return [License]
```
jreidinger: please document that it can also return nil, when license is not found

library/packages/src/lib/y2packager/license.rb
```diff
@@ -12,40 +12,99 @@
...
+    # License unique identifier
+    #
+    # This identifier is based on the given default language translation.
+    #
+    # @return [String,nil] Unique identifier; nil if the license was not found.
+    def id
+      return @id if @id
+      content = content_for(DEFAULT_LANG)
+      return unless content
+      @id = Digest::MD5.hexdigest(content_for(DEFAULT_LANG))
```
jreidinger: is there some trick why not use `content` and instead used again call to `content_for(DEFAULT_LANG)`

library/packages/src/lib/y2packager/license.rb
imobachgs: I see your point (I do not like it either). But at this point of time, we need to support different uses cases for backward compatibility (like, for instance, retrieving the license elsewhere in the code). In the future we should be able to determine which fetcher/provider to use automatically.

Additionaly, we will drop :url for the now.

library/packages/src/lib/y2packager/license.rb
imobachgs: No, fortunatelly it is just an error :)

library/packages/src/lib/y2packager/license.rb
```diff
@@ -12,40 +12,99 @@
...
+    # It will return the empty string ("") if the license does not exist
+    #
     # @param [String] language
     # @return [String,nil] the license translated content or nil if not found
     def content_for(lang = DEFAULT_LANG)
-      translations[lang]
+      return @translations[lang] if @translations[lang]
+      content = fetcher.license_content(lang)
+      return add_content_for(lang, content) if content
+      content_for(DEFAULT_LANG) unless lang == DEFAULT_LANG
```
jreidinger: so now behavior changed and it return untranslated license if translation not found? quite surprising behavior. Be sure to document it. especially when it now return nil when not found.

library/packages/src/lib/y2packager/license.rb
```diff
@@ -12,40 +12,99 @@
...
+    end
+
+    # FIXME: Probably the locales should be obtained through the licenses
+    # translations, and probably could be initialized the first time a license
+    # is instantiated.
+    def locales
+      fetcher.license_locales
+    end
+
+    def license_confirmation_required?
```
jreidinger: I suggest name `confirmation_required?` license_ prefix is not needed in license class

library/packages/src/lib/y2packager/licenses_fetchers.rb
```diff
@@ -11,21 +11,18 @@
...
+  module LicensesFetchers
+    # Return the licenses proper fetcher for a given source
     #
-    # @param language [String]
-    # @param content [String] of the license for the given language
-    def initialize(language:, content:)
-      @lang = language
-      @content = content
+    # @param source [Symbol] :rpm or :url
+    def self.for(source, product_name, options = {})
```
jreidinger: As I mention above I think nicer API would be detect automatically source. Just try to find default lang for all known source and what finds first, then use such fetcher.

library/packages/src/lib/y2packager/licenses_fetchers/dummy.rb
```diff
@@ -0,0 +1,49 @@
...
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+require "y2packager/licenses_fetchers/base"
+
+module Y2Packager
+  module LicensesFetchers
+    # FIXME: Finish implementation
+    class Dummy < Base
```
jreidinger: what is goal of this Dummy implementation? Should it act as NullObject pattern?

library/packages/src/lib/y2packager/licenses_handlers/dummy.rb
```diff
@@ -0,0 +1,38 @@
...
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+module Y2Packager
+  module LicensesHandlers
+    class Dummy
```
jreidinger: same question here, what is goal of it?

library/packages/src/lib/y2packager/licenses_fetchers.rb
imobachgs: Sure, that API would be nicer. But at this point, we were requested to keep changes under control and, in `registration` and `packager` modules, the license content is obtained using different mechanisms.

In the future, that code should be moved to its own fetchers/handlers and, using proper objects, we could autodetect the right source and adapt to it accordingly.

We will write some documentation about what's expected in the future. For instance, a method like `ProductLicense#find_or_create` should not exist, but it's there just to make the transition from the old code.

library/packages/src/lib/y2packager/licenses_fetchers/dummy.rb
imobachgs: As commented before, just a "transitional" class that should be replaced with something better for SP1.

imobachgs: I wanted to keep the same internal API (the license asks its fetcher always about the content). As an alternative, we could use a different approach and just keep a `@content` variable in the licenses that would be use when no fetcher is defined at all.

library/packages/src/lib/y2packager/product.rb
```diff
@@ -229,7 +224,7 @@ def license_confirmed?
     #
     # @return [Array<String>] Language codes ("de_DE", "en_US", etc.)
     def license_locales
-      license_reader.locales
+      license.locales
```
jreidinger: in Dummy it is `license_locales` but I like more this short variant

library/packages/src/lib/y2packager/product_license.rb
```diff
@@ -0,0 +1,93 @@
...
+module Y2Packager
+  class ProductLicense
+    # This class holds the license stuff for a given product
+    #
+    # Why a separate ProductLicense class? First of all, we wanted to extract
+    # the license handling from Y2Packager::Product and moving this logic to
+    # Y2Packager::License was not a good idea because different products could
+    # share the same license. Additionally, this class offers an API to work
+    # with licenses with a proper Product or Addon object is not available
+    # (backward compatibility reasons).
```
jreidinger: after reading this, It still come to my mind and especially when I see that delegators, why ProductLicense is not subclass of License?

library/packages/src/lib/y2packager/product_license.rb
```diff
@@ -0,0 +1,93 @@
...
+    #   Determine whether the license should be accepted or not
+    #   @return [Boolean] true if the license acceptance is required
+    #
+    # @!method license_confirmation=(confirmed)
+    #   Set the license confirmation for the product
+    #   @param confirmed [Boolean] true if it should be accepted; false otherwise
+    def_delegators :@handler, :license_confirmation_required?, :license_confirmation
+
+    # @return [License] Product's license
+    attr_reader :license
```
jreidinger: this is a bit strange, as delegators is used so it looks like private inheritance

library/packages/src/lib/y2packager/product_license.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    class << self
+      # Find license for a given product
+      #
+      # @param product_name [String]    Product's name
+      # @param source       [:rpm,:url] Source to get the license from. For the time being,
+      #   only :rpm is really supported.
+      # @return [ProductLicense]
+      def find(product_name, source = :rpm, options = {})
```
jreidinger: to be honest, always when I see two default params, I get to my mind that keyword params would be better.

jreidinger: also this `options = {}` is everywhere and smell a bit for me. I think it was intended as possible future extension of functionality, but with keyword params, you can also easily add it and do not need to have holder.

library/packages/src/lib/y2packager/licenses_fetchers/dummy.rb
jreidinger: It is fine this way. I just maybe call it `Static` and not dummy. It is static fetcher, that have always same content.

library/packages/src/lib/y2packager/license.rb
teclator: same opinion here, I also get confused with this return

library/packages/src/lib/y2packager/license.rb
```diff
@@ -12,40 +12,99 @@
...
+      private def cache
+        @cache ||= {}
+      end
+    end
+
     # Constructor
     #
     # @param options [Hash<String, String>]
-    def initialize(options = {})
+    def initialize(fetcher)
```
teclator: Of course I'm also polluted with previous implementation, but I would expect to be able to instantiate a new License with its own content directly instead of looking for it. So would prefer named arguments like fetcher and content. Then the dummy fetcher could be used if a fetcher is not given with the given content

library/packages/src/lib/y2packager/product_license.rb
```diff
@@ -0,0 +1,91 @@
...
+    attr_reader :license
+
+    class << self
+      # Find license for a given product
+      #
+      # This method uses a cache to return an already fetched product license.
+      #
+      # @param product_name [String]    Product's name
+      # @param source       [:rpm,:url] Source to get the license from. For the time being,
+      #   only :rpm is really supported.
```
jreidinger: when only rpm is suported, why not having it as default instead of `nil`?

library/packages/src/lib/y2packager/product_license.rb
```diff
@@ -0,0 +1,91 @@
...
+      private def cache
+        @cache ||= {}
+      end
+    end
+
+    # Constructor
+    #
+    # @param product_name [String] Product name to get licenses for
+    # @param source       [Symbol] Backend to use when syncing the licenses acceptance status
+    #   (only :rpm is supported)
```
jreidinger: please document default behavior for nil. As it looks like symbol is mandatory.

imobachgs:
I've made some changes:

* Now we try to sync the libzypp acceptance status.
* I've renamed :rpm to :libzypp because it is more accurate.

library/packages/src/lib/y2packager/product_license.rb
```diff
@@ -81,11 +81,39 @@ def initialize(product_name, license, source: nil)
...
+    # @return [Boolean] true if the license has been accepted; false otherwise.
+    def accept!
+      license.accept!
+      sync_acceptance
+      license.accepted?
+    end
+
+    # Reject the license
+    #
+    # @return [Boolean] true if the license has been accepted; false otherwise.
```
jreidinger: this is a bit confusing. you return true if operation failed?

PR #719, opened 2018-04-05
Use SHA2 instead of MD5

imobachgs:
Replace MD5 with SHA1 when determining whether a license was accepted or not,
as requested in [fate#325461](https://fate.suse.com/325461).

### DIFF
--- a/library/packages/src/lib/y2packager/license.rb
+++ b/library/packages/src/lib/y2packager/license.rb
@@ -110,7 +110,7 @@ module Y2Packager
       return @id if @id
       content = content_for(DEFAULT_LANG)
       return unless content
-      @id = Digest::MD5.hexdigest(content)
+      @id = Digest::SHA1.hexdigest(content)
     end
 
     # Return the license translated content for the given language
--- a/library/packages/src/lib/y2packager/license.rb
+++ b/library/packages/src/lib/y2packager/license.rb
@@ -110,7 +110,7 @@ module Y2Packager
       return @id if @id
       content = content_for(DEFAULT_LANG)
       return unless content
-      @id = Digest::SHA1.hexdigest(content)
+      @id = Digest::SHA2.hexdigest(content)
     end
 
     # Return the license translated content for the given language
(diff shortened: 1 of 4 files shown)


jreidinger:
Seriously? What is probability that such unintentional collision on hashes happen? But speed difference is for sure. let me try to measure it

jreidinger:
OK, I test performance wise and also check FATE. BTW SHA1 is also cryptographic weak and second idea is that we do not need cryptographic result. So I add to test also crc32.

how test look like:

```ruby
require "benchmark"

require "zlib"
require "digest"

n = 50000
teststring = "abcd" * 1000
Benchmark.bm(7) do |x|
  x.report("md5:")  { n.times { Digest::MD5.hexdigest(teststring) } }
  x.report("sha1:") { n.times { Digest::SHA1.hexdigest(teststring) } }
  x.report("sha2:") { n.times { Digest::SHA2.hexdigest(teststring) } }
  x.report("crc:")  { n.times { Zlib.crc32(teststring) } }
end

puts Digest::MD5.hexdigest(teststring)
puts Digest::SHA1.hexdigest(teststring)
puts Digest::SHA2.hexdigest(teststring)
puts Zlib.crc32(teststring)
```

and what is result on my pc?:

```
              user     system      total        real
md5:      0.336595   0.000000   0.336595 (  0.336633)
sha1:     0.246459   0.000000   0.246459 (  0.246477)
sha2:     0.522378   0.003894   0.526272 (  0.546068)
crc:      0.161141   0.000000   0.161141 (  0.161143)
69086d3a75d23ba5b2ce21f7c4c3055b
894aa55a08ddb4e0e94fa50348ddd5fa3ea58a3f
a794d3322e58f529258e0d9331aa9a17eece6e05848aa89fe1e6ffa006231957
766008115
```

So for me result is that if they are fine with crc32 as mentioned in fate, we should use it. Even I am still curious about real probability of finding collision.

jreidinger:
Here is article about that possible hash collision:

http://preshing.com/20110504/hash-collision-probabilities/

imobachgs:
AFAIK, hashing the ID is only a temporary solution. Eventually, we will come up with proper ids. The sha1 looks quite good in terms of speed (we will hash only a few values, one for each license).

PR #722, opened 2018-04-06
Update dependency

joseivanlopez:
Part of PBI https://trello.com/c/jItiacnG/335-readd-system-wide-lock-in-libstorage-ng


joseivanlopez:
I am not sure if it is necessary to update this dependency, because in general, I see that the packages should include `yast2-ruby-bindings` directly.

Closing for now.

PR #724, opened 2018-04-10
Do not use the solver for finding the best product upgrade (bsc#1086734)

lslezak:
- Do not use the solver for finding the best product upgrade
- It does not work correctly in the SLES + sle-module-hpc => SLES_HPC case as the solver finds the new SLES and selects the SLES => SLES upgrade.
- Use the only product available or search from the hardcoded list
- This is a quick fix for SLE15-GA, for SLE15-SP1 we should find a better solution (avoid hardcoding)
- See https://bugzilla.suse.com/show_bug.cgi?id=1086734
- Tested manually in a SLES12-SP3 + sle-module-hpc upgrade to SLE15, see the screenshots below
- Additionally I have added also the openSUSE => SLES upgrade support (not tested)
- 4.0.67

### Original Behavior

Originally the standard SLES product was selected for upgrade as you can see in the license agreement dialog:

sles_hpc_upgrade_license_agreement_broken

### After Applying the Fix

With the fix the SLES_HPC product is displayed:

sles_hpc_upgrade_license_agreement

And later the migration to SLES_HPC is offered:

sles_hpc_upgrade_migration_selection

(For some reason two same migrations are offered, but that's an unrelated issue and probably on  the SCC side, I'll check later...)

### DIFF
--- a/library/packages/src/lib/y2packager/product_upgrade.rb
+++ b/library/packages/src/lib/y2packager/product_upgrade.rb
@@ -20,7 +20,7 @@ module Y2Packager
 
     Yast.import "Pkg"
 
-    # fallback mapping with upgraded products to handle some corner cases,
+    # mapping with upgraded products to handle some corner cases,
     # maps installed products to a new base product
     MAPPING = {
       # SLES12 + HPC module => SLESHPC15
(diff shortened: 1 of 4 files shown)


library/packages/src/lib/y2packager/product_upgrade.rb
```diff
@@ -33,7 +33,9 @@ class ProductUpgrade
       # SLED11 => SLED15
       ["SUSE_SLED"]              => "SLED",
       # SLES4SAP11 => SLES4SAP15
-      ["SUSE_SLES_SAP"]          => "SLES_SAP"
+      ["SUSE_SLES_SAP"]          => "SLES_SAP",
+      # openSUSE => SLES
+      ["openSUSE"]               => "SLES"
```
jreidinger: So what happen when you upgrade with leap?

lslezak: Then the SLES will not be available and this line won't match. The mapping is `[installed products]` => `available product`.

I'll improve the comment.

PR #725, opened 2018-04-12
Fallback hicolor (bsc#1087224)

teclator:
This PR replaces #616 using a branch from the same repository for being able to collaborate pushing to it.

https://bugzilla.suse.com/show_bug.cgi?id=1087224

## When the icon is not found (tested modifiying the Icon name in the desktop file manually)
yasticon

## When the icon is found
lanicon

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Apr 12 12:01:49 UTC 2018 - knut.anderssen@suse.com
+
+- Wizard: Fall back to smaller and/or hicolor icon if the icon for
+  the launched module can not be found (bsc#1087224)
+- 4.0.68
+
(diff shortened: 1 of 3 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Apr 12 12:01:49 UTC 2018 - knut.anderssen@suse.com
+
+- Wizard: Fall back to smaller and/or hicolor icon if the icon for
+  the launched module can not be found (bsc#1087224)
```
jreidinger: just small note, I think that it should be together `cannot`

Hacklin: Correct, it's `cannot` here.
Officially you only use `can not` if the `not` is part of the word after it; for example as `not only` in `can not only`.
However. `can not` is acceptable.

PR #726, opened 2018-04-12
Handle I/O errors in DoneProvide callback

imobachgs:
Fixes [bsc#1088682](https://bugzilla.suse.com/show_bug.cgi?id=1088682).

### DIFF
--- a/library/packages/src/modules/PackageCallbacks.rb
+++ b/library/packages/src/modules/PackageCallbacks.rb
@@ -216,10 +216,13 @@ module Yast
       end
 
       message =
-        if error == 3
-          Builtins.sformat(_("Package %1 is broken, integrity check has failed."), name)
-        elsif error == 2
+        case error
+        when 2
           Builtins.sformat(_("Package %1 could not be downloaded (input/output error)."), name)
+        when 3
+          Builtins.sformat(_("Package %1 is broken, integrity check has failed."), name)
+        else
+          log.warn "DoneProvide: unknown error '#{error}'"
         end
 
       # IO/INVALID
--- a/library/packages/src/modules/PackageCallbacks.rb
+++ b/library/packages/src/modules/PackageCallbacks.rb
@@ -215,13 +215,16 @@ module Yast
         return "C"
       end
 
-      # INVALID
-      if error == 3
+      message =
+        if error == 3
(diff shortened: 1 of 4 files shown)


library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -215,13 +215,16 @@ def DoneProvide(error, reason, name)
         return "C"
       end
 
-      # INVALID
-      if error == 3
+      message =
+        if error == 3
```
jreidinger: please use case here and add also default where you log unknown error code, so we know when something new appear.

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -286,9 +289,8 @@ def DoneProvide(error, reason, name)
             if @showLongInfo
               error_symbol = "ERROR"
 
-              if error == 1
-                error_symbol = "NOT_FOUND"
-              elsif error == 2
+              # https://github.com/openSUSE/libzypp/blob/8dda46306f06440e1acaefb36fb60f6ce909fd42/zypp/ZYppCallbacks.h#L106
```
jreidinger: I do not get why you remove that 1 as it is still valid according to URL you paste

jreidinger: and as bonus, using case again would be really nice, with default branch so we know that libzypp extend their enum.

imobachgs: You only will get to this line when `error` is 2 or 3. The `1` case is handled through the MediaChange callback (see c9816a1cf7 and 546dcfb363). But I am fine with having it back.

jreidinger: then please mention it also in comment, as link you paste indicate that also "1" should be handled, so explain why it is not needed.

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -215,13 +215,19 @@ def DoneProvide(error, reason, name)
         return "C"
       end
 
-      # INVALID
-      if error == 3
+      message =
+        case error
+        when 2
```
lslezak: It would be probably a good idea to add a comment like "The 1 case is handled in the MediaChange callback" comment here.

imobachgs:
Thanks!

PR #729, opened 2018-04-19
[WIP] Apply firewalld changes directly into the running config

teclator:


teclator:
Closing it as it is finally not needed

PR #730, opened 2018-04-24
Helper to wrap text into direction markers

imobachgs:
Related pull requests:

* https://github.com/yast/yast-packager/pull/348
* https://github.com/yast/yast-installation/pull/687

### DIFF
--- a/library/general/src/lib/ui/text_helpers.rb
+++ b/library/general/src/lib/ui/text_helpers.rb
@@ -56,11 +56,13 @@ module UI
 
     # Wrap a given text in direction markers
     #
-    # @param [String] text to be wrapped
+    # @param [String] text to be wrapped. This text may contain tags and they
+    #   will not be escaped
     # @param [String] language code (it gets the current one by default)
     # @return [String] wrapped text
-    def direct_richtext(text, lang = Yast::Language.language)
+    def div_with_direction(text, lang = nil)
       Yast.import "Language"
+      lang ||= Yast::Language.language
       direction = "ltr"
       direction = "rtl" if lang.start_with?("ar") || lang.start_with?("he")
       "<div dir=\"#{direction}\">#{text}</div>"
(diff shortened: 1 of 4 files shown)


library/general/src/lib/ui/text_helpers.rb
```diff
@@ -53,5 +53,17 @@ def wrap_text(text, wrap = 76, separator: " ", prepend_text: "",
...
       lines.join("\n")
     end
+
+    # Wrap a given text in direction markers
+    #
+    # @param [String] text to be wrapped
+    # @param [String] language code (it gets the current one by default)
+    # @return [String] wrapped text
+    def direct_richtext(text, lang = Yast::Language.language)
+      Yast.import "Language"
```
mvidner: I think this is too late for the parameter declaration.
This method is at an intersection of Language and Text but with more of the language, so maybe we should move it to a place that already import Language?

library/general/src/lib/ui/text_helpers.rb
```diff
@@ -53,5 +53,17 @@ def wrap_text(text, wrap = 76, separator: " ", prepend_text: "",
 
       lines.join("\n")
     end
+
+    # Wrap a given text in direction markers
+    #
+    # @param [String] text to be wrapped
+    # @param [String] language code (it gets the current one by default)
+    # @return [String] wrapped text
```
mvidner: We should say (and test) that *text* may contain tags which will be passed through (not escaped)

library/general/src/lib/ui/text_helpers.rb
```diff
@@ -53,5 +53,17 @@ def wrap_text(text, wrap = 76, separator: " ", prepend_text: "",
...
+    def direct_richtext(text, lang = Yast::Language.language)
```
mvidner: This name would confuse me at the call site.
What about `div_with_direction` ?

library/general/src/lib/ui/text_helpers.rb
imobachgs: Or we might just move the initialization to the line below.

```ruby
div_with_direction(text, lang = nil)
  Yast.import "Language"
  lang ||= Yast::Language.language
  #...
end
```

PR #732, opened 2018-05-08
Add handlers

joseivanlopez:
Needed for PBI https://trello.com/c/Lqos3IvN/238-2-partitioner-fix-abort-and-back-buttons


library/cwm/test/dialog_test.rb
```diff
@@ -32,5 +38,23 @@ def contents
 
       expect(subject.class.run).to eq(:launch)
     end
+
+    it "uses the back handler" do
```
ancorgs: Does this actually test if the back handler is used? Is not the test itself the one calling it?

PR #736, opened 2018-05-15
Replace deprecated firewallctl command

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1093111

firewalldstate

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld/api.rb
+++ b/library/network/src/lib/y2firewall/firewalld/api.rb
@@ -103,7 +103,11 @@ module Y2Firewall
         offline? ? run_command("--disable") : Yast::Service.Disable("firewalld")
       end
 
-      # @return [Boolean] The firewalld service state (exit code)
+      # Return the current state of the firewalld service (running or not
+      # running)
+      #
+      # @return [String] firewalld service state
+      # @see http://www.firewalld.org/documentation/man-pages/firewall-cmd.html
       def state
         case Yast::Execute.on_target("firewall-cmd", "--state", allowed_exitstatus: [0, 252])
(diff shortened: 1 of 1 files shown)


library/network/src/lib/y2firewall/firewalld/api.rb
```diff
@@ -105,7 +105,7 @@ def disable!
 
       # @return [Boolean] The firewalld service state (exit code)
```
mvidner: Let's fix the API docs to match the implementation

teclator:
It was reported for TW but checking that in SLES we already ship firewalld 5.2 then it probably make sense to change the branch to GA

library/network/src/lib/y2firewall/firewalld/api.rb
```diff
@@ -105,7 +105,7 @@ def disable!
...
       def state
-        case Yast::Execute.on_target("firewallctl", "state", allowed_exitstatus: [0, 252])
+        case Yast::Execute.on_target("firewall-cmd", "--state", allowed_exitstatus: [0, 252])
```
mvidner: (I very much like the small change but I have no idea if it is correct. I really should get my own 15.0 for testing.)

teclator: I have it and tested manually.

teclator: Will adapt the documentation:

http://www.firewalld.org/documentation/man-pages/firewall-cmd.html

teclator: firewalldstate

PR #738, opened 2018-05-23
merged with sles15

schubi2:


schubi2:
Assuming that is has been merged meanwhile in another point

PR #743, opened 2018-06-25
[WIP] add API to connect service and socket

jreidinger:


library/systemd/src/modules/systemd_service.rb
```diff
@@ -70,12 +72,14 @@ class SystemdServiceClass < Module
     include Yast::Logger
 
     UNIT_SUFFIX = ".service".freeze
+    SERVICE_PROPMAP = SystemdUnit::DEFAULT_PROPMAP.merge(trigerred_by: "TriggeredBy")
```
imobachgs: two typos: *trigerred* should be *triggered*.

jreidinger: ah, thanks for catch. The second one (the string) is systemd name, so it is correct one.

PR #747, opened 2018-06-28
[RFC] Socket widget

jreidinger:
Old Service Status
snimek obrazovky_2018-07-03_09-19-24

Old CWM AutoStart
snimek obrazovky_2018-07-03_09-24-29

new one with all values
snimek obrazovky_2018-07-03_09-25-29
snimek obrazovky_2018-07-03_09-26-27
snimek obrazovky_2018-07-03_09-27-08

### DIFF
--- a/library/systemd/src/lib/yast2/service_configuration.rb
+++ b/library/systemd/src/lib/yast2/service_configuration.rb
@@ -79,11 +79,11 @@ module Yast2
       @action
     end
 
-    ACTIONS = [ :start, :stop, :reload, :restart, :nothing].freeze
+    ACTIONS = [:start, :stop, :reload, :restart, :nothing].freeze
     # sets target action for services
     # @param action[:start, :stop, :restart, :reload, :nothing] for possible values and
     #  its explanation please see return value of {target_action}
-    def action= (action)
+    def action=(action)
       if !ACTIONS.include?(action)
         raise ArgumentError, "Invalid parameter #{action.inspect}"
       end
@@ -108,7 +108,7 @@ module Yast2
       @autostart
     end
 
-    AUTOSTART_OPTIONS = [:on_boot, :on_demand, :manual, :inconsistent]
+    AUTOSTART_OPTIONS = [:on_boot, :on_demand, :manual, :inconsistent].freeze
     # sets autostart configuration.
     # @param [:on_boot, :on_demand, :manual, :inconsistent] autostart
     # configuratio. For explanation please see {autostart} when
@@ -148,16 +148,16 @@ module Yast2
     def read_autostart
       sockets = @services.map(&:socket).compact
       services_without_socket = @services.reject(&:socket)
-      if sockets.all?(&:enabled?) && services_without_socket.all?(&:enabled?)
-        @autostart = :on_demand
+      @autostart = if sockets.all?(&:enabled?) && services_without_socket.all?(&:enabled?)
+        :on_demand
       elsif @services.all?(&:enabled?)
-        @autostart = :on_boot
+        :on_boot
       elsif sockets.none?(&:enabled?) && @services.none?(&:enabled?)
-        @autostart = :manual
+        :manual
       else
-        @autostart = :inconsistent
+        :inconsistent
       end
-    rescue SystemctlError =>e
+    rescue SystemctlError => e
       log.error "systemctl failure: #{e.inspect}"
       @autostart = :unknown
     end
@@ -166,11 +166,11 @@ module Yast2
       case autostart
       when :on_boot then @services.each { |s| s.start_mode = :boot }
       when :on_demand
-        @services.each do |service|
-          if s.start_modes.include?(:demand)
-            s.start_mode = :demand
+        @services.each do |_service|
+          s.start_mode = if s.start_modes.include?(:demand)
+            :demand
           else
-            s.start_mode = :boot
+            :boot
           end
         end
       when :manual then @services.each { |s| s.start_mode = :manual }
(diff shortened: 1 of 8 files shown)


library/systemd/src/lib/yast2/service_configuration.rb
```diff
@@ -0,0 +1,194 @@
...
+  # Its common usage is with {Yast2::ServiceWidget} which allows to modify
+  # configuration and write it when user approve whole dialog.
+  class ServiceConfiguration
+    include Yast::Logger
+
+    # services managed by this configuration
+    attr_reader :services
+
+    # creates new configuration that holds state for given services
+    # @param services<Yast::SystemdServiceClass::Service> services to configure
```
imobachgs: Is this the documentation format correct? I guess it should be `services [<Yast::SystemdServiceClass::Service>] ...`

library/systemd/src/lib/yast2/service_configuration.rb
```diff
@@ -0,0 +1,194 @@
...
+      end
+
+      @services = services
+    end
+
+    # reads system status of services. Can be also used to reread current status.
+    # @raise <Yast::SystemctlError> when read services state failed
+    def read
+      # safe initial state when exception happen
+      @status = :unknown
```
imobachgs: Why not moving this initialization to the `initialize` method?

library/systemd/src/lib/yast2/service_configuration.rb
```diff
@@ -0,0 +1,194 @@
...
+    def action= (action)
+      if !ACTIONS.include?(action)
+        raise ArgumentError, "Invalid parameter #{action.inspect}"
+      end
+
+      @action = action
+    end
+
+    # returns currently set autostart configuration. If it is not yet set it
+    # calls read and return current system state.
```
imobachgs: returns

library/systemd/src/lib/yast2/service_configuration.rb
```diff
@@ -0,0 +1,194 @@
...
+    # @return [:on_boot, :on_demand, :manual, :inconsistent, :unknown] autostart
+    #   configuration:
+    #
+    #   - `:on_boot` all services start during boot.
```
imobachgs: I know it is a different thing, but what about using `:boot` instead of `:on_boot` and `:demand` instead of `:on_demand` to keep consistency? Or just the other way around: I could adjust the `SystemdService` API.

library/systemd/src/lib/yast2/service_configuration.rb
```diff
@@ -0,0 +1,194 @@
...
+    end
+
+    # returns true if any service support reload
+    def support_reload?
+      # TODO: implement it
+      true
+    end
+
+    # returns true if any service has socket start
+    def support_on_demand?
```
imobachgs: And what if some of the services does not support loading on demand? Can we safely assume that on demand is supported for the whole group?

library/systemd/src/lib/yast2/service_configuration.rb
```diff
@@ -0,0 +1,194 @@
...
+      @services.any?(&:socket?)
+    end
+
+  private
+
+    def read_status
+      services_active = @services.map { |s| s.active? }
```
imobachgs: You can write `@services.map(&:active?)` just for consistency.

library/systemd/src/lib/yast2/service_configuration.rb
jreidinger: yes, my fault. I still trying more to precise resulting design and document intentions and then test and document it properly.

library/systemd/src/lib/yast2/service_configuration.rb
jreidinger: because I though that for read it can take some time. So it is more consistent with common yast workflow when read and write have own progress. So you can inform about reading services and also showing progress for it.

library/systemd/src/lib/yast2/service_configuration.rb
jreidinger: well, for me it make sense in such case to start socket for services that support it and start on boot rest, so it will work. But in general I do not expect it to be common case. Probably only one I know now is iscsi client.

library/systemd/src/lib/yast2/service_configuration.rb
jreidinger: right

library/systemd/src/lib/yast2/service_configuration.rb
joseivanlopez: I guess Imo means to only move the assignments, not the read* calls.

joseivanlopez: But moving it to `initialize` makes more difficult to check whether `read` has been already called.

imobachgs: Ok, that's a valid point (you want to know whether `read` was called or not).

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -0,0 +1,165 @@
...
+        _("Unknown")
+      else
+        raise "Unknown status #{service_configuration.status.inspect}"
+      end
+    end
+
+    def action_widget
+      ComboBox(
+        Id(:service_widget_action),
+        "After writting settings:",
```
mvidner: @jreidinger 
1. the correct spelling is "wri**t**ing", with one **t** (because that makes for an open, long syllable: wri-ting, pronounced  [rai-tiŋ], whereas "writ-ting" is a closed, short syllable [ri-tiŋ]. There *are* some sensile rules in English spelling)
2. translation marker missing

PR #748, opened 2018-06-29
Extend service API

imobachgs:
This PR extends `SystemdService` API to properly support sockets. Now, instead of using `enable/disable`, we can specify the `start_mode` which can be `:boot`, `:demand` and `:manual`. Additionally, the API supports querying which modes are supported for a given service.

### DIFF
--- a/library/systemd/src/modules/systemd_service.rb
+++ b/library/systemd/src/modules/systemd_service.rb
@@ -174,6 +174,8 @@ module Yast
 
       # Returns socket associated with service or nil if there is no such socket
       def socket
+        return @socket if @socket
+
         # not triggered
         socket_name = properties.triggered_by
         return unless socket_name
@@ -181,7 +183,7 @@ module Yast
         socket_name = socket_name[/\S+\.socket/]
         return unless socket_name # triggered by non-socket
 
-        Yast::SystemdSocket.find(socket_name)
+        @socket = Yast::SystemdSocket.find(socket_name)
       end
 
       alias_method :enable_service, :enable
--- a/library/systemd/src/modules/systemd_service.rb
+++ b/library/systemd/src/modules/systemd_service.rb
@@ -184,7 +184,105 @@ module Yast
         Yast::SystemdSocket.find(socket_name)
       end
 
+      alias_method :enable_service, :enable
+      private :enable_service
+
+      # Enable a service
+      #
+      # @param mode [Symbol] Start mode (:boot or :demand).
+      # @return [Boolean] true if the service was successfully enabled; false otherwise.
+      def enable
+        self.start_mode = :boot
+      end
+
+      alias_method :disable_service, :disable
+      private :disable_service
+
+      # Disable a service
+      #
+      # If the service has an associated socket, it is disabled too.
+      #
+      # @return [Boolean] true if the service was successfully disabled; false otherwise.
+      def disable
+        self.start_mode = :manual
+      end
+
+      alias_method :enabled_on_boot?, :enabled?
+      private :enabled_on_boot?
+
+      # Determine whether the service is enabled or not
+      #
+      # The service can be enable to be started on boot or on demand.
+      #
+      # @return [Boolean] true if the service is enabled; false otherwise.
+      def enabled?
+        start_mode != :manual
+      end
+
+      # Determine whether the service has an associated socket
+      #
+      # @return [Boolean] true if an associated socket exists; false otherwise.
+      def socket?
+        !socket.nil?
+      end
+
+      # Return the start mode
+      #
+      # @return [Symbol] Start mode (:boot, :demand, :manual)
+      def start_mode
+        return :boot if enabled_on_boot?
+        return :demand if enabled_on_demand?
+        :manual
+      end
+
+      # Set the service start mode
+      #
+      # See {#start_modes} to find out the supported modes for a given service (usually :boot,
+      # :manual and, in some cases, :demand).
+      #
+      # @see #start_modes
+      def start_mode=(mode)
+        if !start_modes.include?(mode)
+          log.warn "Invalid start mode: '#{mode}' for service '#{name}'"
+          return
+        end
+
+        case mode
+        when :boot
+          enable_service
+          socket.disable
+        when :demand
+          disable_service
+          socket.enable
+        when :manual
+          disable_service
+          socket.disable
+        end
+      end
+
+      # Return the list of supported start modes
+      #
+      # * :boot:   The service will be started when the system boots.
+      # * :manual: The service is disabled and it will be started manually.
+      # * :demand: The service will be started on demand (using a Systemd socket).
+      #
+      # @return [Array<Symbol>] List of supported modes.
+      def start_modes
+        return @start_modes if @start_modes
+        @start_modes = [:boot, :manual]
+        @start_modes.insert(1, :demand) if socket?
+        @start_modes
+      end
+
     private
+
+      # Determine whether the service is enabled on demand
+      #
+      # @return [Boolean] true if it is enabled on demand; false otherwise.
+      def enabled_on_demand?
+        return false unless socket
+        socket.enabled?
+      end
 
       def installation_system?
         File.exist?(START_SERVICE_INSTSYS_COMMAND)
--- a/library/systemd/test/systemd_service_test.rb
+++ b/library/systemd/test/systemd_service_test.rb
@@ -130,6 +130,201 @@ module Yast
       end
     end
 
+    describe "#enabled?" do
+      subject(:service) { SystemdService.find("cups") }
+
+      before do
+        allow(service).to receive(:start_mode).and_return(start_mode)
+        stub_services(service: "cups")
+      end
+
+      context "when the start mode is :boot" do
+        let(:start_mode) { :boot }
+
+        it "returns true" do
+          expect(service).to be_enabled
+        end
+      end
+
+      context "when the start mode is :demand" do
+        let(:start_mode) { :demand }
+
+        it "returns true" do
+          expect(service).to be_enabled
+        end
+      end
+
+      context "when the start mode is :manual" do
+        let(:start_mode) { :manual }
+
+        it "returns false" do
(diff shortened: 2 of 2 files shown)


library/systemd/src/modules/systemd_service.rb
```diff
@@ -174,18 +174,118 @@ def restart
 
       # Returns socket associated with service or nil if there is no such socket
       def socket
+        return @socket if @socket
```
jreidinger: this will work only if socket is not found. But I expect that check for triggered_by is quick enough. What I do not like is that returned socket is cached for long time and it still has old status. Maybe at least refresh_properties should be called?

library/systemd/src/modules/systemd_service.rb
```diff
@@ -174,18 +174,118 @@ def restart
...
+      def socket?
+        !socket.nil?
+      end
+
+      # Return the start mode
+      #
+      # @return [Symbol] Start mode (:boot, :demand, :manual)
+      def start_mode
+        return :boot if enabled_on_boot?
+        return :demand if enabled_on_demand?
```
jreidinger: similar issue I am solving in my class. What is start mode when service and also socket is enabled? I agree that boot makes sense, but we should document it.

library/systemd/src/modules/systemd_service.rb
```diff
@@ -174,18 +174,118 @@ def restart
...
+        :manual
+      end
+
+      # Set the service start mode
+      #
+      # See {#start_modes} to find out the supported modes for a given service (usually :boot,
+      # :manual and, in some cases, :demand).
```
jreidinger: I do not see it documented in start_mode.

library/systemd/src/modules/systemd_service.rb
```diff
@@ -174,18 +174,118 @@ def restart
...
+      #
+      # @see #start_modes
+      def start_mode=(mode)
+        if !start_modes.include?(mode)
+          log.warn "Invalid start mode: '#{mode}' for service '#{name}'"
```
jreidinger: agh...no, please no silent ignore of invalid value. Just use `ArgumentError` like any other nice method.

library/systemd/src/modules/systemd_service.rb
jreidinger: ah, I see start_mode and not start_modes

library/systemd/src/modules/systemd_service.rb
```diff
@@ -174,18 +174,118 @@ def restart
...
+
+      # Return the list of supported start modes
+      #
+      # * :boot:   The service will be started when the system boots.
+      # * :manual: The service is disabled and it will be started manually.
+      # * :demand: The service will be started on demand (using a Systemd socket).
+      #
+      # @return [Array<Symbol>] List of supported modes.
+      def start_modes
+        return @start_modes if @start_modes
```
jreidinger: I see here one risk. We use `find` for socket and not `find!` so when due to some temporary error it is not found or some systemctl failed, it will never can use on demand.

library/systemd/src/modules/systemd_service.rb
imobachgs: `start_modes` (see the final 's'). However, I was thinking about calling it `supported_start_modes` or something like that to be more clear.

library/systemd/src/modules/systemd_service.rb
jreidinger: I see here one risk. We use `find` for socket and not `find!` so when due to some temporary error it is not found or some systemctl failed, it will never can use on demand.

library/systemd/src/modules/systemd_service.rb
imobachgs: Yes, it is `:boot`. I will document it.

library/systemd/src/modules/systemd_service.rb
```diff
@@ -174,18 +174,118 @@ def restart
...
       end
 
     private
 
+      # Determine whether the service is enabled on demand
+      #
+      # @return [Boolean] true if it is enabled on demand; false otherwise.
+      def enabled_on_demand?
+        return false unless socket
+        socket.enabled?
```
jreidinger: why not `socket && socket.enabled?` and then separated method is probably not needed?

library/systemd/src/modules/systemd_service.rb
imobachgs: I just wanted to avoid calling systemctl several times. For instance, if we do something like:

```ruby
service.socket.disable if service.socket?
```

We would be calling `#socket` twice. Having some kind of `reload` might help (which could be implemented calling `#find` again).

library/systemd/test/systemd_service_test.rb
```diff
@@ -130,6 +130,201 @@ module Yast
...
+      subject(:service) { SystemdService.find("cups") }
+      let(:socket) { double("socket", disable: true) }
+
+      before do
+        stub_services(service: "cups")
+        allow(service).to receive(:socket).and_return(socket)
+        allow(service).to receive(:disable)
+      end
+
+      context "when no argument is given" do
```
jreidinger: ugh, what? it does not reflect test.

library/systemd/src/modules/systemd_service.rb
imobachgs: I just wanted to avoid asking `systemctl` more than necessary. IMHO, it looks really like a corner case. I would expect more problems if systemctl is failing (even temporarily).

library/systemd/test/systemd_service_test.rb
```diff
@@ -130,6 +130,201 @@ module Yast
...
+    end
+
+    describe "#enable" do
+      subject(:service) { SystemdService.find("cups") }
+
+      before do
+        stub_services(service: "cups")
+      end
+
+      it "sets start_mode to :boot" do
```
jreidinger: I am not sure if this is good test and documention. You want to be sure that when you call enable that it "enable service". This is important and not some internal method for start_mode.

library/systemd/test/systemd_service_test.rb
imobachgs: Oh, a leftover.

library/systemd/src/modules/systemd_service.rb
```diff
@@ -174,18 +174,118 @@ def restart
...
+      alias_method :enabled_on_boot?, :enabled?
+      private :enabled_on_boot?
+
+      # Determine whether the service is enabled or not
+      #
+      # The service can be enable to be started on boot or on demand.
+      #
+      # @return [Boolean] true if the service is enabled; false otherwise.
+      def enabled?
+        start_mode != :manual
```
joseivanlopez: I don't understand this method. `enabled?` definition is based on `start_mode`, but `start_mode` definition is also based on `enabled?`. Probably I am missing something.

library/systemd/test/systemd_service_test.rb
```diff
@@ -130,6 +130,201 @@ module Yast
...
+    end
+
+    describe "#disable" do
+      subject(:service) { SystemdService.find("cups") }
+
+      before do
+        stub_services(service: "cups")
+      end
+
+      it "sets start_mode to :manual" do
```
jreidinger: same here

library/systemd/test/systemd_service_test.rb
imobachgs: I wanted to avoid repetition. To be honest, `enable` now sets the start mode to `:boot` (that's what it does).

library/systemd/src/modules/systemd_service.rb
joseivanlopez: IMHO, start mode should be `:demand` as long as the socket is enabled. In fact, whether the service is enabled or not should not be relevant in that case.

library/systemd/src/modules/systemd_service.rb
joseivanlopez: If socket is not found in a first call, it will try to find it again in subsequent calls. Does it make sense? I don't know.

library/systemd/test/systemd_service_test.rb
jreidinger: well, but we have this method just to be backward compatible. So we should be sure it do what it did before. So we do not break other modules.

library/systemd/src/modules/systemd_service.rb
```diff
@@ -230,10 +230,16 @@ def socket?
 
       # Return the start mode
       #
+      # See {#start_modes} to find out the supported modes for a given service (usually :boot,
+      # :manual and, in some cases, :demand).
+      #
+      # When the service are the service (:boot) and the socket (:demand) are enabled,
```
jreidinger: double "are"

library/systemd/src/modules/systemd_service.rb
imobachgs: No, `start_mode` is not based on `enabled?`. Take into account that `#socket` returns an instance of a different class.

library/systemd/test/systemd_service_test.rb
```diff
@@ -276,26 +276,32 @@ module Yast
...
       before do
         stub_services(service: "cups")
+        allow(service).to receive(:socket).and_return(socket)
       end
 
-      it "sets start_mode to :boot" do
-        expect(service).to receive(:start_mode=).with(:boot)
+      it "enables the service to start on boot" do
+        expect(service).to receive(:enable_service)
+        expect(socket).to receive(:disable)
```
jreidinger: hmm, that in fact realize me that we are not 100% backward compatible, so maybe create separate `it` state for this side effect. As previously service.enable does not disable socket. So we should document it.

library/systemd/src/modules/systemd_service.rb
imobachgs: Clarified offline. `enabled?` (from parent class) is now `enabled_on_boot?` and `enabled?` has been redefined.

jreidinger: this is also backward incompatible change, but I think it is fine.

PR #751, opened 2018-07-03
Add a SystemService class

imobachgs:
As discussed on IRC, this PR adds a SystemService class to bring services and sockets together.
Service finders should be improved (probably even moved to its own class) but, for the time being,
they are good enough.


library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,139 @@
...
+    #
+    # * :on_boot:   The service will be started when the system boots.
+    # * :manual: The service is disabled and it will be started manually.
+    # * :on_demand: The service will be started on demand (using a Systemd socket).
+    #
+    # @return [Array<Symbol>] List of supported modes.
+    def start_modes
+      return @start_modes if @start_modes
+      @start_modes = [:on_boot, :manual]
+      @start_modes.insert(1, :on_demand) if socket?
```
jreidinger: is order really so important? I do not see any note about this order in documentation. So maybe put it just to the end?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,139 @@
...
+      @start_modes
+    end
```
jreidinger: in general I think we need to delegate a lot of methods in this class so it can behave like real service and not just wrapper around pair of socket and service. Like start, stop, restart and other methods that behaves smartly if there is socket or not.

imobachgs: Yes, that's true. But extracting this class is the first step. We will more logic in the near future.

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,139 @@
...
+      # SystemdService.find_many.
+      #
+      # @param names [Array<String>] Service names to find
+      # @return [Array<SystemService>] Found system services
+      def find_many(names)
+        Yast::SystemdService.find_many(names).compact.map { |s| new(s) }
+      end
+    end
+
+    # @param service [SystedService]
```
joseivanlopez: NP: typo SystedService.

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,139 @@
...
+    def initialize(service)
+      @service = service
+    end
+
+    # Returns socket associated with service or nil if there is no such socket
```
joseivanlopez: NP: please, document return type.

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,139 @@
...
+      return :on_boot if service.enabled?
+      return :on_demand if socket && socket.enabled?
+      :manual
+    end
+
+    # Set the service start mode
+    #
+    # See {#start_modes} to find out the supported modes for a given service (usually :on_boot,
+    # :manual and, in some cases, :on_demand).
+    #
```
joseivanlopez: NP: I would comment the exception case.

PR #753, opened 2018-07-03
Merge SLE-12-SP4 into SLE-15-GA

teclator:


library/network/src/lib/network/susefirewall2.rb
```diff
@@ -1627,7 +1627,7 @@ def ActivateConfiguration
       # starting firewall during second stage can cause deadlock in systemd - bnc#798620
       # Moreover, it is not needed. Firewall gets started via dependency on multi-user.target
       # when second stage is over.
-      if Mode.installation
+      if Mode.installation && !Mode.autoinst
```
teclator: Although merged, the complete class should be removed in short as it should not used anymore

PR #760, opened 2018-07-05
Add a SystemService#save method

imobachgs:
* SystemService store changes in memory and commit them to the underlying system
  when `#save` is called.
* The `#socket` method is moved back to the `SystemdService` class.

### DIFF
--- a/library/system/src/lib/yast2/system_service.rb
+++ b/library/system/src/lib/yast2/system_service.rb
@@ -28,6 +28,24 @@ module Yast2
   # When talking about systemd, it might happen that a service is compose by a set of units
   # (services, sockets, paths and so on). This class is able to group those units and
   # offer an API to handle them together.
+  #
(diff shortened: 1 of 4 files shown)


library/system/src/lib/yast2/system_service.rb
```diff
@@ -136,8 +164,57 @@ def start_mode=(mode)
...
+        socket.enable
+      when :manual
+        service.disable
+        socket.disable
+      end
+    end
+
+    # Sets service status
+    def set_current_status
+      if changes[:active] && !service.active?
```
joseivanlopez: Should the socket be started?

imobachgs: Good point. I would say 'yes', but in this case the "activate" is about having the service running (the "Start/Stop" button). I will add a note in the PBI to think/discuss more about it.

PR #761, opened 2018-07-06
Add methods

joseivanlopez:


library/system/test/yast2/system_service_test.rb
```diff
@@ -336,4 +336,26 @@
       end
     end
   end
+
+  describe "#search_terms" do
+    before do
+      allow(service).to receive(:id).and_return("cups.service")
+    end
+
+    context "when the service has not associated socket" do
```
imobachgs: np: "has not associated" or "has no associated" or "does not have an associated" ... not sure

PR #762, opened 2018-07-06
Add a SystemService#changed_value?

imobachgs:
* Add changed_value?
* Alias SystemService#active? to SystemService#active.

### DIFF
--- a/library/system/src/lib/yast2/system_service.rb
+++ b/library/system/src/lib/yast2/system_service.rb
@@ -49,8 +49,6 @@ module Yast2
   class SystemService
     extend Forwardable
 
-    Change = Struct.new(:old_value, :new_value)
-
     # @return [Yast::SystemdService]
     attr_reader :service
 
@@ -121,7 +119,7 @@ module Yast2
       if mode == current_start_mode
         unregister_change(:start_mode)
       else
-        register_change(:start_mode, current_start_mode, mode)
+        register_change(:start_mode, mode)
       end
     end
 
@@ -134,7 +132,7 @@ module Yast2
       if value == service.active?
         unregister_change(:start_mode)
       else
-        register_change(:active, service.active?, value)
+        register_change(:active, value)
       end
     end
 
@@ -142,8 +140,7 @@ module Yast2
     #
     # @return [Boolean] true if the service must be active; false otherwise
     def active
-      new_value = new_value_for(:active)
-      return new_value unless new_value.nil?
+      return new_value_for(:active) if changed_value?(:active)
       service.active?
     end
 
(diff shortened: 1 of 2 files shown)


library/system/src/lib/yast2/system_service.rb
```diff
@@ -128,19 +128,24 @@ def start_mode=(mode)
...
+    # Determine whether the service will be active after calling #save
+    #
+    # @return [Boolean] true if the service must be active; false otherwise
     def active
-      return changes[:active] unless changes[:active].nil?
+      return new_value_for(:active) if changed_value?(:active)
       service.active?
     end
 
+    alias_method :active?, :active
```
joseivanlopez: Why not `active?` directly?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -128,19 +128,24 @@ def start_mode=(mode)
     # The given value will be applied after calling #save.
     #
     # @param value [Boolean] true to set this service as active
-    def active=(active)
-      if active == service.active?
-        changes.delete(:active)
+    def active=(value)
+      if value == service.active?
+        unregister_change(:start_mode)
```
joseivanlopez: :start_mode? I guess it should be `unregister_change(:active)`

joseivanlopez: I would add a test for `#active=` and `#active=`.

library/system/src/lib/yast2/system_service.rb
imobachgs: It is a `leftover` (I kept them for compatibility reasons). But not needed anymore, so I've removed it.

imobachgs:
Having those `#start`, `#stop` methods makes sense. I went for `active=` simply because it was more convenient while adapting the `ServicesManagerService`.

library/system/test/yast2/system_service_test.rb
```diff
@@ -162,69 +162,129 @@
...
+
+      it "ignores the change" do
+        system_service.start
+        expect(system_service.changed?).to eq(false)
+      end
+    end
 
-      it "sets #active to true" do
-        expect { system_service.active = true }.to change { system_service.active? }
+    context "and the services was set to be stopped" do
```
joseivanlopez: NP: service, in singular.

library/system/test/yast2/system_service_test.rb
```diff
@@ -162,69 +162,129 @@
...
+        expect(system_service.active?).to eq(true)
+      end
+    end
+
+    context "when service was set to be stopped" do
+      before do
+        system_service.stop
+      end
+
+      it "returns true" do
```
joseivanlopez: Should be "returns false".

PR #763, opened 2018-07-16
add table of states and actions

jreidinger:


joseivanlopez:
Already included with https://github.com/yast/yast-yast2/pull/773

PR #767, opened 2018-07-17
Add errors support to SystemService

imobachgs:
* Add an API to query about errors when changing system services settings.

### DIFF
--- a/library/system/src/lib/yast2/system_service.rb
+++ b/library/system/src/lib/yast2/system_service.rb
@@ -49,20 +49,15 @@ module Yast2
   class SystemService
     extend Forwardable
 
-    # @return [Yast::SystemdService]
+    # @return [Yast::SystemdServiceClass::Service]
     attr_reader :service
 
-    def_delegators :@service, :running?, :name, :description
+    # @return [Hash<Symbol,Object>] Errors when trying to write changes to the
+    #   underlying system.
+    attr_reader :errors
 
-    # @!method state
-    #
-    # @return [String]
-    def_delegator :@service, :active_state, :state
-
-    # @!method substate
-    #
-    # @return [String]
-    def_delegator :@service, :sub_state, :substate
+    def_delegators :@service, :running?, :start, :stop, :restart, :active?,
+      :active_state, :sub_state, :name, :description
 
     class << self
       # Find a service
@@ -89,6 +84,7 @@ module Yast2
     def initialize(service)
       @service = service
       @changes = {}
+      @errors = {}
     end
 
     # Returns the start mode
@@ -156,6 +152,7 @@ module Yast2
     #
     # @param set_status [Boolean] Do not change service status. Useful when running on 1st stage.
     def save(ignore_status: false)
+      clear_errors
       save_start_mode
       set_current_status unless ignore_status
       reload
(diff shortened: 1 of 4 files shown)


library/system/src/lib/yast2/system_service.rb
```diff
@@ -49,20 +49,15 @@ module Yast2
...
 
-    # @!method state
-    #
-    # @return [String]
-    def_delegator :@service, :active_state, :state
-
-    # @!method substate
-    #
-    # @return [String]
-    def_delegator :@service, :sub_state, :substate
```
joseivanlopez: I think we should not remove those lines. Systemd methods `#active_state` and `#sub_state` are mapped to `#state` and `#substate`. The idea was to avoid the usage of systemd names but more general ones.

joseivanlopez: In fact, there should be code in yast-services-manager using those methods.

library/system/test/yast2/system_service_test.rb
```diff
@@ -370,13 +383,34 @@
...
         end
 
-        it "does not active the service if the status must be ignored" do
-          expect(service).to_not receive(:start)
-          system_service.save(ignore_status: true)
+        context "and the service is successfully started" do
+          it "does not register any error" do
+            allow(service).to receive(:start).and_return(true)
+            system_service.save
+            expect(system_service.errors).to_not have_key(:activate)
```
joseivanlopez: The key to check is `:active` instead of `:activate`.

library/system/test/yast2/system_service_test.rb
```diff
@@ -392,13 +426,34 @@
...
         end
 
-        it "does not stop the service if the status must be ignored" do
-          expect(service).to_not receive(:start)
-          system_service.save(ignore_status: true)
+        context "and the service is successfully stopped" do
+          it "does not register any error" do
+            allow(service).to receive(:stop).and_return(true)
+            system_service.save
+            expect(system_service.errors).to_not have_key(:activate)
```
joseivanlopez: Same here, key should be `:active`.

library/system/src/lib/yast2/system_service.rb
```diff
@@ -239,27 +246,43 @@ def current_start_mode
...
+      result =
+        case changes[:start_mode]
+        when :on_boot
+          service.enable && socket.disable
+        when :on_demand
+          service.disable && socket.enable
+        when :manual
+          service.disable && socket.disable
+        end
+      register_error(:start_mode) unless result
```
jreidinger: to be honest I do not like much this error attribute way. It is a bit like C style. Can we add least add save! that will raise well defined exception when it failed?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -56,6 +56,16 @@ class SystemService
...
+    #
+    # @return [String]
+    def_delegator :@service, :active_state, :state
+
+    # @!method substate
+    #
+    # @return [String]
+    def_delegator :@service, :sub_state, :substate
+
     def_delegators :@service, :running?, :start, :stop, :restart, :active?,
```
joseivanlopez: NP: I would not forward `#active_state` and `sub_state` directly.  It is true that right now we are simply forwarding that methods with a different name, but those systemd values should be "internal" details.

library/system/src/lib/yast2/system_service.rb
imobachgs: Well, it is more like Rails style. But yes, we can improve it later, I will add a new task to the PBI.

PR #768, opened 2018-07-17
Add SystemService class

joseivanlopez:
Needed for PBI: https://trello.com/c/fZC7vbpA/108-3-fate319428-mu-display-correct-status-information-on-services-list


library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+  # When talking about systemd, it might happen that a service is compose by a set of units
+  # (services, sockets, paths and so on). This class is able to group those units and
+  # offer an API to handle them together.
+  #
+  # Additionally, this class does not modify the underlying system until the #save
+  # method is called.
+  #
+  # @example Enabling a service
+  #   cups = SystemService.find("cups")
+  #   cups.start_mode = :boot
```
jreidinger: I am not sure after today call if this is good name for API. Maybe we can have it `cups.after_reboot = :start_service` or `start_socket` ?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+
+    # @return [Yast::SystemdService]
+    attr_reader :service
+
+    def_delegators :@service, :running?, :name, :description
+
+    # @!method state
+    #
+    # @return [String]
+    def_delegator :@service, :active_state, :state
```
jreidinger: this looks wrong as it ignore socket status, not?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+
+      # Finds service names
+      #
+      # This method finds a set of system services. Currently it is just a wrapper around
+      # SystemdService.find_many.
+      #
+      # @param names [Array<String>] Service names to find
+      # @return [Array<SystemService>] Found system services
+      def find_many(names)
+        Yast::SystemdService.find_many(names).compact.map { |s| new(s) }
```
jreidinger: to be honest, I do not like much here that it just ignore names that is not found. why not return in map nils?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+    # Toggles the service status
+    #
+    # @see #active=
+    def toggle
+      self.active = !active?
+    end
+
+    # Saves changes to the underlying system
+    #
+    # @param set_status [Boolean] Do not change service status. Useful when running on 1st stage.
```
jreidinger: Do we need param? Can't we simple skip it in first stage and detect it ourself?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+    def save(ignore_status: false)
+      save_start_mode
+      set_current_status unless ignore_status
+      reload
+    end
+
+    # Reverts stored changes
+    def reload
```
jreidinger: I really really dislike this API name as it is confusing as service also has method reload. Why not simply call it read?

jreidinger: or reset?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+      @start_modes
+    end
+
+    # Terms to search for this service
+    #
+    # In case the service has an associated socket, the socket name
+    # is included as search term.
+    #
+    # @return [Array<String>] e.g., #=> ["tftp.service", "tftp.socket"]
+    def search_terms
```
jreidinger: I do not like API name. It is too much reveal that it is used in journal. Why not call it e.g. "systemd_units"?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+      terms = [service.id]
+      terms << socket.id if socket
+      terms
+    end
+
+    # Determines whether a value has been changed
+    #
+    # @return [Boolean] true if the value has been changed; false otherwise
+    def changed_value?(key)
```
jreidinger: NP: why not have `changed?` with default param nil which acts like now and if something is passed it acts like this? so call looks like `changed?(:active)` ?

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+      changes.key?(key)
+    end
+
+  private
+
+    # @return [Hash<String,Object>]
```
jreidinger: is it correct? below it mentions symbols as key

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+    attr_reader :changes
+
+    # Sets whether the service should be active or not
+    #
+    # The given value will be applied after calling #save.
+    #
+    # @param value [Boolean] true to set this service as active
+    def active=(value)
```
jreidinger: to be honest, I really dislike this API, as it ignore that we maybe want also restart or reload. I think we also discuss that this ability will be useful also for services manager to allow restart or reload.
What I do not like is that you set target state, not action that should be done on top of it.

library/system/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,301 @@
...
+        socket.enable
+      when :manual
+        service.disable
+        socket.disable
+      end
+    end
+
+    # Sets service status
+    def set_current_status
+      if new_value_for(:active) && !service.active?
```
jreidinger: again this a bit ignore sockets.

PR #769, opened 2018-07-17
Add SystemService

joseivanlopez:
Needed for PBI: https://trello.com/c/fZC7vbpA/108-3-fate319428-mu-display-correct-status-information-on-services-list

NOTE: ~~Take into account all review comments from https://github.com/yast/yast-yast2/pull/768~~ (done)


mvidner:
OK, I am going to ask stupid questions:

Why do we need a SystemService thing? The name is IMHO awful: we already have System**d**Service. A "service" is a very overloaded term, and adding "system" does not really help. What **is** it, really? Why does not systemd itself need it?

library/systemd/README.md
```diff
@@ -0,0 +1,76 @@
+# Systemd actions
+
+The following table describes the actions to perform over a systemd service and its associated socked (if any) when we try to start, stop, restart or reload the service.
```
mvidner: s/socked/socket/

library/systemd/README.md
```diff
@@ -0,0 +1,76 @@
+# Systemd actions
```
mvidner: This file should be visible in the YARD output, and linked from the relevant classes. Maybe it should be inlined in the relevant classes instead?

joseivanlopez: Well, I can link the file in relevant classes, but can I make it part of YARD output?

jreidinger: @joseivanlopez add it to .yardopts

jreidinger:
@mvidner I think it is a kind of useful. Do you have sources somewhere, so we can edit it to current state? And probably split yast2 part and service manager part.

library/systemd/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+The following table describes the actions to perform over a systemd service and its associated socked (if any) when we try to start, stop, restart or reload the service.
+
+
+| Status | Start | Stop | Restart | Reload |
```
mvidner: Note that the Restart column could be rephrased and unified, to always perform "stop, then start" (where "start" operates in the mode "nothing is running").
This would fix a bug in the current table: if both service+socket are running, and you have Manual start mode, and you Restart, it leaves the socket running.

library/systemd/README.md
```diff
@@ -0,0 +1,76 @@
...
+|---|---|---|---|---|
+| Socket running | nothing | stop socket | stop socket and start service/socket (depending on start mode) | nothing |
+| Service running | nothing  | stop service  | stop service and start service/socket (depending on start mode) | reload service (if it support, otherwise restart) |
+| Socket & Service running | nothing | stop service and socket | stop service and start service/socket (depending on start mode) | reload service (if it support, otherwise restart) |
+| nothing runs | start service/socket (depending on start mode) | nothing | start service/socket (depending on start mode) | start service/socket (depending on start mode) |
+
+
+## Actions over a `SystemService`
+
+In systemd, it might happen that a service is compose by a set of units (services, sockets, paths and so on). This class `SystemService` is able to group those units and handle them together.
```
mvidner: This might confuse the reader in the direction of CompoundService, which is something else.
As I understand it, this class only handles the concept of Laziness: activating the real Service via a Socket or a Path, or a Timer, which is its light-weight stand-in/proxy. In case we want to add Paths and Timers (there are no other proxies for services*), we could use the word Proxy (or StandIn but that is a hyphenated-word, bother).

*: For systemd.mount there is a systemd.automount stand-in, but we do not handle mounts.

mvidner: Examples:
- ca-certificates.path that runs ca-certificates.service (Type=oneshot) when a cert file is added/removed.
- backup-rpmdb.timer (oneshot, backs it up daily)

library/systemd/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### `SystemService` status
+
+* [1] A service is stopped when neither the systemd socket nor service are running
+* [2] A service is running when the systemd socket and/or service are running
+
+### Start
+
+* Target
```
mvidner: Please use some other word that does not have a meaning in systemd. "Goal" or "Intent"

library/systemd/README.md
```diff
@@ -0,0 +1,76 @@
...
+      * starts systemd service
+
+### Stop
+
+* Target
+  * service is stopped (see [1])
+
+* Actions
+  * stops systemd service if it is running
+  * stops systemd socket if it is running
```
mvidner: In reverse order, otherwise there is a race condition.

library/systemd/README.md
```diff
@@ -0,0 +1,76 @@
...
+  * service is stopped (see [1]) and started again (see [2])
+
+* Actions
+  * calls to "stop service action" (see stop section) if the service is running (see [2])
+  * calls to "start service action" (see start section)
+
+### Reload
+
+* Target
+  * service is reloaded or restarted or started
```
mvidner: IIUC the intention is "the configuration has changed and I want it to take effect".
In that case, we should name this ForceReload, to match the corresponding [LSB action](http://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html):

> 
action | description
-- | --
start | start the service
stop | stop the service
restart | stop and restart the service if the service is already running, otherwise start the service
try-restart | restart the service if the service is already running
reload | cause the configuration of the service to be reloaded without actually stopping and restarting the service
force-reload | cause the configuration to be reloaded if the service supports this, otherwise restart the service if it is running
status | print the current status of the service

(LSB (Linux Standard Base) is from the System V times, maybe systemd has a different opinion on this)

jreidinger: systemd has reload-or-restart action. The main goal is "service is running with new configuration and tries to do it without service disruption"

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+    # Constructor
+    #
+    # @param service [Yast::SystemdServiceClass::Service]
+    def initialize(service)
+      @service = service
+      @changes = {}
+      @errors = {}
+    end
+
+    # State of the service
```
mvidner: "state" is a very overloaded term. Why not name this method `active_state` ? That would also kind of document the allowed values of the string.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+    #
+    # In case the service is not active but socket, the socket state is considered
```
mvidner: s/but socket/but the socket is/

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+          :on_demand
+        else
+          :manual
+        end
+    end
+
+    # Whether the service is currently active in the system
+    #
+    # @return [Boolean]
+    def current_active?
```
mvidner: EnglishError: s/current_active?/currently_active?/

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+    # In case the service is not active but socket, the socket substate is considered
+    #
+    # @return [String]
+    def substate
+      return socket.sub_state if socket_active? && !service.active?
+
+      service.sub_state
+    end
+
+    # Gets the current start_mode (as read from the system)
```
mvidner: missing: a note about why/how this is different from {#start_mode}

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+    end
+
+    # Whether the service is currently active in the system
+    #
+    # @return [Boolean]
+    def current_active?
+      service.active? || socket_active?
+    end
+
+    # Returns the list of supported start modes
```
mvidner: += "for this service (if a socket unit is available, :on_demand is supported, otherwise not)"

mvidner: Duh, it's kind of obvious, but someone might wonder why this is not a class constant instead.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+      keywords = [service.id]
+      keywords << socket.id if socket
+      keywords
+    end
+
+    # Sets the service to be started after calling {#save}
+    #
+    # @see #active=
+    #
+    # @return [Symbol] :start
```
mvidner: Why is the return value important? I would think it is accidental, and not document it, or explicitly documented a [void]

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+    def reload
+      register_change(:active, true)
+      self.action = :reload
+    end
+
+    # Saves changes into the underlying system
+    #
+    # @note Cached changes are reset and the underlying service is refreshed.
+    #
+    # @param keep_state [Boolean] Do not change service status. Useful when running on 1st stage.
```
mvidner: Again, status (state) is so overloaded. We mean the active_state, right?

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+      clear_errors
+      save_start_mode
+      perform_action unless keep_state
+      reset && refresh
+    end
+
+    # Reverts cached changes
+    #
+    # @return [Boolean] true if the service was reset correctly. Actually, the
+    #   service always can be reset.
```
mvidner: Does this mean this method is a part of a team of methods with the same interface? Use a `@see` tag?

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,490 @@
...
+    end
+
+    # Sets start mode to the underlying system
+    def save_start_mode
+      return unless changes[:start_mode]
+
+      result =
+        case changes[:start_mode]
+        when :on_boot
+          service.enable && (socket ? socket.disable : true)
```
mvidner: SystemdUnit currently does not document the return value of #enable and #disable, please fix that.

library/systemd/src/lib/yast2/system_service.rb
joseivanlopez: Well, we have discarded `#active_state` because we want the API to be as agnostic as possible. `#active_state` is too related to systemd.

library/systemd/src/lib/yast2/system_service.rb
joseivanlopez: Well, in public methods I always like documenting the return value. But yes, I think it is better to use void in this case.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -0,0 +1,573 @@
...
+    #   @return [Boolean]
+    # @!method description
+    #   @see Yast::SystemdServiceClass::Service#description
+    #   @return [String]
+    def_delegators :@service, :name, :static?, :running?, :description
+
+    class << self
+      # Finds a service by its name
+      #
+      # @param name [String] service name with or without extension (e.g., "cups" or "cups.service")
```
mvidner: So... if "cups.service" works, does "cups.socket" work too? That is, am I able to find the SystemService using any of its unit names? If not, wouldn't it be better to allow only the short name?

joseivanlopez: I don't think so. Both, "cups" and "cups.service" work because SystemdServiceClass#find supports it. `ServiceManager` simply delegates the call, see https://github.com/yast/yast-yast2/blob/master/library/systemd/src/modules/systemd_service.rb#L77

PR #770, opened 2018-07-18
Alternative naming

jreidinger:
original screenshots at https://github.com/yast/yast-yast2/pull/747

with alternative naming:

snimek obrazovky_2018-07-18_08-20-28
snimek obrazovky_2018-07-18_08-28-26
snimek obrazovky_2018-07-18_08-22-37


library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,202 @@
...
+      if services.any? { |s| !s.is_a?(Yast2::SystemdService) }
+        raise ArgumentError, "Services can be only System Service - #{services.inspect}"
+      end
+
+      @services = services
+    end
+
+    # writes services new status
+    # @raise <Yast::SystemctlError> when set service to target state failed
+    def save(ignore_status: false)
```
joseivanlopez: NP: In SystemService I have rename this param as `keep_state`. Maybe we should do the same here.

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,202 @@
...
+    def action
+      # TODO: check for inconsistencies?
+      services.first.action
+    end
+
+    ACTIONS = [:start, :stop, :reload, :restart, :nothing].freeze
+    # sets target action for services
+    # @param action[:start, :stop, :restart, :reload, :nothing] for possible values and
+    #  its explanation please see return value of {target_action}
+    def action=(action)
```
joseivanlopez: Actually needed? We have separate actions `#start`, `#stop`, etc.

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,202 @@
...
+    def has_current_mode(start_mode)
+      ->(service) { service.current_start_mode == start_mode }
+    end
+
+    def has_mode(start_mode)
+      ->(service) { service.start_mode == start_mode }
+    end
+
+    def services_mode(mode_method)
+      if services.all?(&mode_method.call(:enabled?))
```
joseivanlopez: `:enabled?` ?

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,202 @@
...
+      elsif services_with_socket.all?(&mode_method.call(:on_demand)) &&
+          services_without_socket.all?(&mode_method.call(:on_boot))
+        :on_demand
+      else
+        :inconsistent
+      end
+    end
+
+    def services_with_socket
+      @services_with_socket ||= services.select{ |s| s.start_modes.include?(:on_demand) }
```
joseivanlopez: I guess we will start to use here `#support_start_on_demand`.

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,202 @@
+require "yast"
+
+require "yast2/system_service"
+
+module Yast2
+  # Class that represents several related services that need to be synchronized.
+  # It mimics behavior of single SystemService, just adds additional possible states.
+  class CompoungService
```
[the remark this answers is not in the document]
jreidinger: thanks

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,202 @@
...
+    include Yast::Logger
+
+    # managed services
+    # @return [Yast2::ServiceConfiguration]
```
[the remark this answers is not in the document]
jreidinger: right, overlooked.

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,202 @@
...
+    attr_reader :services
+
+    # creates new configuration that holds state for given services
+    # @param services<Yast::SystemdServiceClass::Service> services to configure
```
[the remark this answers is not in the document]
jreidinger: also wrong. will fix it.

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,202 @@
...
+    # @param reload<true,false> if use reload instead of restart action. If
+    #   service does not support reload or does not run, then restart is used.
+    #
+    # @example three services
+    #   config = Yast2::ServiceConfiguration.new(s1, s2, s3)
+    def initialize(*services)
+      if services.any? { |s| !s.is_a?(Yast2::SystemdService) }
```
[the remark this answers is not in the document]
jreidinger: thanks for catch. I plan to write some unit tests when we agreed on API which hopefully should catch it.

library/systemd/src/lib/yast2/compound_service.rb
jreidinger: yes, plan is to have identical API.

library/systemd/src/lib/yast2/compound_service.rb
jreidinger: no, not needed, I forgot to remove it.

library/systemd/src/lib/yast2/compound_service.rb
jreidinger: good catch, I use wrong symbol. It should be `:on_boot`

library/systemd/src/lib/yast2/compound_service.rb
jreidinger: yes, I write it before we agreed on it.

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -33,24 +55,24 @@ module Yast2
...
   class ServiceWidget
     include Yast::I18n
     include Yast::Logger
     include Yast::UIShortcuts
-    # creates new widget instance for given service configuration
-    # @param service_configuration [Yast2::ServiceConfiguration] configuration holder
-    def initialize(service_configuration)
+
+    # creates new widget instance for given service
+    # @param service_configuration [Yast2::SystemService,Yast2::CompoundService] service
```
joseivanlopez: Param name `service_configuration` is wrong.

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -33,24 +55,24 @@ module Yast2
...
+      @service = service
     end
 
     # gets widget term
     # @return <Yast::Term>
     def content
-      # TODO: disabling invalid action and autostart or kick it out?
       Frame(
         _("Service Configuration"),
         VBox(
```
joseivanlopez: Why not to use `HSpacing` instead of `Label(" ")` ?

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,209 @@
...
+      services.each { |s| s.save(keep_state: keep_state) }
+    end
+
+    # returns current running status.
+    # @return [true, false, :inconsistent] current status of
+    #   services:
+    #
+    #   - `true` when all services are active
+    #   - `false` when all services are inactive
+    #   - `:inconsistent` when part of services is active and part not.
```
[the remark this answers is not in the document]
jreidinger: that is problematic for service_widget. I need to use identical API. Any new method is problematic. So I need somehow keep existing API and add there a state that shows that only part runs and I found this the easiest solution. Adding new method breaks idea of having same API only with slightly different return states. ( and if someone ignore it, then it acts like if it is running, so he can do restart or reload which ensure everything is running, or stop which stops the rest ).

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,209 @@
...
+
+    def start_modes
+      @start_modes ||= services.map(&:start_modes).reduce(:+).uniq
+    end
+
+    def keywords
+      services.map(&:keywords).reduce(:+).uniq
+    end
+
+    # returns currently set target action. If it is not yet set it returns
```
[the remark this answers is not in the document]
jreidinger: yep, it was old behavior and I forgot to change documentation.

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -80,86 +102,90 @@ def handle_input(event_id)
     # Stores current configuration. Should be called always even when going
```
joseivanlopez: NOTE: This comment is for method `handle_input`. The documentation of `handle_input` says `@return [void]`. Is that correct?

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -0,0 +1,209 @@
...
+    # @param [:on_boot, :on_demand, :manual, :inconsistent] sets start_mode.
+    #   See {Yast2::SystemService.start_mode=}. Additionally inconsistent is used
+    #   to keep current start mode. For each service.
+    def start_mode=(configuration)
+      if !AUTOSTART_OPTIONS.include?(configuration)
+        raise ArgumentError, "Invalid parameter #{configuration.inspect}"
+      end
+
+      if configuration == :inconsistent
+        reset(exclude: [:action])
```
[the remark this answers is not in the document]
jreidinger: yes, maybe whitelist makes more sense.
and inconsistent start mode is needed for widget for option "keep it as it is". Which does not make sense for single service, as state is already defined.

library/systemd/src/lib/yast2/service_widget.rb
jreidinger: Well, I basically copy ui from ServiceStatus. Maybe it is needed in ncurses? I do not know

library/systemd/src/lib/yast2/service_widget.rb
jreidinger: in fact no. It is comment for store. You should call store after leaving dialog to remember user selection, so when going again forward it will be set correctly. Maybe I wrote it badly?

handle_input should has nil as return as it means in cwm to continue with dialog and not return given value from it.

library/systemd/src/lib/yast2/compound_service.rb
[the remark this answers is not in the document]
joseivanlopez: Yes, but in the widget, the start mode is not set when the selected value is :inconsistent [1], so theoretically we never try to assign that value. And I am not sure if it makes sense to assign that value at all.

[1] https://github.com/yast/yast-yast2/pull/770/files#diff-aeb8aef16363d75a091e840d2710574aR129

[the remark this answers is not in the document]
jreidinger: It is value that is set when original state is inconsistent. What can then happen that in proposal or module it is set to e.g. `:on_boot`  or user select it to something and then he changed mind and wants to keep it as it is now. In such case he selects "Keep current state" and widget assign `:inconsistent`.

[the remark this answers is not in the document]
jreidinger: but you are right that widget now use generic reset and do configuration from scratch. But I would like to have consistent API, which can be used also at other places. So for me it makes sense to set this value to say do no action with start_mode.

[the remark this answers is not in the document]
jreidinger: at least it is consistent in a way that all values that "start_mode" can return can be also assigned back.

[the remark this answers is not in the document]
jreidinger: so e.g. `service1.start_mode = service2.start_mode` will always work

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -154,6 +156,14 @@ def support_start_on_demand?
       @services.any?(&:support_start_on_demand?)
     end
 
+    # @return [Hash<Symbol,Object>] Errors when trying to write changes to the
+    #   underlying system.
+    def errors
+      @services.each_with_object({}) do |s, result|
```
dgdavid: NP: `@` is not needed here

library/systemd/src/lib/yast2/compound_service.rb
```diff
@@ -47,9 +47,11 @@ def initialize(*services)
     end
 
     # writes services new status
-    # @raise <Yast::SystemctlError> when set service to target state failed
+    # @return [true,false] returns false if saving any service failed
```
joseivanlopez: NP: I prefer `@return [Boolean]`

library/systemd/test/test_helper.rb
```diff
@@ -6,6 +6,21 @@
 Yast.import "SystemdService"
 Yast.import "SystemdTarget"
 
+# Find a Term in given content
+#
+# @param content [Yast::Term] the content in which perform the search
+# @param type [Symbol] wanted Term type
+# @param id [Symbol] wanted Term id
+#
+# @return [Yast::Term|nil] a Term when found; nil otherwise
```
joseivanlopez: NP: I am not sure if `|` delimiter is accepted in YARD, but we usually use comma: `[Yast::Term, nil] `

joseivanlopez: NP: I am not sure if `|` delimiter is accepted in YARD, but we usually use comma: `[Yast::Term, nil] `

dgdavid: Interesting. Until now I always used `|` and probably it works also. But if the common practice is to use comma, I am going to change it :)

Thanks

joseivanlopez:
Closed in favor of https://github.com/yast/yast-yast2/pull/791

PR #772, opened 2018-07-19
Move class and define API

joseivanlopez:


library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -97,6 +97,52 @@ def initialize(service)
...
+          :on_boot
+        elsif socket && socket.enabled?
+          :on_demand
+        else
+          :manual
+        end
+    end
+
+    # TODO
+    def current_active?
```
jreidinger: well, I have one problem with it. How to recognize here if restart or reload is selected? this allows only to distinguish between start and stop.

joseivanlopez: Ah, yes. I forgot the #action method.

PR #773, opened 2018-07-19
[WIP] Add actions to SystemService

joseivanlopez:


library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -286,18 +292,48 @@ def save_start_mode
...
-    def set_current_status
-      return if changes[:active].nil?
-      result =
-        if changes[:active] && !service.active?
-          service.start
-        elsif changes[:active] == false && service.active?
-          service.stop
-        end
+    def perform_action
+      result = send("perform_#{action}")
```
jreidinger: where is action defined?

jreidinger: or better, what happen when action is nil as defined above?

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -56,6 +56,14 @@ class SystemService
     #   underlying system.
     attr_reader :errors
 
+    # @!method action
+    #
+    # Action to perform when the service is saved (see {#save})
+    #
+    # @return [Symbol, nil] :start, :stop, :restart or :reload. It returns nil
+    #   if no action has been requested yet.
+    attr_reader :action
```
jreidinger: I think that action should be set to nil in `reset` method.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -175,92 +191,115 @@ def start_mode=(mode)
...
-      set_current_status unless ignore_status
-      reset
+      perform_action unless keep_state
+      refresh
     end
 
-    # Reverts stored changes
+    # Reverts cached changes
+    #
+    # @return [Boolean]
```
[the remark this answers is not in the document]
joseivanlopez: For example, we can have `#refresh` and `#refresh!`. The first one returns true/false, depending on the action was performed or not. The second one would raise an exception if the action was not performed.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -175,92 +175,134 @@ def start_mode=(mode)
...
+
+      true
+    end
+
+    # Refreshes the underlying system and discards cached changes
+    #
+    # @return [Boolean]
+    def refresh
+      service.refresh!
+      @start_modes = []
```
jreidinger: this looks wrong as def start_modes check if it is nil and not empty array.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -175,92 +175,134 @@ def start_mode=(mode)
...
       @current_start_mode = nil
+
+      reset
```
jreidinger: why refresh do reset? I think main reason to have that refresh is to just recheck current state and recompute if changes still makes sense. But I think that it should not reset actions, not?

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -175,92 +175,134 @@ def start_mode=(mode)
...
     end
 
-    # Determines whether the system has been changed or not
+    # Whether the system has been changed
```
jreidinger: has been changed? I think it should be will be changed.

library/systemd/src/lib/yast2/system_service.rb
joseivanlopez: Yes, it is better to have `#reset` and `#refresh` as completely separate methods and call both inside `#save`.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -66,8 +70,12 @@ class SystemService
     # @return [String]
     def_delegator :@service, :sub_state, :substate
 
-    def_delegators :@service, :running?, :start, :stop, :restart, :active?,
-      :active_state, :sub_state, :name, :description, :static?
+    # @!method rupport_reload?
```
jreidinger: typo *rupport*

library/systemd/src/lib/yast2/system_service.rb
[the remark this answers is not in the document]
jreidinger: well, in this case please document when it can return true or false.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -274,30 +319,82 @@ def active=(value)
...
     end
 
+    # Starts the service in the underlying system
+    #
+    # @return [Boolean] true if the service was correctly started
+    def perform_start
+      result = true
+
+      if socket && start_mode == :on_demand
+        result &= socket.start unless socket_active?
```
jreidinger: I think bitwise and is wrong here and should be `&&=` logical and instead

library/systemd/test/yast2/system_service_test.rb
```diff
@@ -78,46 +94,128 @@
...
+
+      it "returns the new start mode" do
+        expect(system_service.start_mode).to eq(:manual)
+      end
+    end
+  end
 
-    it "sets the wanted start_mode" do
+  describe "#start_mode=" do
+    it "caches the wanted start_mode" do
```
jreidinger: I am not sure if caches is right term. Maybe better `sets` ?

library/systemd/test/yast2/system_service_test.rb
```diff
@@ -140,43 +238,143 @@
...
+      allow(service).to receive(:id).and_return("cups.service")
+    end
 
-    context "when no associated socket exists" do
+    context "when the service has no associated socket" do
       let(:socket) { nil }
 
-      it "returns :on_boot and :manual" do
-        expect(system_service.start_modes).to eq([:on_boot, :manual])
+      it "returns only the service full name" do
```
jreidinger: I would specify also type and remove only so "returns array with service full name"

library/systemd/test/yast2/system_service_test.rb
```diff
@@ -140,43 +238,143 @@
...
+    end
+
+    context "when the service has an associated socket" do
+      let(:socket) { service_socket }
+
+      before do
+        allow(socket).to receive(:id).and_return("cups.socket")
+      end
+
+      it "returns the service and socket full names" do
```
jreidinger: and to be consistent here "returns array with service and socket full names"

library/systemd/src/lib/yast2/system_service.rb
joseivanlopez: I didn't realize I was doing bitwise operations. Thanks!

PR #776, opened 2018-07-23
Backport simpleidn

teclator:
Trello card: https://trello.com/c/3c9gw00k
It backport #685 for fixing https://bugzilla.suse.com/show_bug.cgi?id=1101851

### DIFF
--- a/package/yast2.spec
+++ b/package/yast2.spec
@@ -39,6 +39,8 @@ BuildRequires:  rubygem(%{rb_default_ruby_abi}:cheetah)
 BuildRequires:  update-desktop-files
 # For running RSpec tests during build
 BuildRequires:  rubygem(%{rb_default_ruby_abi}:rspec)
+# For converting to/from punycode strings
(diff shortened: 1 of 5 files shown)


teclator:
Travis was failing because the image used did not contain the simpleidn library but I have forced an update of the docker image and also have tested the changes manually, so i expect it to pass once the new image is available.

package/yast2.spec
```diff
@@ -39,6 +39,8 @@ BuildRequires:  rubygem(%{rb_default_ruby_abi}:cheetah)
 BuildRequires:  update-desktop-files
 # For running RSpec tests during build
 BuildRequires:  rubygem(%{rb_default_ruby_abi}:rspec)
+# For converting to/from punycode strings
+BuildRequires:  rubygem(%{rb_default_ruby_abi}:simpleidn)
```
lslezak: Is the package available in SLE12-SP4?

teclator: Yes it is, It is commented in the bug.

https://build.suse.de/request/show/168697

As commented previously I already tested it manually with the rpm from IBS.

PR #780, opened 2018-07-25
Fix typos and readability

Fraser-Bell:
Contributing more as I learn.  Fixed some typos in the comments, made some minor adjustments for readability.

Now **this** is what I call good commenting.

I have just barely begun to study Ruby, have not coded in 30 years (other than html markup and some related), last languages I worked in were Pascal and Assembler, some C.

Yet, this was so well documented that I was able to actually read the lines of code and understand what they were doing, and how they were doing it.

With documentation like this, a relative "newbie" to coding, such as I, would easily be able to work on this kind of code to make fixes.


Fraser-Bell:
Yes, I thought that was the case.  But, it is helping me learn, so worth doing.

PR #785, opened 2018-07-31
Add a SystemdService.build method

imobachgs:
Needed to support handling services during 1st stage. Needed for https://github.com/yast/yast-services-manager/pull/164.

### DIFF
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -145,7 +145,7 @@ module Yast2
       # @param name [String] Service name
       # @return [SystemService] System service based on the given name
       #
-      # @see Yast::SystemdServiceClass.build
+      # @see Yast::SystemdServiceClass#build
       def build(name)
         new(Yast::SystemdService.build(name))
       end
@@ -418,7 +418,7 @@ module Yast2
 
     # Sets start mode to the underlying system
     def save_start_mode
-      return if changes[:start_mode].nil? || changes[:start_mode] == current_start_mode
+      return unless changed?(:start_mode)
 
       result =
         case changes[:start_mode]
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -566,7 +566,7 @@ module Yast2
 
     # Correspondence between changed values and methods to calculate their current value
     CURRENT_VALUE_METHODS = {
-      active: :currently_active?,
+      active:     :currently_active?,
       start_mode: :current_start_mode
     }.freeze
 
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -278,8 +278,8 @@ module Yast2
     #
     # @return [Boolean] true if the service must be active; false otherwise
     def active?
-      return new_value_for(:active) if changed?(:active)
-      currently_active?
+      new_value = new_value_for(:active)
+      new_value.nil? ? currently_active? : new_value
     end
 
     # Keywords to search for this service
@@ -391,7 +391,7 @@ module Yast2
     #
     # @return [Boolean]
     def changed?(key = nil)
-      key ? changes.key?(key) : changes.any?
+      key ? changed_value?(key) : any_change?
     end
 
   private
@@ -560,8 +560,31 @@ module Yast2
     # @param key [Symbol] Change key
     # @return [Object] New value
     def new_value_for(key)
-      return nil unless changed?(key)
       changes[key]
+    end
+
+    # Correspondence between changed values and methods to calculate their current value
+    CURRENT_VALUE_METHODS = {
+      active: :currently_active?,
+      start_mode: :current_start_mode
+    }.freeze
+
+    # Determines whether a value has been changed
+    #
+    # @param key [Symbol] Changed value
+    # @return [Boolean] true if it has changed; false otherwise.
+    def changed_value?(key)
+      new_value = new_value_for(key)
+      return false if new_value.nil?
+      new_value != send(CURRENT_VALUE_METHODS[key])
+    end
+
+    # Determines whether some value has been changed
+    #
+    # @param key [Symbol] Changed value
+    # @return [Boolean] true if it has changed; false otherwise.
+    def any_change?
+      CURRENT_VALUE_METHODS.keys.any? { |k| changed_value?(k) }
     end
   end
 end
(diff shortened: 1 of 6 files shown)


library/systemd/test/yast2/system_service_test.rb
```diff
@@ -102,6 +102,28 @@
...
+
+    it "returns a systemd service even if it does not exist" do
+      expect(Yast::SystemdService).to receive(:build).with("other")
+        .and_return(systemd_service)
+      system_service = described_class.build("other")
+      expect(system_service.service).to eq(systemd_service)
+    end
+  end
+
+  describe ".build" do
```
dgdavid: It seems to me that this `describe` is duplicated.

imobachgs: Oh, I messed up things rebasing :) Thanks.

imobachgs:
Finally, I've introduced an important modification related to how changes are handled. In the past, this class did not register a change if it was not going to effectively modify the service. For instance: calling `SystemService#start` on an already active service did not modify the list of changes.

As this class can now be used in the 1st stage, I had decided to register all changes and just check whether the change should be applied or not when calling `#save`, so you can check the real status of the service at that point.

However, to make this work, we should refresh the service as the first step of calling #save. Or we can leave this responsability to the caller (if you know nothing has changed, you can skip it). Or, actually, I think that we should not check anything and just rely on `Service` module (it knows how to cope with this situations).

library/systemd/src/modules/systemd_service.rb
```diff
@@ -138,6 +136,20 @@ def all(propmap = {})
...
     end
 
+    # Instantiate a SystemdService object based on the given name
+    #
+    # Use with caution as the service might exist or not. If you need to react when
+    # the service does not exist, use SystemdServiceClass.find.
+    #
+    # @param service_name [String] "foo" or "foo.service"
+    # @param propmap [SystemdUnit::PropMap]
+    # @return [Service] `nil` if not found
```
joseivanlopez: I don't get the explanation of return value. Can it be nil?

library/systemd/test/systemd_service_test.rb
```diff
@@ -29,6 +29,23 @@ module Yast
       end
     end
 
+    describe ".build" do
```
joseivanlopez: Actually this is an instance method.

library/systemd/test/systemd_service_test.rb
```diff
@@ -29,6 +29,23 @@ module Yast
...
+        ["sshd", "sshd.service"].each do |service_name|
+          service = SystemdService.build(service_name)
+          expect(service).to be_a(SystemdUnit)
+          expect(service.unit_type).to eq("service")
+          expect(service.unit_name).to eq("sshd")
+        end
+      end
+
+      it "returns a service instance even if the real service does not exist" do
+        stub_services(service: "unknown")
```
joseivanlopez: But you are mocking the "unknown" service, so you are not checking the case when the service does not exist, right?

library/systemd/test/yast2/system_service_test.rb
```diff
@@ -102,6 +102,17 @@
     end
   end
 
+  describe ".build" do
+    let(:systemd_service) { instance_double(Yast::SystemdServiceClass::Service) }
+
+    it "returns a systemd service even if it does not exist" do
```
joseivanlopez: Strictly speaking, `.build` does not return a `SystemdService`. I would write two different statements:

it "returns a service even if the systemd service does not exist"
it "uses a systemd service even if it does not exist"

library/systemd/test/yast2/system_service_test.rb
```diff
@@ -458,137 +458,43 @@
...
-      it "sets the service to stay as active" do
-        expect { system_service.start }.to change { system_service.active? }
-          .from(false).to(true)
-      end
-
-      it "cancels the previous change" do
-        system_service.start
-        expect(system_service.changed?).to eq(false)
-      end
+    it "sets the service as changed" do
```
joseivanlopez: This remembers me that we needs to adapt Services Manager, see https://github.com/yast/yast-services-manager/blob/SLE-15-GA/src/lib/services-manager/widgets/services_table.rb#L282

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -406,16 +412,12 @@ def changed?(key = nil)
...
-      else
-        register_change(:active, value)
-      end
+      register_change(:active, value)
     end
 
     # Sets start mode to the underlying system
     def save_start_mode
-      return unless changes[:start_mode]
+      return if changes[:start_mode].nil? || changes[:start_mode] == current_start_mode
```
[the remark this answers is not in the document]
imobachgs: The advantage of checking `nil` is that you won't call `current_start_mode` unless is strictly needed because, if `changes[:start_mode]` is `nil`, there is nothing to compare.

Anyway, I've reimplemented the changes calculation mechanism so this line looks simpler now.

library/systemd/src/modules/systemd_service.rb
imobachgs: No, it cannot. Copy&paste error :)

library/systemd/test/systemd_service_test.rb
imobachgs: It is tricky: it looks like a `SystemdService` class method but it is a `SystemdServiceClass` instance method, as you say.

library/systemd/test/yast2/system_service_test.rb
imobachgs: I've reimplemented the changes calculation mechanism so it should work out-of-the-box.

library/systemd/test/yast2/system_service_test.rb
imobachgs: Actually, this method does not know whether the underlying service exists or not. The point is that it calls `build` which always returns a `SystemdService` object.

library/systemd/test/systemd_service_test.rb
imobachgs: I did not realize that an "unknown" service was already available as a mock :P

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -565,8 +561,31 @@ def clear_changes
...
+    # @return [Boolean] true if it has changed; false otherwise.
+    def changed_value?(key)
+      new_value = new_value_for(key)
+      return false if new_value.nil?
+      new_value != send(CURRENT_VALUE_METHODS[key])
+    end
+
+    # Determines whether some value has been changed
+    #
+    # @param key [Symbol] Changed value
```
joseivanlopez: Param doc is leftover.

PR #789, opened 2018-08-02
Replaced obsolete API

mchf:
- [bsc#964856](https://bugzilla.suse.com/show_bug.cgi?id=964856)
- following up on https://github.com/yast/yast-network/pull/661
- https://trello.com/c/LFSe0QLU

Purpose is to remove obsolete API which was used (e.g.) for device type recognition a long time ago. Since persistent device names were introduced, usage of this API can lead to incomplete and / or incorrect result.

Old API was already replaced by new implementation. Device type detection is currently based on sysfs and / or device specific options. Aliases management support "options with label" (like IPADDR_1 and others - see man 5 sysconfig)

For necessary updates in yast2-network see https://github.com/yast/yast-network/pull/670


mvidner:
So what is the situation with Additional Addresses? Do they work at all before/after this change?

mvidner:
About "Breaking everything": that's the downside. What's the upside? Is the code that is left presenting a better API?

mchf:
@mvidner: I've updated description to provide more details about purpose of this PR.

library/network/test/network_interfaces_test.rb
```diff
@@ -238,6 +230,7 @@
     end
 
     it "returns an array of devices which have got a different key,value than given ones" do
+      skip("needs improvement due to obsolete API removal")
```
mvidner: LocateNOT is only used by ListDevicesExcept which is unused

library/network/testsuite/tests/NetworkInterfaces2.rb
```diff
@@ -1,210 +0,0 @@
...
-      TEST(->() { NetworkInterfaces.Write("eth") }, [@READ], nil)
-      TEST(->() { NetworkInterfaces.Write("ppp") }, [@READ], nil)
-      TEST(->() { NetworkInterfaces.Write("ippp") }, [@READ], nil)
-      TEST(->() { NetworkInterfaces.Write("trx") }, [@READ], nil)
-      TEST(->() { NetworkInterfaces.Write("atm|tr") }, [@READ], nil)
-      TEST(->() { NetworkInterfaces.Write("") }, [@READ], nil)
-
-      @exported = nil
-
-      DUMP("NetworkInterfaces::Export")
```
mvidner: Export should be tested

library/network/testsuite/tests/NetworkInterfaces2.rb
```diff
@@ -1,210 +0,0 @@
...
-      @exported = NetworkInterfaces.Export("")
-      DUMP(Builtins.sformat("exported=%1", @exported))
-
-      # Test import canonicalizing
-      DUMP("NetworkInterfaces::Import")
```
mvidner: Import should be tested

library/network/test/network_interfaces_test.rb
```diff
@@ -143,6 +145,7 @@
 
     context "given a list of device types and a regex" do
       it "returns device types that don't match the given regex" do
+        skip("needs improvement due to obsolete API removal")
```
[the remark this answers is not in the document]
mvidner: I was wrong, Import uses it

mchf:
PR for yast2-network is in progress - see description, where it is mentioned as TODO. Will update the changelog.

mvidner:
Please remove the `publish` calls too.

TIP: when removing APIs, grep the code for the names that you have removed:

```console
$ for sym in ifcfg_part device_type device_num alias_num GetFreeDevices GetFreeDevice LocateNOT ListDevicesExcept; do
>   find -name \*.rb | xargs grep -n -w $sym;
> done
./library/general/src/modules/Arch.rb:215:          device_type = Convert.to_map(
[more false positives in Arch.rb where device_type is a variable name]
./library/network/src/modules/NetworkInterfaces.rb:1596:    publish function: :device_type, type: "string (string)"
./library/network/src/modules/NetworkInterfaces.rb:1614:    publish function: :GetFreeDevices, type: "list <string> (string, integer)"
./library/network/src/modules/NetworkInterfaces.rb:1615:    publish function: :GetFreeDevice, type: "string (string)"
```

I actually use [ripgrep](https://build.opensuse.org/package/show/openSUSE:Factory/ripgrep) which is very fast.

PR #794, opened 2018-08-09
Improve systemd socket detection

imobachgs:
* Introduce a new `SystemdSocketFinder` which speeds up the detection of
  service sockets.
* Does not fetch TriggeredBy anymore as it was not reliable.
* SystemdService.find_many_at_once does not return nil values

### DIFF
--- a/library/systemd/src/lib/yast2/systemd_socket_finder.rb
+++ b/library/systemd/src/lib/yast2/systemd_socket_finder.rb
@@ -39,6 +39,7 @@ module Yast2
   class SystemdSocketFinder
     # Returns the socket for a given service
     #
+    # @param service_name [String] Service name (without the `.service` extension)
     # @return [Yast2::SystemdSocketClass::Socket,nil]
     def for_service(service_name)
       socket_name = socket_name_for(service_name)
@@ -60,6 +61,8 @@ module Yast2
 
     # Builds a map between services and sockets
     #
+    # @note When more than one socket triggers the service, the last one will be used.
+    #
     # @return [Hash<String,String>] Sockets indexed by the name of the service they trigger
     def sockets_map
       return @sockets_map if @sockets_map
@@ -78,9 +81,11 @@ module Yast2
 
     # Returns the systemctl show command to get sockets details
     #
+    # @note The list is alphabetically ordered.
+    #
     # @return [String] systemctl show command
     def show_triggers_cmd
-      format(SHOW_TRIGGERS_CMD, unit_names: unit_names.join(" "))
+      format(SHOW_TRIGGERS_CMD, unit_names: unit_names.sort.join(" "))
     end
 
     # Returns the names of the socket units
--- a/library/systemd/src/lib/yast2/systemd_socket_finder.rb
+++ b/library/systemd/src/lib/yast2/systemd_socket_finder.rb
@@ -1,0 +1,100 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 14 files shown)


library/systemd/src/lib/yast2/systemd_socket_finder.rb
```diff
@@ -0,0 +1,100 @@
...
+  # systemd is not fully operational, this class just tries to find a socket named after the
+  # service.
+  #
+  # @example Finding a socket
+  #   finder = Yast2::SystemdSocketFinder.new
+  #   finder.for_service("cups").class # => Yast::SystemdSocketClass::Socket
+  class SystemdSocketFinder
+    # Returns the socket for a given service
+    #
+    # @return [Yast2::SystemdSocketClass::Socket,nil]
```
joseivanlopez: Please, document the parameter. I guess the `service_name` is the name of the service without extension, isn't it?

library/systemd/src/lib/yast2/systemd_socket_finder.rb
```diff
@@ -0,0 +1,100 @@
...
+    # @return [String,nil]
+    def socket_name_for(service_name)
+      return service_name if Yast::Stage.initial
+      sockets_map[service_name]
+    end
+
+    # Builds a map between services and sockets
+    #
+    # @return [Hash<String,String>] Sockets indexed by the name of the service they trigger
+    def sockets_map
```
joseivanlopez: What happens when a service is triggered by several sockets? I guess it takes the last socket.

library/systemd/src/modules/systemd_socket.rb
```diff
@@ -89,6 +90,14 @@ def all(propmap = {})
       sockets.select { |s| s.properties.supported? }
     end
 
+    # Returns the socket for a given service
+    #
+    # @return [Yast::SystemdSocketClass::Socket,nil]
```
joseivanlopez: Please, document the parameter.

library/systemd/test/systemd_socket_test.rb
```diff
@@ -43,6 +44,27 @@ module Yast
...
+
+        before do
+          allow(Yast2::SystemdSocketFinder).to receive(:new).and_return(finder)
+        end
+
+        it "returns the socket for the given service" do
+          expect(subject.for_service("cups")).to eq(socket)
+        end
+
+        context "when there is not associated socket" do
```
joseivanlopez: Typo: "there is not" => "there is no"

library/systemd/test/yast2/system_service_test.rb
```diff
@@ -94,14 +94,6 @@
         expect(system_service.service).to eq(systemd_service)
       end
     end
-
```
joseivanlopez: Why is this test leftover?

library/systemd/test/yast2/system_service_test.rb
```diff
@@ -129,18 +121,6 @@
       expect(system_services).to be_all(Yast2::SystemService)
       expect(system_services.map(&:service)).to eq([apparmor, cups])
     end
-
-    context "when some service is not found" do
```
joseivanlopez: I would add a test to reflect to happens when a service is not found.

library/systemd/src/lib/yast2/systemd_socket_finder.rb
imobachgs: Yes, that's it. I would document that behaviour, because it is something I would like to discuss for the future.

library/systemd/test/yast2/system_service_test.rb
imobachgs: Although this class it is not responsible for that anymore, you are right that we should cover that case somehow. So I've added an additional test to `SystemdService` which covers that case.

PR #799, opened 2018-08-14
Reorganize Yast2 service classes

dgdavid:
Part of https://trello.com/c/0tG0L56x/161-reorganize-yast2-service-classes


library/systemd/src/lib/yast2/systemd/service.rb
```diff
@@ -0,0 +1,216 @@
...
+    # Systemd.service unit control API
+    #
+    # @example How to use it in other yast libraries
+    #    require 'yast'
+    #    require 'yast2/systemd/service'
+    #
+    #    ## Get a service unit by its name
+    #    ## If the service unit can't be found, you'll get nil
+    #
+    #    service = Yast::Systemd::Service.find('sshd') # service unit object
```
imobachgs: It should be `Yast2::Systemd::Service`. Please, check other occurrences in this documentation.

library/systemd/src/lib/yast2/systemd/service.rb
```diff
@@ -0,0 +1,216 @@
+require "yast2/systemd/unit"
+
+module Yast2
+  module Systemd
+    # Represent a missed service
```
imobachgs: I would say *missing*.

library/systemd/src/lib/yast2/systemd/service.rb
```diff
@@ -0,0 +1,216 @@
...
+      Yast.import "Stage"
+      include Yast::Logger
+
+      UNIT_SUFFIX = ".service".freeze
+
+      class << self
+        # @param service_name [String] "foo" or "foo.service"
+        # @param propmap [Systemd::Unit::PropMap]
+        # @return [Service,nil] `nil` if not found
+        def find(service_name, propmap = {})
```
imobachgs: It looks strange that `propmap` is expected to be an instance of `Yast2::Systemd::Unit::Propmap` but the default value is a hash.

library/systemd/src/lib/yast2/systemd/service.rb
```diff
@@ -0,0 +1,216 @@
...
+          return services unless services.empty?
+
+          log.info "Retrying one by one"
+          service_names.map { |n| find(n, propmap) }
+        end
+
+        # @param propmap [Systemd::Unit::PropMap]
+        # @return [Array<Service>]
+        def all(propmap = {})
+          Systemctl.service_units.map { |service_unit| new(service_unit, propmap) }
```
imobachgs: Please, follow the [SUSE Ruby style guide](https://github.com/SUSE/style-guides/blob/master/Ruby.md) and use a single letter identifier (see the [parameters](https://github.com/SUSE/style-guides/blob/master/Ruby.md#parameters) section).

library/systemd/src/lib/yast2/systemd/service.rb
```diff
@@ -0,0 +1,216 @@
...
+        @socket ||= Yast2::Systemd::Socket.for_service(name)
+      end
+
+      # Determines whether the service has an associated socket
+      #
+      # @return [Boolean] true if an associated socket exists; false otherwise.
+      def socket?
+        !socket.nil?
+      end
+
```
imobachgs: `#installation_system?` and `#run_instsys_command` used to be private. Any reason to change that? (I am not against, only curiosity).

See https://github.com/yast/yast-yast2/blob/master/library/systemd/src/modules/systemd_service.rb#L202

library/systemd/src/lib/yast2/systemd/socket.rb
```diff
@@ -0,0 +1,110 @@
...
+        end
+
+        # @param propmap [SystemdUnit::PropMap]
+        def find!(socket_name, propmap = {})
+          find(socket_name, propmap) || raise(Systemd::SocketNotFound, socket_name)
+        end
+
+        # @param propmap [SystemdUnit::PropMap]
+        def all(propmap = {})
+          sockets = Systemctl.socket_units.map { |socket_unit| new(socket_unit, propmap) }
```
imobachgs: Please, use a 1-letter identifier.

library/systemd/src/lib/yast2/systemd/unit.rb
```diff
@@ -0,0 +1,356 @@
...
+
+module Yast2
+  module Systemd
+    ###
+    #  Use this class always as a parent class for implementing various systemd units.
+    #  Do not use it directly for ad-hoc implementation of systemd units.
+    #
+    #  @example Create a systemd service unit
+    #
+    #     class Service < Yast::Systemd::Unit
```
imobachgs: `Yast2::Systemd::Unit`

library/systemd/src/lib/yast2/systemd/unit.rb
```diff
@@ -0,0 +1,356 @@
...
+      def run_command!
+        error.clear
+        command_result = yield
+        error << command_result.stderr
+        refresh!
+        command_result.exit.zero?
+      end
+
+      # Structure holding  properties of systemd unit
+      class Properties < OpenStruct
```
imobachgs: What about moving this class to its own file?

library/systemd/src/lib/yast2/systemd/unit.rb
```diff
@@ -0,0 +1,356 @@
...
+        def read_enabled_state
+          # If UnitFileState property is missing (due to e.g. legacy sysvinit service) or
+          # has an unknown entry (e.g. "bad") we must use a different way how to get the
+          # real status of the service.
+          if SUPPORTED_STATES.include?(unit_file_state)
+            state_name_enabled?(unit_file_state)
+          else
+            # Check for exit code of `systemctl is-enabled systemd_unit.name` ; additionally
+            # test the stdout of the command for valid values when the service is enabled
+            # http://www.freedesktop.org/software/systemd/man/systemctl.html#is-enabled%20NAME...
```
imobachgs: The correct link has changed to `http://www.freedesktop.org/software/systemd/man/systemctl.html#is-enabled UNIT…` (not nice anchor name IMHO).

library/systemd/src/lib/yast2/systemd/unit.rb
```diff
@@ -0,0 +1,356 @@
...
+      # 1. Checks for existence of the unit based on the stderr from the command
+      #    `systemctl is-enabled`
+      # 2. Retrieves the status enabled|disabled which is needed in the installation
+      #    system. There are currently only 3 commands available for systemd in
+      #    inst-sys/chroot: `systemctl enable|disable|is-enabled`. The rest will return
+      #    the error message mentioned above in this comment.
+      #
+      # Once the inst-sys has running dbus/systemd, this class definition can be removed
+      # together with the condition for Stage.initial in the Systemd::Unit#show.
+      class InstallationProperties < OpenStruct
```
imobachgs: Please, move this into its own file too.

library/systemd/src/lib/yast2/systemd/service.rb
dgdavid: I agree. Shall we change it now? WDYT?

imobachgs: No, as a Propmap is basically a Hash, let's keep it out of the scope of this PR.

library/systemd/src/lib/yast2/systemd/service.rb
dgdavid: No reason at all. I guess I deleted the `private` key word accidentally :innocent: The good new it that I am going to restore it ;)

dgdavid: BTW, Thank you :)

library/systemd/src/lib/yast2/systemd/service.rb
dgdavid: Ups, sorry. I did not read that, and [it is already changed at the same time to extract related `Yast2::Systemd::Unit` classes to their own files](https://github.com/yast/yast-yast2/pull/799/commits/00c3fb703d2db27f25ec546f9b0ef1803571e13e).

dgdavid:
You're right @imobachgs 

Thank you

PR #800, opened 2018-08-14
Add support for on-demand only services

imobachgs:
Fixes [bsc#1104568](https://bugzilla.suse.com/show_bug.cgi?id=1104568).

### DIFF
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -239,7 +239,8 @@ module Yast2
     def start_modes
       @start_modes = [:on_boot, :manual, :on_demand] unless found?
       return @start_modes if @start_modes
-      @start_modes = [:on_boot, :manual]
+      @start_modes = [:manual]
+      @start_modes << :on_boot unless service.static?
       @start_modes << :on_demand if socket
       @start_modes
     end
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -434,9 +434,9 @@ module Yast2
         when :on_boot
           service.enable && (socket ? socket.disable : true)
         when :on_demand
-          service.disable && (socket ? socket.enable : false)
+          disable_service && (socket ? socket.enable : false)
         when :manual
-          service.disable && (socket ? socket.disable : true)
+          disable_service && (socket ? socket.disable : true)
         end
 
       register_error(:start_mode) unless result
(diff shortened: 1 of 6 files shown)


library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -239,7 +239,8 @@ def currently_active?
     def start_modes
       @start_modes = [:on_boot, :manual, :on_demand] unless found?
       return @start_modes if @start_modes
-      @start_modes = [:on_boot, :manual]
+      @start_modes = [:manual]
+      @start_modes << :on_boot unless service.static?
```
jreidinger: does it make sense at all? If service is static, then it will always run like udev. so for me different should be on_boot and on_demand, but even for on_demand e.g. dbus will start automatic during system boot.

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -254,10 +254,10 @@ def initialize(systemd_unit, property_text)
...
         extract_properties
         self[:active?]     = ACTIVE_STATES.include?(active_state)
-        self[:running?]    = sub_state    == "running"
-        self[:loaded?]     = load_state   == "loaded"
-        self[:not_found?]  = load_state   == "not-found"
-        self[:static?]     = load_state   == "static"
+        self[:running?]    = sub_state  == "running"
+        self[:loaded?]     = load_state == "loaded"
+        self[:not_found?]  = load_state == "not-found"
+        self[:static?]     = unit_file_state == "static"
```
jreidinger: why this change is needed?

library/systemd/test/yast2/system_service_test.rb
```diff
@@ -302,6 +304,14 @@
       end
     end
 
+    context "when the service has no associated socket" do
```
jreidinger: this looks like wrong description

library/systemd/src/lib/yast2/systemd_unit.rb
imobachgs: Because it was basically wrong.

library/systemd/src/lib/yast2/system_service.rb
imobachgs: Think, for instance, about tftpd.Perhaps you do not want that service to be started.

jreidinger: well, documentation says that static services cannot be enabled neither disabled as it missing "Install" section. It means it can be run only as depency either for socket or by other services ( like udev ). So I am not sure if manual is right one, but neither on_boot sounds good for me. But for now we do not have anything better, so lets keep it with manual. But please ensure that we will never run enable/disable on it as it won't work.

PR #804, opened 2018-08-21
[RFC] Add exception to distinguish a `systemctl show` fail from others

dgdavid:
Related to https://trello.com/c/XZUCetog/160-create-new-exception-for-failure-when-refreshing-a-service

### DIFF
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -354,9 +354,15 @@ module Yast2
 
       result = errors.none?
 
-      result && reset && refresh
+      if result
+        reset && refresh
+      end
 
       result
+    rescue Yast::SystemctlError
+      register_error(:active)
+
+      false
     end
 
     # Reverts cached changes
@@ -456,9 +462,8 @@ module Yast2
       result = send("perform_#{action}")
       register_error(:active) if result == false
       result
-    rescue => e
-      register_error(:active) if e.kind_of?(Yast::SystemctlError)
-      false
+    rescue Yast::CouldNotRefreshUnitError
+      true
     end
 
     # Starts the service in the underlying system
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -342,10 +342,8 @@ module Yast2
 
     # Saves changes into the underlying system
     #
-    # @note All cached changes are reset and the underlying service is refreshed
-    #   when the changes are correctly applied.
-    #
-    # @raise [Yast::SystemctlError] if the service cannot be refreshed
+    # @note All cached changes are reset and the underlying service will be refreshed when the changes
+    #   are correctly applied.
     #
     # @param keep_state [Boolean] Do not change service status. Useful when running on 1st stage.
     # @return [Boolean] true if the service was saved correctly; false otherwise.
@@ -354,7 +352,11 @@ module Yast2
       save_start_mode
       perform_action unless keep_state
 
-      errors.none? && reset && refresh!
+      result = errors.none?
+
+      result && reset && refresh
+
+      result
     end
 
     # Reverts cached changes
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -375,7 +375,7 @@ module Yast2
     # @return [Boolean] true if the service was refreshed correctly; false otherwise.
     def refresh
       refresh!
-    rescue Yast::SystemctlError
+    rescue Yast::CouldNotRefreshUnitError
       false
     end
 
@@ -453,14 +453,9 @@ module Yast2
 
       result = send("perform_#{action}")
       register_error(:active) if result == false
-
       result
-
-      # FIXME: SystemdService#{start, stop, etc} calls to refresh! internally, so when
-      # this exception is raised we cannot distinguish if the action is failing or
-      # refresh! is failing. For SP1, refresh! should raise a new kind of exception.
-    rescue Yast::SystemctlError
-      register_error(:active)
+    rescue => e
+      register_error(:active) if e.kind_of?(Yast::SystemctlError)
       false
     end
 
--- a/library/systemd/src/lib/yast2/systemd_unit.rb
+++ b/library/systemd/src/lib/yast2/systemd_unit.rb
@@ -3,6 +3,14 @@ require "yast2/systemctl"
 require "ostruct"
 
 module Yast
+  # Represent a fail when systemctl command tries to refresh the service
+  class CouldNotRefreshUnitError < StandardError
+    # @param unit [Yast::SystemdUnit]
+    def initialize(unit)
+      super "Could not refresh #{unit.name}"
+    end
+  end
+
   ###
   #  Use this class always as a parent class for implementing various systemd units.
   #  Do not use it directly for ad-hoc implementation of systemd units.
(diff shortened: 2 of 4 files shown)


dgdavid:
Guys, this PR it is only to discuss it this changes are enough to solve the https://trello.com/c/XZUCetog/160-create-new-exception-for-failure-when-refreshing-a-service or we should take the opportunity to improve the design of `Yast::Systemctl#execute` as we were talking in the #yast freenode channgel.

If so, could you please give me a hand in order to not break everything related to `Yast::Systemctl#execute`?

dgdavid:
Additionally, I must say that I do not want to merge this PR, because in my opinion it is better to finish #799 first and do the changes in a new branch from master after merge it.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -354,7 +352,11 @@ def save(keep_state: false)
       save_start_mode
       perform_action unless keep_state
 
-      errors.none? && reset && refresh!
+      result = errors.none?
+
+      result && reset && refresh
```
jreidinger: I prefer more explicit `if result` and not hide it inside boolean expression.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -453,14 +455,9 @@ def perform_action
...
-
       result
-
-      # FIXME: SystemdService#{start, stop, etc} calls to refresh! internally, so when
-      # this exception is raised we cannot distinguish if the action is failing or
-      # refresh! is failing. For SP1, refresh! should raise a new kind of exception.
-    rescue Yast::SystemctlError
-      register_error(:active)
+    rescue => e
+      register_error(:active) if e.kind_of?(Yast::SystemctlError)
```
jreidinger: Hmm, I do not like it. It basically catch all errors instead of systemctl ones, is it correct?

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -3,6 +3,14 @@
 require "ostruct"
 
 module Yast
+  # Represent a fail when systemctl command tries to refresh the service
+  class CouldNotRefreshUnitError < StandardError
```
jreidinger: just thinking about it, why not having some reasonable ancestor for all that systemd errors?

jreidinger: or service errors?

library/systemd/src/lib/yast2/systemd_unit.rb
```diff
@@ -121,11 +129,13 @@ def initialize(full_unit_name, propmap = {}, property_text = nil)
...
     end
 
-    # @raise [Yast::SystemctlError] if 'systemctl show' cannot be executed
+    # @raise [Yast::CouldNotRefreshUnitError] if 'systemctl show' cannot be executed
     def refresh!
       @properties = show
       @error = properties.error
       properties
+    rescue Yast::SystemctlError
+      raise Yast::CouldNotRefreshUnitError, self
```
jreidinger: shouldn't it set also `@error` when show failed? Otherwise I worry we will be in an inconsistent state.

jreidinger:
@dgdavid fine for me to having it after merge as it can need long discussion about how exception structure for system service should look like.

library/systemd/src/lib/yast2/system_service.rb
dgdavid: I agree

library/systemd/src/lib/yast2/system_service.rb
dgdavid: It seem to me, according to the expected result when  `#save` is called. If I am wrong, the other approach could be

```ruby
def save
...
rescue Yast::SystemctlError
...
rescue Yast::CouldNotRefreshUnitError
...
end
```

library/systemd/src/lib/yast2/systemd_unit.rb
dgdavid: Good catch. 

I am not sure about what to do.

@joseivanlopez, thoughts?

library/systemd/src/lib/yast2/systemd_unit.rb
dgdavid: Make sense to me, as long as we decide to cover more scenarios adding more specific exceptions.

library/systemd/src/lib/yast2/system_service.rb
joseivanlopez: I expect that `#perform_action` only catches refresh exception. The idea is to not fail if the action was performed but refresh fails, so I would do:

```
rescue Yast::SystemctlRefreshError
  true
```

library/systemd/src/lib/yast2/systemd_unit.rb
joseivanlopez: Yes, I agree with @jreidinger. I would create a new exception based on existing `SystemctlError`, for example:

```
class SystemctlRefreshError < SystemctlError; end
```

Using a subclass we ensure that all code catching current `SystemctlError` will continue catching the new exception `SystemctlRefreshError`.

library/systemd/src/lib/yast2/system_service.rb
jreidinger: better or at least use `rescue Yast::SystemctlError, Yast::CouldNotRefreshUnitError => e` so it don't catch other problems and silent them.

joseivanlopez: In other words, `#perform_action` should continue raising the `SystemcltError` when the action is not performed.

library/systemd/src/lib/yast2/systemd_unit.rb
joseivanlopez: Something like this? `@error = "Fails to refresh unit"` ?

library/systemd/src/lib/yast2/system_service.rb
dgdavid: But, what about to `register_error`?

library/systemd/src/lib/yast2/systemd_unit.rb
dgdavid: Okey, thank you

library/systemd/src/lib/yast2/system_service.rb
dgdavid: Okey, I moved `register_error` to `#save` method.

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -453,15 +461,9 @@ def perform_action
...
       result
-
-      # FIXME: SystemdService#{start, stop, etc} calls to refresh! internally, so when
-      # this exception is raised we cannot distinguish if the action is failing or
-      # refresh! is failing. For SP1, refresh! should raise a new kind of exception.
-    rescue Yast::SystemctlError
-      register_error(:active)
-      false
+    rescue Yast::CouldNotRefreshUnitError
+      true
```
jreidinger: and what happens when result is false so action failed and also following refresh failed?

dgdavid: Ouch! Nice catch. Let me think about it.

dgdavid:
After discuss this a little more deep offline, we concluded that for now it is not really necessary to distinguish that exception because, for the shake of consistency, the `#perform_action` method must be return `false` as soon as something went wrong.

So, the only change that we will to do, in a new PR, is to rescue the `SystemctlError` also in the `#save_start_mode`.

PR #806, opened 2018-08-28
Drop SuSEfirewall2 completely

teclator:
Drop old SUSEFirewalld2 code which should not be used anymore.

See https://github.com/yast/yast-yast2/pull/806#issuecomment-416569010

PR #807, opened 2018-08-29
[WIP] Added support for modifying :short, :description and :target zone att…

teclator:
…ributes

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld.rb
+++ b/library/network/src/lib/y2firewall/firewalld.rb
@@ -83,7 +83,7 @@ module Y2Firewall
     # @return [Boolean] true
     def read
       return false unless installed?
-      @zones = ZoneParser.new(api.zones, api.list_all_zones).parse
+      @zones = ZoneParser.new(api.zones, api.list_all_zones(verbose: true)).parse
       @log_denied_packets = api.log_denied_packets
       @default_zone       = api.default_zone
       # The list of services is not read or initialized because takes time and
--- a/library/network/src/lib/y2firewall/firewalld/api/zones.rb
+++ b/library/network/src/lib/y2firewall/firewalld/api/zones.rb
@@ -73,15 +73,23 @@ module Y2Firewall
         # @param permanent [Boolean] if true it adds the --permanent option the
         # command to be executed
         # @return [Array<String>] list of all information for given zone
-        def list_all(zone, permanent: permanent?)
-          string_command("--zone=#{zone}", "--list-all", permanent: permanent).split(" ")
+        def list_all(zone, permanent: permanent?, verbose: false)
+          if verbose
+            string_command("--zone=#{zone}", "--verbose", "--list-all", permanent: permanent).split(" ")
+          else
+            string_command("--zone=#{zone}", "--list-all", permanent: permanent).split(" ")
+          end
         end
 
         # @param permanent [Boolean] if true it adds the --permanent option the
         # command to be executed
         # @return [Array<String>] list of all information for all firewall zones
-        def list_all_zones(permanent: permanent?)
-          string_command("--list-all-zones", permanent: permanent).split("\n")
+        def list_all_zones(permanent: permanent?, verbose: false)
+          if verbose
+            string_command("--list-all-zones", "--verbose", permanent: permanent).split("\n")
+          else
+            string_command("--list-all-zones", permanent: permanent).split("\n")
+          end
         end
 
         ### Interfaces ###
@@ -258,6 +266,30 @@ module Y2Firewall
           return true if !masquerade_enabled?(zone)
 
           run_command("--zone=#{zone}", "--remove-masquerade")
+        end
+
+        def short(zone)
+          string_command("--zone=#{zone}", "--get-short")
+        end
+
+        def short=(zone, short_description)
+          string_command("--zone=#{zone}", "--set-short=#{short_description}")
+        end
+
+        def description(zone)
+          string_command("--zone=#{zone}", "--get-description")
+        end
+
+        def description=(zone, long_description)
+          run_command("--zone=#{zone}", "--set-description=#{long_description}")
+        end
+
+        def target(zone)
+          string_command("--zone=#{zone}", "--get-target")
+        end
+
+        def target=(zone,target)
+          run_command("--zone=#{zone}", "--set-target=#{target}")
         end
       end
     end
--- a/library/network/src/lib/y2firewall/firewalld/zone_parser.rb
+++ b/library/network/src/lib/y2firewall/firewalld/zone_parser.rb
@@ -51,7 +51,6 @@ module Y2Firewall
         @zones_definition.reject(&:empty?).each do |line|
           attribute, _value = line.split("\s")
           next if !attribute
-
           if @zone_names.include?(attribute)
             zone = Zone.new(name: attribute)
             zones << zone
@@ -61,10 +60,14 @@ module Y2Firewall
           next unless zone
 
           attribute, value = line.lstrip.split(":\s")
+          attribute = "short" if attribute == "summary"
+          attribute = "rich_rules" if attribute == "rich rules"
 
           next unless zone.respond_to?("#{attribute}=")
           if BOOLEAN_ATTRIBUTES.include?(attribute)
             zone.public_send("#{attribute}=", value == "yes" ? true : false)
+          elsif zone.attributes.include?(attribute.to_sym)
+            zone.public_send("#{attribute}=", value.to_s)
           else
             zone.public_send("#{attribute}=", value.to_s.split)
           end
--- a/library/network/src/lib/y2firewall/firewalld/relations.rb
+++ b/library/network/src/lib/y2firewall/firewalld/relations.rb
@@ -31,6 +31,35 @@ module  Y2Firewall
         end
       end
 
+      # @param attributes [Array<Symbol] relation or attribute names
+      def has_attribute(*attributes, cache: false)
+        define_method "attributes" do
+          attributes
+        end
+
+        attributes.each do |attribute|
+          attr_reader attribute
+
+          define_method "#{attribute}=" do |item|
+            instance_variable_set("@#{attribute}", item)
+
+            modified!(attribute) if cache
+          end
+
+          define_method "current_#{attribute}" do
+            api.public_send(attribute, name)
+          end
+        end
+
+        define_method "apply_attributes_changes!" do
+          attributes.each do |attribute|
+            next if cache && !modified?(attribute)
+            api.public_send("#{attribute}=", name, public_send(attribute))
+          end
+          true
+        end
+      end
+
       # Defines a set of methods to operate over array based firewalld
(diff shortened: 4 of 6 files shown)


library/network/src/lib/y2firewall/firewalld/api/zones.rb
```diff
@@ -342,48 +342,73 @@ def remove_protocol(zone, protocol, permanent: permanent?)
...
-          run_command("--zone=#{zone}", "--remove-masquerade")
+          run_command("--zone=#{zone}", "--remove-masquerade", permanent: permanent)
         end
 
+        # Full name or short description of the zone
+        #
+        # @param zone [String] The firewall zone
         def short(zone)
-          string_command("--zone=#{zone}", "--get-short")
+          string_command("--zone=#{zone}", "--get-short", permanent: offline? ? false : true)
```
jreidinger: why not `permanent: !offline?` ?

library/network/src/lib/y2firewall/firewalld/api/services.rb
```diff
@@ -44,8 +44,12 @@ def services
         # @param permanent [Boolean] if true it adds the --permanent option the
         # command to be executed
         # @return [Array<String>] list of all information for the given service
-        def info_service(service, permanent: permanent?)
-          string_command("--info-service", service.to_s, permanent: permanent).split("\n")
+        def info_service(service, permanent: permanent?, verbose: false)
+          if verbose
+            string_command("--info-service=#{service}", "--verbose", permanent: permanent).split("\n")
```
imobachgs: What about building an array (and inserting the `--verbose` string if needed) and having only one call to `string_command`?

library/network/src/lib/y2firewall/firewalld/api/zones.rb
```diff
@@ -69,19 +85,51 @@ def list_sources(zone, permanent: permanent?)
           string_command("--zone=#{zone}", "--list-sources", permanent: permanent).split(" ")
         end
 
+        # @param zone [String] The firewall zone
+        # @param permanent [Boolean] if true it adds the --permanent option the
```
imobachgs: I would omit mentioning `--permanent`. What about something like `if true it modifies the "permanent" configuration`? (or something similar).

library/network/src/lib/y2firewall/firewalld/api/zones.rb
```diff
@@ -69,19 +85,51 @@ def list_sources(zone, permanent: permanent?)
...
+        # command to be executed
```
imobachgs: Please, indent this line. Otherwise, the documentation won't be rendered correctly (same for all the "command to be executed" lines in this file).

library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -83,14 +89,33 @@ def initialize
...
       @log_denied_packets = api.log_denied_packets
       @default_zone       = api.default_zone
       # The list of services is not read or initialized because takes time and
       # affects to the performance and also the services are rarely touched.
       @read = true
     end
 
+    # Add a not existent zones to the list of zones
+    #
+    # @return [Boolean] true if a new one is added; false otherwise
```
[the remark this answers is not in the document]
teclator: Well, it says that returns true only if a new one is added :), but will try to clarify it a little bit :+1:

library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -143,12 +165,32 @@ def write
...
+    # delete all the new or removed zones depending on each case.
+    def apply_zones_changes!
+      zones.each do |zone|
+        api.create_zone(zone.name) unless current_zone_names.include?(zone.name)
+        zone.apply_changes! if zone.modified?
+      end
+      current_zone_names.each do |name|
+        api.delete_zone(name) unless zones.any? { |zone| zone.name == name }
+      end
+      true
```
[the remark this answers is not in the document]
teclator: I just returned true by now when all the changes or no changes were already made. I'm still not sure if we would like to handle / rescue from some exception in the future or return false in some circunstances

library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -215,5 +257,26 @@ def start
     def read?
       @read
```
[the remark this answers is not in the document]
teclator: Well, take in account that there is also the method read

library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -215,5 +257,26 @@ def start
...
     end
+
+    # Convenience method for initializing and retrieving an API instance
+    def api
+      @api ||= Api.new
+    end
+
+  private
+
+    # Convenience method for instantiate a new zone parser
```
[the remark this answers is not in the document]
teclator: Probably I'm :see_no_evil: I wil check it

library/network/src/lib/y2firewall/firewalld/api.rb
```diff
@@ -109,7 +109,7 @@ def disable!
       # @return [String] firewalld service state
       # @see http://www.firewalld.org/documentation/man-pages/firewall-cmd.html
       def state
-        case Yast::Execute.on_target("firewall-cmd", "--state", allowed_exitstatus: [0, 252])
+        case Yast::Execute.on_target!("firewall-cmd", "--state", allowed_exitstatus: [0, 252])
```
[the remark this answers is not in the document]
teclator: I changed it to start using the bang for not failing silently in some cases and specially with the test units not mocked properly, maybe I will put back the call without the bang but just in this case, In run_command I thing we should extend the documentation

library/network/src/lib/y2firewall/firewalld/api/services.rb
```diff
@@ -44,8 +44,12 @@ def services
         # @param permanent [Boolean] if true it adds the --permanent option the
         # command to be executed
         # @return [Array<String>] list of all information for the given service
-        def info_service(service, permanent: permanent?)
-          string_command("--info-service", service.to_s, permanent: permanent).split("\n")
+        def info_service(service, permanent: permanent?, verbose: false)
```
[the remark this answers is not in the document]
teclator: The verbose option adds the `summary` (`short name`) and the `description` to the output. The same is applied to zones. It is the default in `firewalld`, and it probably would be used for a simple parse which can just parse the config in brief mode, although the impact is so minimal that we could hide this parameter

library/network/src/lib/y2firewall/firewalld/api/services.rb
teclator: Yep, I already was thinking on that

library/network/src/lib/y2firewall/firewalld/zone_parser.rb
```diff
@@ -46,33 +51,69 @@ def initialize(zone_names, zones_definition)
...
       def parse
         return [] if !@zone_names || @zone_names.empty?
-        zone = nil
+        parse_zones
+        initialize_zones
+      end
+
+    private
+
+      def initialize_zones
```
imobachgs: I know it is a private method, but it looks complex enough to be documented.

library/network/src/lib/y2firewall/firewalld/zone_parser.rb
```diff
@@ -46,33 +51,69 @@ def initialize(zone_names, zones_definition)
...
           end
 
-          next unless zone
+          zone.untouched!
+        end
+
+        zones
+      end
+
+      def parse_zones
```
imobachgs: Please, document it.

library/network/test/y2firewall/firewalld/service_parser_test.rb
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+#
+# Copyright (c) [2017] SUSE LLC
```
imobachgs: Please, year++ :)

library/network/test/y2firewall/firewalld/service_parser_test.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  before do
+    allow(firewalld).to receive(:api).and_return(api)
+    `echo true`
+  end
+
+  describe "#parse" do
+    let(:service_info) do
+      [
+        "radius",
```
imobachgs: Parsing this stuff looks pretty fragile to me.

library/network/test/y2firewall/firewalld/service_parser_test.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    context "when the service is not present" do
+      let(:service_name) { "not_present" }
+      before do
+        allow(api).to receive(:info_service).with(service_name)
+        allow($CHILD_STATUS).to receive(:exitstatus).and_return(101)
+      end
+
+      it "raises a non Found exception" do
```
imobachgs: NotFound (or not found)

library/network/src/lib/y2firewall/firewalld/api/zones.rb
```diff
@@ -240,24 +342,73 @@ def remove_protocol(zone, protocol, permanent: permanent?)
...
+
+        # Long description of the zone
+        #
+        # @param zone [String] The firewall zone
+        def description(zone)
+          string_command("--zone=#{zone}", "--get-description",
+            permanent: !offline?)
+        end
+
+        # Modiy the long description of the zone
```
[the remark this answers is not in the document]
teclator: I fear that it will not be possible.

library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -176,6 +176,8 @@ def write_only
       read unless read?
       apply_zones_changes!
       apply_attributes_changes!
+      # remove zones that is not in new zones map
+      (api.zones - zones.map(&:name)).each { |n| api.delete_zone(n) }
```
teclator: **Note:** api calls are slow, and we avoid them as much as possible.  You can use current_zone_names already

Also, this responsibility was originally added to the `apply_zones_changes!` method hidden it from the object itself.
```ruby
      zones.each do |zone|
        api.create_zone(zone.name) unless current_zone_names.include?(zone.name)
        zone.apply_changes! if zone.modified?
      end
      current_zone_names.each do |name|
        api.delete_zone(name) if !zones.any? { |z| z.name == name }
      end
```

```

teclator:
It looks ok,maybe we could also drop the old SuSEFirewall classes although it can be done in a separate PR. 

So probably what is missing is the check of the builtins zones in case of removing them as we were discussing, not just in the interface but also in the API.

jreidinger:
@teclator well, in API result is that it will show popup that builtin cannot be removed, which is ok for me as it is not silent failure.

teclator:
@jreidinger https://github.com/yast/yast-yast2/pull/806 and this was my PR for dropping the old code, so will try to rebase and prepare for merging after this one.

teclator:
@jreidinger although I miss the version and changelog entries

PR #808, opened 2018-08-30
Increase the timeout for systemctl command execution

dgdavid:
> Since systemd by itself has a 30 sec timeout, YaST should add some more
> seconds to allow for systemctl to exit gracefully even if it has a
> timeout and not just kill the systemctl command just one seconds before
> it would time out itself.

See https://bugzilla.suse.com/show_bug.cgi?id=1098910#c3

---

Related to https://trello.com/c/udasbOtt/135-lhf-osdistribution-p4-1098910-yast2-services-manager-internal-error-writing-run-configuration-to-disk

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 30 11:21:14 UTC 2018 - David Díaz <dgonzalez@suse.com>
+
+- Increase timeout for the execution of systemctl commands
+  (bsc#1098910).
+- 4.0.86
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Aug 30 11:21:14 UTC 2018 - David Díaz <dgonzalez@suse.com>
+
+- Increase timeout for the execution of systemctl commands
```
joseivanlopez: Please, mention bug number.

dgdavid: Ops! :sweat_smile: Thank you :)

PR #813, opened 2018-09-03
[WIP] Firewalld fixes for the new UI

mvidner:
- for https://trello.com/c/Xcwdi3XK
- related to https://github.com/yast/yast-firewall/pull/71

TODO: when firewalld is running, the unit tests cause a Polkit pop-up


teclator:
I would expect in both cases to be already read although adding this extra checks probably will be more secure.

mvidner:
This is a robustness improvement that did not make it in its time. RIP.

PR #817, opened 2018-09-07
Backport changes

joseivanlopez:


kobliha:
Guys, what about translations? I don't know when exactly is this used, but the text is hard-coded as it seems. Or did I miss something?

```Yast::Popup.YesNo("Really go next?")```

PR #818, opened 2018-09-10
Extended API supporting sinle-value attributes

teclator:
It basically takes most of the changes added in #807 except the creation or deletion of zones, and the definition of more complex relations.


library/network/src/lib/y2firewall/firewalld/zone_reader.rb
```diff
@@ -0,0 +1,132 @@
...
+      # Iterates over the zones definition filling the zone entries with the
+      # parsed information of each zone
+      def parse_zones
+        current_zone = nil
+        current_attribute = nil
+        zones_definition.each do |line|
+          next if line.lstrip.empty?
+          # If  the entry looks like a zone name
+          if line.start_with?(/\w/)
+            attribute, _value = line.split(/\s*\(active\)\s*$/)
```
imobachgs: What about `line.split(" ")`?

library/network/src/lib/y2firewall/firewalld/zone_reader.rb
```diff
@@ -0,0 +1,132 @@
...
+            attribute = nil unless zone_names.include?(attribute)
+            current_zone = attribute
```
imobachgs: If the `attribute` value is not relevant anymore for this iteration (when the zone name is not included in `zone_names`), I would write something like `current_zone = zone_names.include?(attribute) ? attribute : nil`.

library/network/src/lib/y2firewall/firewalld/zone_reader.rb
```diff
@@ -0,0 +1,132 @@
...
+        end
+      end
+
+      # Iterates over the zone entries instantiating a zone object per each of
+      # the entries and returning an array with all of them.
+      #
+      # @return [Array<Y2Firewall::Firewalld::Zone] the list of zones obtained
+      #   from the parsed definition
+      def initialize_zones
+        zones = []
```
imobachgs: What about `each_with_object`?

library/network/src/lib/y2firewall/firewalld/zone_reader.rb
```diff
@@ -0,0 +1,132 @@
...
+        zone_entries.each do |name, config|
+          zone = Zone.new(name: name)
+          zones << zone
+
+          config.each do |attribute, entries|
+            attribute = "short" if attribute == "summary"
```
imobachgs: What about using a hash to map these attributes?

library/network/src/lib/y2firewall/firewalld/zone_reader.rb
```diff
@@ -0,0 +1,132 @@
...
+      #
+      # @return [Array<Y2Firewall::Firewalld::Zone] the list of zones obtained
+      #   from the parsed definition
+      def initialize_zones
+        zones = []
+        zone_entries.each do |name, config|
+          zone = Zone.new(name: name)
+          zones << zone
+
+          config.each do |attribute, entries|
```
imobachgs: What about moving this part to a separate method?

library/network/test/y2firewall/firewalld/service_reader_test.rb
```diff
@@ -0,0 +1,81 @@
...
+require "y2firewall/firewalld"
+require "y2firewall/firewalld/service_reader"
+
+describe Y2Firewall::Firewalld::ServiceReader do
+  let(:firewalld) { Y2Firewall::Firewalld.instance }
+  let(:api) { instance_double("Y2Firewall::Firewalld::Api", state: "not_running") }
+
+  before do
+    allow(firewalld).to receive(:api).and_return(api)
+    `echo true`
```
imobachgs: What is this line?

library/network/test/y2firewall/firewalld/service_reader_test.rb
```diff
@@ -0,0 +1,81 @@
...
+        allow(api).to receive(:info_service).with(service_name)
+          .and_return(service_info)
+        allow($CHILD_STATUS).to receive(:exitstatus).and_return(1)
+      end
+
+      it "returns the service with the parsed configuration" do
+        service = subject.read(service_name)
+        expect(service.short).to eq("RADIUS")
+        expect(service.ports).to eq(["1812/tcp", "1812/udp", "1813/tcp", "1813/udp"])
+        expect(service.protocols).to eq([])
```
imobachgs: Adding some protocol would be nice. Otherwise, we would not know whether the protocol parsing is working or not.

library/network/test/y2firewall/firewalld_test.rb
```diff
@@ -218,6 +233,18 @@
...
+    it "stores the list of available zone names" do
+      expect(firewalld.current_zone_names).to eq([])
+      firewalld.read
+      expect(firewalld.current_zone_names).to eq(known_zones)
+    end
+
+    it "stores the list of available service names" do
+      expect(firewalld.current_service_names).to eq([])
+      firewalld.read
+      expect(firewalld.current_service_names).to eq(known_services)
```
imobachgs: np: You could use the `#change` method:

```ruby
expect { firewalld.read }.to change { firewall.current_service_names }.from([]).to(known_services)
```

But it is up to you, as for some people it might be less readable.

library/network/test/y2firewall/relations_test.rb
```diff
@@ -0,0 +1,96 @@
...
+    subject { Dummy.new }
+
+    let(:api) do
+      instance_double("Y2Firewall::Firewalld::API",
+        description: "dummy text", dummies: ["john", "doe"])
+    end
+
+    before { allow(subject).to receive(:api).and_return(api) }
+
+    it "defines an getter an setter for each given attribute" do
```
imobachgs: an -> and a

library/network/test/y2firewall/relations_test.rb
```diff
@@ -0,0 +1,96 @@
...
+      expect(subject.respond_to?("description")).to eq(true)
+      expect(subject.respond_to?("description=")).to eq(true)
+    end
+
+    it "defines the 'attributes' method" do
+      expect(subject.attributes).to eq([:description])
+    end
+
+    it "defines the \"current_'attribute}'\" method" do
```
imobachgs: I guess the `}` is just a typo.

library/network/test/y2firewall/relations_test.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    it "defines the 'read_attributes' method" do
+      subject.read_attributes
+      expect(subject.description).to eq("dummy text")
+    end
+
+    it "defines the 'apply_attributes_changes!' method" do
+      subject.description = "modified dummy text"
+      expect(api).to receive(:public_send).with("description=", "modified dummy text")
```
imobachgs: Why an expect? If you are not testing this, it should be an `allow`.

library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -143,12 +144,18 @@ def write
...
+      apply_zones_changes!
+      apply_attributes_changes!
+      untouched!
       true
     end
 
+    # Apply the changes done in each of the modified zones. It will create or
+    # delete all the new or removed zones depending on each case.
+    def apply_zones_changes!
+      zones.each { |z| z.apply_changes! if z.modified? }
```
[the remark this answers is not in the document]
teclator: Probably better for this line of code, I just modified from the master version, which will be two lines one creating zones if new and then applying changes, so will update this for GA but will change again for master :)

library/network/src/lib/y2firewall/firewalld.rb
```diff
@@ -211,9 +218,30 @@ def start
...
+    # @return [ZoneReader]
+    def zone_reader
+      ZoneReader.new(api.zones, api.list_all_zones(verbose: true))
+    end
+
+    # Convenience method to instantiate a services reader
+    #
+    # @return [ServiceReader]
+    def service_reader
+      ServiceReader.new
```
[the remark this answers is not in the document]
teclator: I was thinking on adding different readers and that is why I moved to a convenience method but for services does not make too much sense so... :+1:

library/network/test/y2firewall/relations_test.rb
teclator: Although it tests the definition of the method, I also wanted to add an expect for the behavior expecting the defined method to work as expected.

It might be a different test for that of course.

library/network/test/y2firewall/firewalld/service_reader_test.rb
teclator: I will add a comment, it was a workaround added for the definition of $CHILD_STATUS, in other case it will be nil at the moment of allowing calls over it.

library/network/src/lib/y2firewall/firewalld/zone_reader.rb
teclator: Originally I wanted `_value` to be always nil and preferred that split but it doesn't matter at the end, so using an space for splitting the entry is also fine.

library/network/test/y2firewall/relations_test.rb
imobachgs: But you are not testing `#description`. Yes, I prefer a separate test (or reflect the content of the test in the description somehow).

teclator:
The big coverage negative impact is because I removed the tests of deprecated code for avoiding some Polkit popup because of not mocking object.

The code will be moved in master: see https://github.com/yast/yast-yast2/pull/806

imobachgs:
Yes, I am fine with some decrease. But this code does not look like well covered according to coveralls: https://coveralls.io/builds/18930261/source?filename=library%2Fnetwork%2Fsrc%2Flib%2Fy2firewall%2Ffirewalld%2Fapi.rb#L138.

library/network/test/y2firewall/relations_test.rb
teclator: No, I'm not testing description but apply_attributes_changes!

description is an attribute, and if you change apply_attributes_changes should call the api for reflecting the changes. Like when you check that after a method call some instance variable has an specific value, it usually is an expectation

teclator:
@imobachgs I decided to bring back the old tests except the one for SuSEFirewallProposal that can be completely removed. The rest of deprecated code will be moved only in master as well as them tests.

Also I have increased the api test coverage.

library/network/test/y2firewall/firewalld/api/services_test.rb
```diff
@@ -95,12 +95,78 @@
...
+      allow(api).to receive(:string_command)
+        .with("--service=test", "--get-ports", permanent: api.permanent?)
+        .and_return("80/tcp 443/tcp")
+
+      expect(api.service_ports("test")).to eql(["80/tcp", "443/tcp"])
+    end
+  end
+
+  describe "#service_protocols" do
+    it "returns the list of protocols opened by the given service" do
```
imobachgs: np: I would say "used" instead of "opened".

library/network/test/y2firewall/firewalld/api/zones_test.rb
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+#
+# Copyright (c) [2017] SUSE LLC
```
imobachgs: Please, adjust the year :)

library/network/test/y2firewall/firewalld/api/zones_test.rb
```diff
@@ -0,0 +1,213 @@
...
+
+    it "returns true if the port is allowed by the zone" do
+      allow(api).to receive(:query_command)
+        .with("--zone=public", "--query-port=22/tcp", permanent: api.permanent?)
+        .and_return(true)
+      expect(subject.port_enabled?("public", "22/tcp")).to eql(true)
+    end
+  end
+
+  describe "#protocol_enabled?" do
```
imobachgs: This is just a matter of taste, but using a before block could save some repetition:

```ruby
describe "#protocol_enabled?" do
  before do
    allow(api).to receive(:query_command)
      .with("--zone=public", "--query-protocol=igmp", permanent: api.permanent?)
      .and_return(allowed?)
  end

  context "when the protocol is not allowed by the zone" do
    let(:allowed?) { false }

    it "returns false" do
      expect(subject.protocol_enabled?("public", "igmp")).to eql(false)
    end
  end

  context "when the protocol is allowed by the zone" do
    let(:allowed?) { true }

    it "returns true" do
      expect(subject.protocol_enabled?("public", "igmp")).to eql(true)
    end
  end
end
```

Just a suggestion, but it is ok as it is right now.

teclator: OK, will move next time, it save the repetition but needs to add the contexts and lets per each, so added the explicit allowements :see_no_evil:

imobachgs: Yes, that's the point. In my opinion the intention is more clear. You set the "preconditions" outside of the test and the it block only contains the relevant part. But, I've said, it is a matter of taste.

teclator:
Yep, I will change it in next iteration :+1:

PR #821, opened 2018-09-12
Bring back susefirewall proposal

teclator:
It was dropped in #818, we will wait for https://github.com/yast/yast-network/pull/668 and should be dropped in master only.

### DIFF
--- a/library/network/src/lib/network/susefirewalld.rb
+++ b/library/network/src/lib/network/susefirewalld.rb
@@ -1017,6 +1017,12 @@ module Yast
       true
     end
 
+    # FIXME: this method currently does nothing at all and has been added just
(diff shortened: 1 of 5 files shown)


library/network/src/lib/network/susefirewalld.rb
```diff
@@ -1017,6 +1017,12 @@ def SetAcceptExpertRules(zone, expert_rules)
       true
     end
 
+    # FIXME: this method currently does nothing at all and has been added just
+    # for having the same API than SuSEFirewall2 but it is deprecated
```
dgdavid: What about to use `@deprecated` tag instead of FIXME?

PR #826, opened 2018-09-14
Improved self-update to use some packages as an add-on (bsc#1101016)

lslezak:
- See https://bugzilla.suse.com/show_bug.cgi?id=1101016

## Notes

The full functionality depends also on these other pull requests:
- https://github.com/yast/yast-packager/pull/374
- https://github.com/yast/yast-installation/pull/741

## Tested Manually

- Tested manually with a customized `skelcd-control-SLES` package (Knut, thanks! :wink: ), applied using a DUD with `self_update=` boot parameter pointing to a repository with the new skelcd.
- Also tested an AutoYaST installation
- The extra addon repository is removed at the end

self-update-role

### DIFF
--- a/library/control/src/modules/WorkflowManager.rb
+++ b/library/control/src/modules/WorkflowManager.rb
@@ -1494,10 +1494,13 @@ module Yast
       log.info "Merging #{product.label} workflow"
 
       if merged_base_product
-        Yast::WorkflowManager.RemoveWorkflow(:package, 0, merged_base_product.installation_package)
+        Yast::WorkflowManager.RemoveWorkflow(
+          :package,
+          merged_base_product.installation_package_repo,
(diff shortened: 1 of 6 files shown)


library/control/src/modules/WorkflowManager.rb
```diff
@@ -1604,12 +1608,18 @@ def package_repository(package_name)
       if pkgs.empty?
         log.warn("The installer extension package #{package_name} was not found")
         return nil
-      elsif pkgs.size > 1
+      end
+
```
teclator: NP: What about adding the bug number here as reference

lslezak: I think the code it self-explaining and does not need any bug reference. If needed you can find it in the Git history.

PR #828, opened 2018-09-17
Fix firewall-offline-cmd call for removing services from zones (bsc#1108628)

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1108628

With the offline cmd (firewall-offline-cmd), the `--remove-service` call can be used only with the default zone, in other case, it has to be called as `--remove-service-from-zone`

See: https://firewalld.org/documentation/man-pages/firewall-offline-cmd.html

Bug introduced when extending the API:

https://github.com/yast/yast-yast2/pull/819/files#diff-fade9352dfa13333ea0f66a045fe53eaL217

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld/api/zones.rb
+++ b/library/network/src/lib/y2firewall/firewalld/api/zones.rb
@@ -131,8 +131,7 @@ module Y2Firewall
         #   modifies the permanent configuration
         # @return [Boolean] True if the interface was removed from the zone
         def remove_interface(zone, interface, permanent: permanent?)
-          modify_command("--zone=#{zone}", "--remove-interface=#{interface}",
-            permanent: permanent)
+          modify_command("--zone=#{zone}", "--remove-interface=#{interface}", permanent: permanent)
         end
 
         # @param zone [String] The firewall zone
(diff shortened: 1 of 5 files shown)


library/network/src/lib/y2firewall/firewalld/api/zones.rb
```diff
@@ -131,8 +131,7 @@ def add_interface(zone, interface, permanent: permanent?)
         #   modifies the permanent configuration
         # @return [Boolean] True if the interface was removed from the zone
         def remove_interface(zone, interface, permanent: permanent?)
-          modify_command("--zone=#{zone}", "--remove-interface=#{interface}",
-            permanent: permanent)
+          modify_command("--zone=#{zone}", "--remove-interface=#{interface}", permanent: permanent)
```
imobachgs: What's changed here? I only see the `permanent` argument being moved to the same line, right?

library/network/src/lib/y2firewall/firewalld/api/zones.rb
```diff
@@ -234,7 +233,9 @@ def add_protocol(zone, protocol, permanent: permanent?)
         #   modifies the permanent configuration
         # @return [Boolean] True if service was removed from zone
         def remove_service(zone, service, permanent: permanent?)
-          modify_command("--zone=#{zone}", "--remove-service=#{service}", permanent: permanent)
+          remove_arg = offline? ? "--remove-service-from-zone" : "--remove-service"
+          modify_command("--zone=#{zone}", "#{remove_arg}=#{service}",
```
imobachgs: np: just for consistency, why not put the `permanent` argument in the same line?

imobachgs: To be honest, the CLI interface does not look consistent to me. But I may be missing something.

library/network/src/lib/y2firewall/firewalld/api/zones.rb
teclator: Right ;)

library/network/src/lib/y2firewall/firewalld/api/zones.rb
teclator: Already pushed

PR #831, opened 2018-09-18
Add CWM popup

imobachgs:
Extract some reusable bits from yast2-storage-ng.


library/cwm/src/lib/cwm/ui_state.rb
```diff
@@ -0,0 +1,209 @@
...
+  #
+  class UIState
+    include Yast::I18n
+
+    # Constructor
+    #
+    # Called through {.create_instance}, starts with a blank situation (which
+    # means default for each widget will be honored).
+    def initialize
+      textdomain(textdomain_name)
```
jreidinger: what is a goal of this call? I do not see any translatable string.

library/cwm/src/lib/cwm/ui_state.rb
```diff
@@ -0,0 +1,209 @@
...
+    end
+
+    # Method to be called when the user operates in a row of a table.
+    #
+    # @param row_id [Object] row identifier
+    def select_row(row_id)
+      self.row_id = row_id
+    end
+
+    # Select the page to open in the general tree after a redraw
```
jreidinger: It is strange to mention select when in fact you do find.

library/cwm/src/lib/cwm/ui_state.rb
```diff
@@ -0,0 +1,209 @@
...
+    # Whether the given page matches with the candidate tree node
+    #
+    # @param page [CWM::Page]
+    # @param candidate [Integer, String]
+    # @return boolean
+    def matches?(page, candidate)
+      page.label == candidate
+    end
+
+    class << self
```
jreidinger: why not use Singleton module?

library/cwm/test/Makefile.am
```diff
@@ -4,6 +4,7 @@ TESTS = \
 	custom_widget_test.rb \
 	dialog_test.rb \
 	pager_test.rb \
+	popup_test.rb \
```
jreidinger: ui state test missing.

library/cwm/src/lib/cwm/ui_state.rb
imobachgs: To be able to create a new instance during tests.

library/cwm/src/lib/cwm/ui_state.rb
imobachgs: It was in the original implementation, so I've decided to keep it. I will double check if it is still needed.

library/cwm/src/lib/cwm/ui_state.rb
imobachgs: Yes, it sounds confusing. But I decided to keep the original API as much as possible.

library/cwm/src/lib/cwm/ui_state.rb
jreidinger: it is possible even with standard Singleton module:

just try this code snippet

```
jreidinger@linux-vvcf:~/prace/yast/storage-ng> cat /tmp/test.rb 
require "singleton"

class A
  include Singleton

  attr_accessor :text
end

A.instance.text = "test"

puts A.instance.text.inspect
Singleton.__init__(A)
puts A.instance.text.inspect

```

library/cwm/src/lib/cwm/ui_state.rb
jreidinger: My problem with it is that it just adds requirements ( defining textdomain name) without any benefits visible for me.

library/cwm/src/lib/cwm/ui_state.rb
```diff
@@ -0,0 +1,209 @@
...
+  #       [:notify, :immediate]
+  #     end
+  #
+  #     def handle(event)
+  #       MyUIState.instance.select_row(value) if event["EventReason"] == "SelectionChanged"
+  #       nil
+  #     end
+  #   end
+  #
+  class UIState
```
jreidinger: well, I see it quite late in y2-storage. But from API and API usability POV, is possible to have it as simple mixin for CWM Widgets ( or CWM dialog ), that register itself to watch for changes and then in next init modify initial state?

So from usability POV something like
```
class MyDialog < CWM::Dialog
  include UIState
end
```
and it will works automatically with registering for UI events and query for CWM object states? It will be more robust and less error prone from my POV as now it really depends on not forgetting to call special methods when some method appear and also widgets usually need to be aware of it.

library/cwm/src/lib/cwm/ui_state.rb
imobachgs: To be honest, calling to a method called `__init__` looks wrong to me. I prefer to just add a few lines of code than relying on a method which is not expected to be used outside. It is a common pattern in yast2-storage-ng.

library/cwm/src/lib/cwm/ui_state.rb
imobachgs: UIState does not have support for remembering the widget values. It is limited to store navigation information (open page, tab and selected row). On the other and, I can see your point, so I will think about an alternate solution.

imobachgs:
Regarding the `UIState` module, I would consider this PR as a first step in that direction but, as it is for `SLE-15-GA`, I would not implement the mixin mechanism to include those methods in the widgets. I would rather prefer postpone the discussion. What about creating a PBI in order to come up with a future proof and better solution?

jreidinger:
@imobachgs ok, I just worry it ends as other idea PBIs in collumn never.

imobachgs:
@jreidinger That's a fair point. But I've been discussing these days (before doing this changes) with @ancorgs and @dgdavid about managing state, so we will try to come up with something.

PR #834, opened 2018-09-19
Return zone object

imobachgs:
* `Y2Firewall::Firewalld::Interface#zone` returns an object.
* Add unit tests to the class.

### DIFF
--- a/library/network/src/modules/CWMFirewallInterfaces.rb
+++ b/library/network/src/modules/CWMFirewallInterfaces.rb
@@ -161,7 +161,11 @@ module Yast
     # @return a list of interfaces that will be opened
     def Selected2Opened(ifaces, _nm_ifaces_have_to_be_selected)
       log.info("Selected ifaces: #{ifaces}")
-      zone_names = ifaces.map { |i| interface_zone(i) || default_zone.name }.uniq
+      zone_names = ifaces.map do |name|
+        zone = interface_zone(name)
+        zone ? zone.name : default_zone.name
+      end
+      zone_names.uniq!
       log.info("Ifaces zone names: #{zone_names}")
 
       zone_ifaces =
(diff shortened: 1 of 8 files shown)


teclator:
It looks good, but CWMFirewallInterfaces probably will need to be adapted too

teclator:
Something like:

```bash
--- a/library/network/src/modules/CWMFirewallInterfaces.rb
+++ b/library/network/src/modules/CWMFirewallInterfaces.rb
@@ -251,7 +251,7 @@ module Yast
       zones =
         known_interfaces.each_with_object([]) do |known_interface, a|
           if allowed_interfaces.include?(known_interface.name)
-            zone_name = known_interface.zone || firewalld.default_zone
+            zone_name = known_interface.zone ? known_interface.zone.name : firewalld.default_zone
             a << zone_name
           end
         end
diff --git a/library/network/test/test_helper.rb b/library/network/test/test_helper.rb
index a62c678c..e52b4e8b 100644
--- a/library/network/test/test_helper.rb
+++ b/library/network/test/test_helper.rb
@@ -82,7 +82,8 @@ module NetworkStubs
   }.freeze
 end
 
-def mock_firewalld_interface(id, name, zone)
+def mock_firewalld_interface(id, name, zone_name)
+  zone = instance_double("Y2Firewall::Firewalld::Zone", name: zone_name)
   instance_double("Y2Firewall::Firewalld::Interface",
     id: id, name: id.to_s, device_name: name, zone: zone)
 end
```

library/network/src/modules/CWMFirewallInterfaces.rb
```diff
@@ -161,7 +161,11 @@ def UpdateFirewallStatus
     # @return a list of interfaces that will be opened
     def Selected2Opened(ifaces, _nm_ifaces_have_to_be_selected)
       log.info("Selected ifaces: #{ifaces}")
-      zone_names = ifaces.map { |i| interface_zone(i) || default_zone.name }.uniq
```
teclator: I was also fine with

```zone_names = ifaces.map { |i| (interface_zone(i) || default_zone).name }.uniq```

PR #836, opened 2018-09-24
Moved firewall selection to a specific class

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1093052
- Trello Card: https://trello.com/c/VyxzXmLR


library/network/src/lib/network/firewall_chooser.rb
```diff
@@ -0,0 +1,140 @@
...
+      if running_backends.empty? && enabled_backends.size > 1
+        Builtins.y2warning("Both SuSEfirewall2 and firewalld services are enabled. " \
+                            "Defaulting to SuSEfirewall2")
+        enabled_backends[0] = :sf2
+      end
+
+      # Set a good default. The running one takes precedence over the enabled one.
+      selected_backend = running_backends[0] ? running_backends[0] : enabled_backends[0]
+      # Fallback to the first installed backend or to SuSEFirewall2 if not
+      selected_backend = (installed_backends.first || :sw2) if selected_backend.to_s.empty?
```
teclator: Main change

Use the first installed_backed or :sw2

library/network/src/modules/SuSEFirewall.rb
```diff
@@ -29,131 +29,10 @@
...
-          # emit a warning
-          if running_backends.empty? && enabled_backends.size > 1
-            Builtins.y2warning("Both SuSEfirewall2 and firewalld services are enabled. " \
-                               "Defaulting to SuSEfirewall2")
-            enabled_backends[0] = :sf2
-          end
-
-          # Set a good default. The running one takes precedence over the enabled one.
-          selected_backend = running_backends[0] ? running_backends[0] : enabled_backends[0]
-          selected_backend = :sf2 if selected_backend.to_s.empty? # SF2 still the default
```
teclator: When there is no firewall running it fallback to SuSEfirewall2 even when it is not installed

library/network/src/lib/network/firewall_chooser.rb
```diff
@@ -0,0 +1,137 @@
...
+    #
+    # @raise [SuSEFirewallMultipleBackends] if firewalld and SuSEfirewalld2 are
+    #   running
+    # @return [Symbol] the backend that should be used
+    def self.detect
+      # Old testsuite
+      return :sf2 if Mode.testsuite
+
+      # Only one running backend is permitted.
+      raise SuSEFirewallMultipleBackends if running_backends.size > 1
```
imobachgs: It is documented. I was just wondering whether we should take this case into account. However, it won't hurt.

library/network/src/lib/network/firewall_chooser.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      # If both firewalls are enabled, then make SF2 the default one and
+      # emit a warning
+      if running_backends.empty? && enabled_backends.size > 1
+        Builtins.y2warning("Both SuSEfirewall2 and firewalld services are enabled. " \
+                            "Defaulting to SuSEfirewall2")
+        enabled_backends[0] = :sf2
```
imobachgs: What about using a different variable instead of modifying this array (does it work at all)?

library/network/src/lib/network/firewall_chooser.rb
```diff
@@ -0,0 +1,137 @@
...
+      end
+
+      # Set a good default. The running one takes precedence over the enabled one.
+      selected_backend = running_backends[0] ? running_backends[0] : enabled_backends[0]
```
imobachgs: `running_backends.empty?` does not work in this case? It looks more readable.

library/network/src/lib/network/firewall_chooser.rb
```diff
@@ -0,0 +1,137 @@
...
+    DEFAULT_MESSAGE = "Multiple firewall backends are running. " \
+      "One needs to be shutdown to continue.".freeze
+    def initialize(message = DEFAULT_MESSAGE)
+      super message
+    end
+  end
+
+  # Factory for construction of appropriate firewall object based on
+  # desired backend.
+  class FirewallChooser
```
imobachgs: np: I am not a big fan of having a class which only has class methods.

library/network/src/lib/network/firewall_chooser.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+
+    # Determine which firewall should be selected as the backend depending on
+    # which one is enabled, running and/or installed. SuSEfirewall2 is the
+    # predefined one in case there is no way to decide.
+    #
+    # @raise [SuSEFirewallMultipleBackends] if firewalld and SuSEfirewalld2 are
+    #   running
+    # @return [Symbol] the backend that should be used
+    def self.detect
```
imobachgs: np: `detect` says nothing at all to me. What about `find_preferred`, or another name? Sure you can come up with a much better alternative :)

teclator:
@imobachgs thnks for the feedback

library/network/src/lib/network/firewall_chooser.rb
```diff
@@ -0,0 +1,142 @@
...
+require "network/susefirewalld"
+require "network/susefirewall2"
+
+Yast.import "Service"
+Yast.import "PackageSystem"
+Yast.import "Mode"
+
+module Yast
+  class SuSEFirewallMultipleBackends < StandardError
+    DEFAULT_MESSAGE = "Multiple firewall backends are running. " \
```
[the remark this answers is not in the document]
teclator: We usually do not translated them, so I would omit it.

PR #837, opened 2018-09-26
Do not modify the value if it is the same.

teclator:
A simple fix for http://bugzilla.suse.com/show_bug.cgi?id=1109812

The bug appeared because this change:

https://github.com/yast/yast-yast2/pull/818/files#diff-d15e2fb4c53a9dc67f63ddf07c2b6aaeL148

Now we do not check current value when writing, with this change we should not mark it as modified if the value is the same.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Sep 28 08:05:15 UTC 2018 - knut.anderssen@suse.com
+
+- Y2Firewall::Firewalld: Single attributes setter will not modify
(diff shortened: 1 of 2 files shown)


library/network/src/lib/y2firewall/firewalld/relations.rb
```diff
@@ -183,6 +184,7 @@ def has_many(*relations, scope: nil, cache: false) # rubocop:disable Style/Predi
           attr_reader relation
 
           define_method "#{relation}=" do |item|
+            return item if public_send(relation) == item
```
teclator: It could be improved, doing a sort of the items but will need other checks, so not added by now.

PR #838, opened 2018-09-27
Fixed internal error

mchf:
caused by a cleanup here https://github.com/yast/yast-yast2/pull/789/commits/27b496df69b15af1f4a3002b4837ec325dd8e249

### DIFF
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -1354,10 +1354,8 @@ module Yast
       end
       Builtins.y2debug("ChangeDevice(%1)", name)
 
-      devmap = deep_copy(newdev)
-
-      Ops.set(devsmap, name, devmap)
-      Ops.set(@Devices, t, devsmap)
+      # newdev is already a deep_copy created above
+      @Devices[t] = @Devices.fetch(t, {}).merge(name => newdev)
 
       Builtins.y2debug("Devices=%1", @Devices)
       true
--- a/library/network/test/network_interfaces_test.rb
+++ b/library/network/test/network_interfaces_test.rb
@@ -280,4 +280,9 @@ describe Yast::NetworkInterfaces do
     end
   end
 
+  describe "Change2" do
+    it "passes smoke test" do
+      expect { subject.Change2("eth0", {}, false) }.not_to raise_error
+    end
+  end
 end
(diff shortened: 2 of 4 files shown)


library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1354,10 +1354,8 @@ def Change2(name, newdev, check)
       end
       Builtins.y2debug("ChangeDevice(%1)", name)
 
-      devmap = deep_copy(newdev)
-
-      Ops.set(devsmap, name, devmap)
-      Ops.set(@Devices, t, devsmap)
+      # newdev is already a deep_copy created above
+      @Devices.fetch(t, {})[name] = newdev
```
jreidinger: this looks wrong for me. As if `@devices` is `nil`, then after this call it will be still `nil` and new newdev is not asigned anywhere.

jreidinger: so what I propose is simple
```
@Devices[t] ||= {}
@Devices[t][name] = newdev
```

jreidinger: just try it yourself:

```
jreidinger@pepa:~/yast/storage-ng> irb
irb(main):001:0> a = {}
=> {}
irb(main):002:0> a.fetch(:a, {})[:b] = "test"
=> "test"
irb(main):003:0> a
=> {}
```

jreidinger: and this reminds me, that I would really like test for it, that test exactly both cases so another internal error is prevented.

mchf: ```@Devices``` is never nil - it is global assumption in the module, see also https://github.com/yast/yast-yast2/blob/master/library/network/src/modules/NetworkInterfaces.rb#L83

teclator: What about using `add_device(name, newdev)` ?, or I'm missing something

teclator: Although maybe we could modify the current method to accept a type if we already know it:

```ruby
    def add_device(device, ifcfg, type = nil ) # type could be a named parameter also
      # if possible use dev type as available in /sys otherwise use ifcfg config
      # as a fallback for device type detection
      devtype = type || GetTypeFromIfcfgOrName(device, ifcfg)
      @Devices[devtype] ||= {}
      @Devices[devtype][device] = ifcfg
end
```

PR #841, opened 2018-10-02
Make the service status label stretchable (bsc#1110407)

lslezak:
- When changing the label from "Active" to "Inactive" the longer label was not displayed correctly in the ncurses UI.
- Use `HSpacing(1)` instead of `Label(" ")` - if there is lack of space libyui first decreases/removes the spacings, this might be enough without cutting off the labels
- openQA: https://openqa.opensuse.org/tests/765329#step/yast2_dns_server/35
- bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1110407
- 4.1.20

### DIFF
--- a/library/systemd/src/lib/yast2/service_widget.rb
+++ b/library/systemd/src/lib/yast2/service_widget.rb
@@ -82,8 +82,8 @@ module Yast2
           Left(
             HBox(
               Label(_("Current status:")),
-              Label(" "),
-              Label(Id(:service_widget_status), status)
+              HSpacing(1),
+              Label(Id(:service_widget_status), Opt(:hstretch), status)
             )
           ),
           Left(action_widget),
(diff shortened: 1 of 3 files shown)


library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -82,8 +82,8 @@ def content
           Left(
             HBox(
               Label(_("Current status:")),
-              Label(" "),
-              Label(Id(:service_widget_status), status)
+              HSpacing(1),
+              Label(Id(:service_widget_status), Opt(:hstretch), status)
```
jreidinger: what is exactly problem? In past I see similar issue with CWM and correct solution should be usage of UI.RecalcLayout https://yast-ui-bindings.surge.sh/classYCP__UI.html#a7da317092ab6d90ca7bbfe06c196b54d

lslezak: You need to add a hook to the change event, moreover it recalculates the *whole* layout which is not trivial. This only affects the label.

PR #845, opened 2018-10-09
Sle 15 ga bsc 1108942

schubi2:


library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -101,11 +102,86 @@ def remove_port(port_or_range, protocol, interface)
...
+      zone.name
+    end
+
+    # Check if the service belongs to the zone
+    #
+    # @param service [String] service name
+    # @param zone [String] zone name
+    #
+    # @return [Boolean] true if service is in zone
+    def is_service_in_zone(service, zone_name)
```
teclator: same than above

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -101,11 +102,86 @@ def remove_port(port_or_range, protocol, interface)
...
+    end
+
+    # Return an array with all the known (sysconfig configured) firewalld
+    # interfaces.
+    #
+    # @return [Array<Hash>] known interfaces
+    #        e.g. [{ "id":"eth0", "name":"Askey 815C", "zone":"EXT"} , ... ]
+    def all_known_interfaces
+      ret = []
+      Y2Firewall::Firewalld::Interface.known.each do |interface|
```
teclator: So, it is just a map
```ruby
def all_known_interfaces
  Y2Firewall::Firewalld::Interface.known.map do |interface|
    { "id" => ... }
  end
end
```

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -101,11 +102,86 @@ def remove_port(port_or_range, protocol, interface)
...
+      end
+      ret
+    end
+
+    # sets status for several services on several network interfaces.
+    #
+    # @param	list <string> service ids
+    # @param	list <string> network interfaces
+    # @param	boolean new status of services
+    def set_services(services, interfaces, status)
```
teclator: mmm not sure about this method name, I would maybe suggest something like `modify_interface_services`

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -101,11 +102,86 @@ def remove_port(port_or_range, protocol, interface)
       zone.remove_port(port)
     end
 
+    # Check whether the firewalld service is enable or not
+    #
+    # @return [Boolean] true if it is enable; false otherwise
+    def is_enabled
```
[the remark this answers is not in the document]
schubi2: Perl does not like it. Have had before....

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -101,11 +102,86 @@ def remove_port(port_or_range, protocol, interface)
...
+      firewalld.enabled?
+    end
+
+    # Return true if the logging config or any of the zones where modified
+    # since read
+    #
+    # @return [Boolean] true if the config was modified; false otherwise
+    def is_modified
```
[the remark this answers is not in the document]
schubi2: same than above :-)

library/network/src/modules/firewalld_wrapper.rb
schubi2: same than above :-)

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -101,11 +102,84 @@ def remove_port(port_or_range, protocol, interface)
...
+      firewalld.modified?
+    end
+
+    # Evaluate the zone name of an interface
+    #
+    # @param interface [String] interface name
+    #
+    # @return [String] zone name (nil; not found)
+    def zone_name_of_interface(interface)
```
teclator: np: zone_name_of(interface) is good enough

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -101,11 +102,84 @@ def remove_port(port_or_range, protocol, interface)
...
+    end
+
+    # Return an array with all the known (sysconfig configured) firewalld
+    # interfaces.
+    #
+    # @return [Array<Hash>] known interfaces
+    #        e.g. [{ "id":"eth0", "name":"Askey 815C", "zone":"EXT"} , ... ]
+    def all_known_interfaces
+      Y2Firewall::Firewalld::Interface.known.map do |interface|
+        { "id" => interface.name, "zone" => interface.zone.name,
```
teclator: If the interface zone is `nil` this method will crash, so I suggest to use `zone_name_of_interface(interface)`

library/network/src/modules/firewalld_wrapper.rb
teclator: It's fine as it is, but it somehow sounds redundant as the parameter is the interface itself.

library/network/src/modules/firewalld_wrapper.rb
```diff
@@ -101,11 +102,84 @@ def remove_port(port_or_range, protocol, interface)
...
+    end
+
+    # Return an array with all the known (sysconfig configured) firewalld
+    # interfaces.
+    #
+    # @return [Array<Hash>] known interfaces
+    #        e.g. [{ "id":"eth0", "name":"Askey 815C", "zone":"EXT"} , ... ]
+    def all_known_interfaces
+      Y2Firewall::Firewalld::Interface.known.map do |interface|
+        { "id" => interface.name, "zone" => zone_name_of_interface(interface.name),
```
teclator: For me it is fine as it is, I just would check the old behavior in case it was something like `zone_name || firewalld.default_zone`

Just to be sure we are not breaking anything

PR #846, opened 2018-10-09
Faster CWM by removing deep_copy

mvidner:
- https://trello.com/c/V9ebL3I2

So far covering with tests to make sure the fixes don't break it. Converted a legacy YCP test.

In the end this is not really significant, as better profiling has revealed the culprits to be elsewhere, but in the meantime I have added test coverage.


library/cwm/test/cwm_test.rb
```diff
@@ -0,0 +1,229 @@
...
+    end
+  end
+
+  describe "#PrepareDialog" do
+    it "returns early if the term is empty" do
+      expect(subject).to_not receive(:ProcessTerm)
+      expect(subject.PrepareDialog(VBox(), test_widgets)).to eq(VBox())
+    end
+
+    it "prepares args for ProcessTerm" do
```
jreidinger: please test here that it does not modify stringterm parameter, which is quite important thing when we reduce deep_copies

library/cwm/test/cwm_test.rb
```diff
@@ -0,0 +1,229 @@
...
+    end
+
+    # used by other CWM modules+classes
+    it "creates the '_cwm_key' keys" do
+      expect(cw[0]["_cwm_key"]).to eq("w1")
+      expect(cw[1]["_cwm_key"]).to eq("w2")
+    end
+  end
+
+  # used by CWMTab
```
jreidinger: please check, maybe CWMTab is already dead? Originally it was created for bootloader, but there we use object version.

mvidner: Not dead. 16 matches for https://github.com/search?q=org%3Ayast+CWMTab.CreateWidget&type=Code

library/cwm/src/lib/cwm/replace_point.rb
```diff
@@ -36,6 +36,7 @@ def init
     # handled.
     # @param widget [CWM::AbstractWidget] widget to display and process events
     def replace(widget)
+      Yast::Builtins.y2milestone("REPLACE BEG %1", t1 = Time.now.to_f)
```
jreidinger: btw `Yast.y2milestone` also works ;)

library/cwm/src/modules/CWM.rb
```diff
@@ -726,7 +725,9 @@ def CreateWidgets(names, source)
         deep_copy(m)
       end
       ret = Builtins.maplist(ret) { |w| prepareWidget(w) }
-      deep_copy(ret)
+      r = deep_copy(ret)
```
jreidinger: I think this deepcopy is not needed, as above is called maplist which itself call deepcopy - https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/builtins.rb#L121

library/cwm/src/modules/CWM.rb
```diff
@@ -742,7 +742,6 @@ def PrepareDialog(dialog, widgets)
         [widget_key, w]
       end.to_h
       t = ProcessTerm(dialog, m)
-      t = deep_copy(t)
```
jreidinger: please document it in method documentation that it is not deep copy, so can modify part of passed dialog.

library/cwm/test/cwm_test.rb
mvidner: This is surprisingly hard to test in general, because Ruby does not have `deep_clone` or `deep_freeze`. Fortunately `deep_copy` helps :-)

library/cwm/src/modules/CWM.rb
```diff
@@ -1120,6 +1101,13 @@ def widgets_contents(contents)
 
       res
     end
+
+  private
+
+    # such an accessor enables testing
+    def processed_widget=(w)
```
jreidinger: this looks like `attr_writer` for me

PR #851, opened 2018-10-16
Check if a service is supported from the list of current ones

teclator:
Since this [PR](https://github.com/yast/yast-yast2/pull/818) was merged, the `firewalld` instance already knows about the supported list of services, so, we can avoid the extra `API` calls made by the widget improving the user experience.

Example from https://github.com/SUSE/yast2-rmt/pull/34

## Before the fix
test

## After the fix

test_2

With more services it is probably easiest to appreciate.

### DIFF
--- a/library/network/src/modules/CWMFirewallInterfaces.rb
+++ b/library/network/src/modules/CWMFirewallInterfaces.rb
@@ -211,8 +211,6 @@ module Yast
     def InitAllowedInterfaces(services)
       service_status = {}
 
-      services.each { |s| firewalld.find_service(s) }
-
       zone_services(services).each do |_s, status|
         status.each do |iface, en|
           service_status[iface] = service_status.fetch(iface, true) && en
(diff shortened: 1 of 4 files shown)


library/network/src/modules/CWMFirewallInterfaces.rb
```diff
@@ -211,8 +211,6 @@ def DisplayFirewallDetailsPopupHandler(widget)
     def InitAllowedInterfaces(services)
       service_status = {}
 
-      services.each { |s| firewalld.find_service(s) }
```
teclator: This calls reads the service definition what is not needed at all.

library/network/src/modules/CWMFirewallInterfaces.rb
```diff
@@ -833,7 +831,7 @@ def CreateOpenFirewallWidget(settings)
         return { "widget" => :custom, "custom_widget" => not_installed_widget, "help" => "" }
       end
 
-      if services.any? { |s| !firewalld.api.service_supported?(s) }
+      if services.any? { |s| !firewalld.current_service_names.include?(s) }
```
teclator: we already know the list of available service, so we avoid to call the api as it is slow.

PR #858, opened 2018-10-24
Ensure installation source dir

jreidinger:


library/control/test/installation_test.rb
```diff
@@ -0,0 +1,28 @@
...
+
+describe Yast::Installation do
+  subject { Yast::Installation }
+
+  describe ".sourcedir" do
+    before do
+      allow(::File).to receive(:exist?).and_return(true)
+    end
+
+    it "returns string" do
```
lslezak: NP: for `returns string` test I'd expect something like `expect(...).to be_a(String)`.

jreidinger: well, I want be more strict in test and less in description. So if we in future decide to change, we need just adapt text and not its test.

PR #859, opened 2018-10-24
[RFC] Rubocop 0.59.2

lslezak:
# RFC - Rubocop Update

- This is just a testing PR to see the changes in the latest Rubocop version
- See also https://github.com/yast/yast-devtools/pull/134 for the default config update

## Discussion

- Please check the comments below about the new checks
- Check also the diff, if you do not like some change just comment it here
- This discussion should be mainly about the new global defaults, as usually we can fine tune the config per package

## Details

- Check the [full Rubocop output](https://gist.githubusercontent.com/lslezak/06fe223e022e273dbb4d98ad9f118912/raw/feed7acb6f53da375ca54892e3a2720931473b7f/rubocop.out)

<details>
<summary>Here is a summary (statistics) of the new reported issues</summary>
<pre>
1114  Layout/Tab
507   Style/EmptyLambdaParameter
300   Metrics/BlockLength
260   Style/Encoding
235   Layout/EmptyLineAfterGuardClause
209   Style/IfUnlessModifier
117   Style/NumericPredicate
71    Naming/UncommunicativeMethodParamName
62    Layout/MultilineMethodCallIndentation
41    Style/SymbolArray
34    Layout/EmptyLinesAroundArguments
31    Naming/VariableNumber
29    Style/FormatStringToken
27    Style/YodaCondition
22    Style/MultipleComparison
21    Style/EmptyMethod
20    Layout/EmptyLineAfterMagicComment
16    Lint/Syntax
14    Style/RedundantReturn
12    Layout/AccessModifierIndentation
12    Style/AccessModifierDeclarations
11    Lint/AmbiguousBlockAssociation
11    Style/ExpandPathArguments
11    Style/FormatString
7     Layout/ClosingParenthesisIndentation
7     Layout/IndentHeredoc
7     Lint/ScriptPermission
6     Style/CommentedKeyword
6     Style/InverseMethods
6     Style/MultilineIfModifier
5     Layout/MultilineOperationIndentation
5     Style/MixinUsage
5     Style/PercentLiteralDelimiters
4     Style/Dir
4     Style/RescueStandardError
4     Style/TernaryParentheses
3     Performance/Caller
3     Style/ConditionalAssignment
3     Style/NonNilCheck
3     Style/RedundantConditional
2     Layout/AlignParameters
2     Lint/DuplicateMethods
2     Naming/UncommunicativeBlockParamName
2     Style/IdenticalConditionalBranches
2     Style/MissingRespondToMissing
2     Style/UnneededCondition
1     Layout/EmptyComment
1     Layout/EmptyLinesAroundExceptionHandlingKeywords
1     Layout/FirstParameterIndentation
1     Layout/LeadingBlankLines
1     Layout/RescueEnsureAlignment
1     Lint/MissingCopEnableDirective
1     Lint/UselessAssignment
1     Naming/HeredocDelimiterNaming
1     Performance/InefficientHashSearch
1     Performance/UnneededSort
1     Security/YAMLLoad
1     Style/MethodMissingSuper
1     Style/StringLiterals
1     Style/WhileUntilModifier
--
3291  Total
</pre>
</details>

## Comments

Here are my comments about some new checks which need some discussion. The other new checks are IMHO fine and should be enabled.

### Style/NumericPredicate

```
library/cwm/src/modules/CWM.rb:119:30: C: Style/NumericPredicate: Use args.zero? instead of args == 0.
      return deep_copy(t) if args == 0
                             ^^^^^^^^^
```

I'd disable this cop, I do not see any readability improvement. Moreover `args.zero?` does not work with `nil` while `args == 0` does. This could cause crashes in some corner cases.

### Style/SymbolArray

```
library/cwm/src/modules/CWM.rb:813:20: C: Style/SymbolArray: Use %i or %I for an array of symbols.
      save_exits = [:next, :ok]
                   ^^^^^^^^^^^^
```
```diff
-      save_exits = [:next, :ok]
+      save_exits = %i[next ok]
```
I'd disable this check, I'm always confused what all that percent variants mean. (See the list at https://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Literals#The_%_Notation.)

### Metrics/BlockLength
```
library/cwm/test/abstract_widget_test.rb:8:1: C: Metrics/BlockLength: Block has too many lines. [147/25]
describe CWM::AbstractWidget do ...
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Whole rspec test is a single block so this will fail for almost all unit tests. On the other hand having large blocks in the real code is not good (just like having long methods). So I'd disable the check for unit tests (`*_test.rb` or `*_spec.rb` files) and increase the limit. For `Metrics/MethodLength` we use 30 so something similar would make sense also here.

### Style/EmptyMethod
```
library/cwm/test/replace_point_test.rb:40:5: C: Style/EmptyMethod: Put empty method definitions on a single line.
    def cleanup ...
    ^^^^^^^^^^^
```
```diff
-      def cleanup
-      end
+      def cleanup; end
```

It's a matter of your personal taste, but for me the current style looks a bit better. In the new style the name is a bit "hidden" in the middle of the line, so a bit less readable.

### Naming/UncommunicativeMethodParamName

This actually checks two issues, upper case letters and the length:

```
library/cwm/test/pager_test.rb:11:18: C: Naming/UncommunicativeMethodParamName: Method parameter must be at least 3 characters long.
  def initialize(n)
                 ^
```
```
library/general/src/modules/PackagesProposal.rb:86:21: C: Naming/UncommunicativeMethodParamName: Only use lowercase characters for method parameter.
    def CheckParams(unique_ID, type)
                    ^^^^^^^^^
```

IMHO it would make sense to enable globally, but as it might require a lot of renames in the old code we could optionally disable it for specific packages (or files).

### Layout/Tab

```
library/general/src/modules/Arch.rb:25:10: C: Layout/Tab: Tab detected.
# Module:       yast2
         ^
```

This might be a bit controversial, but I'd enable it. Tabs might be rendered differently in each editor and also GiHub has some issues with displaying diff with tabs (it looks wrong indented, sometimes you have to fetch the change and check it locally to see if it is really correct).

### Naming/VariableNumber

```
library/general/testsuite/tests/FileUtils.rb:54:7: C: Naming/VariableNumber: Use normalcase for variable numbers.
      @READ_EXISTS_1 = {
      ^^^^^^^^^^^^^^
```
The rule makes sense, but this is usually used in the old testsuite, so either globally disable for all `testsuite/**/*.rb` files or disable per package.

### Style/FormatString
```
library/general/src/modules/OSRelease.rb:62:47: C: Style/FormatString: Favor format over String#%.
          _("Release file %{file} not found") % { file: release_file }
```
```diff
-          _("Release file %{file} not found") % { file: release_file }
+          format(_("Release file %{file} not found"), file: release_file)
```

This check enforces single style for text formatting, it allows three options: `format`, `sprintf` and `percent`. See the [documentation](https://docs.rubocop.org/en/latest/cops_style/#styleformatstring)

```ruby
# format
puts format('%10s', 'hoge')
# sprintf
puts sprintf('%10s', 'hoge')
# percent
puts '%10s' % 'hoge'
```

I'm not sure which style should be preferred, we use the percent style for translations, some places use `format` when formatting a non-translatable string. So probably disable this.

### Style/FormatStringToken
```
library/network/src/modules/CWMFirewallInterfaces.rb:869:36: C: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over template tokens (like %{foo}).
          _("Firewall Settings for %{firewall}") % { firewall: Y2Firewall::Firewalld::SERVICE },
```

Using tokens like `%<firewall>s` in the translated messages would very likely result in broken translations, I guess that the `s` (type) would be missing quite often in the translations and cause this ugly issue:

```
"%<foo>s" % { foo: "bar" }
=> "bar"

"%<foo>" % { foo: "bar" }
=> "%"
```

The cop allows two more settings, but we sometimes we use "%{foo}", in some simple cases just "%s". I do not think we need to enforce a single style for this. 

Additionally this would probably break a lot of translations so I'd propose to disable this check.


jreidinger:
btw if you are interesting in the latest rubocop, I intentionally enable it for CFA. So I use there also config for it. see - https://github.com/config-files-api/config_files_api/blob/master/.rubocop.yml e.g. for that block lenght I have there exception for tests. So you can take it as inspiration.

lslezak:
Not relevant anymore, we have Rubocoop 0.71 now, PR #946

PR #861, opened 2018-10-25
spec: use noun phrase

jengelh:
Use an improved write style for package descriptions.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Tue Apr  9 08:03:00 UTC 2019 - Jan Engelhardt <jengelh@inai.de>
+
+- Use noun phrase in summary.
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jengelh:
review reminder?

jengelh:
updated

package/yast2.changes
```diff
@@ -279,6 +279,11 @@ Fri Oct 26 11:46:16 UTC 2018 - jreidinger@suse.com
   not exist or is empty (bsc#948729)
 - 4.1.32
 
+-------------------------------------------------------------------
+Thu Oct 25 09:37:50 UTC 2018 - Jan Engelhardt <jengelh@inai.de>
+
+- Use noun phrase in summary.
```
jreidinger: I am sorry, but this won't work. Changes have to strictly incrementing and no backward inserting is allowed as there are users which check top of changes for new things. So please change date and put it on top or remove this part and use again `osc vc`

jengelh:
updated, #2.

jreidinger:
@jengelh thanks for your patch

PR #862, opened 2018-10-26
Improve view anymsg

jreidinger:


library/system/src/lib/yast2/clients/view_anymsg.rb
```diff
@@ -0,0 +1,308 @@
...
+          "Many system components now log into systemd journal.\n" \
+          "Do you want to start YaST module for systemd journal?"
+        ),
+        buttons: :yes_no,
+        focus:   :no
+      ) == :yes
+
+      return false unless res
+
+      if !Package.Installed("yast2-journal")
```
jreidinger: for whole this part I expect that there have to be some existing method that do all that asking and checking availability, but do not find it. Any pointers?

PR #866, opened 2018-11-16
check return value of OpenDialog to prevent crash

aschnell:
For https://trello.com/c/zkLss9CU/460-tw-1115745-p3-yast2-software-management-could-not-load-plug-in-qt-pkg.

### DIFF
--- a/library/packages/src/modules/PackagesUI.rb
+++ b/library/packages/src/modules/PackagesUI.rb
@@ -312,7 +312,8 @@ module Yast
         widget_options
       )
 
-      raise "Opening package selector failed." if !UI.OpenDialog(
+      # exception text
+      raise _("Opening package selector failed.") if !UI.OpenDialog(
         Opt(:defaultsize),
         if !widget_options.empty?
           PackageSelector(Id(:packages), widget_options, "")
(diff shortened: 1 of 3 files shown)


library/packages/src/modules/PackagesUI.rb
```diff
@@ -312,7 +312,7 @@ def RunPackageSelector(options)
         widget_options
       )
 
-      UI.OpenDialog(
+      raise "Opening package selector failed." if !UI.OpenDialog(
```
lslezak: Should we translate the error message? We do so for the other errors so it would make sense also here IMHO.

aschnell: I could not find other exceptions being translated so I also did not translate this one. But surely I can add translations.

aschnell: BTW: Most of the messages in the generic exception popup itself are not translated.

jreidinger: in general we should not translate text in exceptions as generic exceptions catcher should be as trivial as possible as even I18n can be down or in inconsistent state. Where translation makes sense is place where you catch such specific exception and where you provide some nice message for user what he can do.

PR #867, opened 2018-11-20
WIP: Allow system role to default to no local user

shundhammer:
## Trello

https://trello.com/c/PAVZthmL

## Changes

Added global parameter _enable_local_users_

## Related PRs

https://github.com/yast/yast-installation/pull/764
https://github.com/yast/yast-installation-control/pull/73

### DIFF
--- a/library/control/src/modules/ProductFeatures.rb
+++ b/library/control/src/modules/ProductFeatures.rb
@@ -69,6 +69,7 @@ module Yast
           "vendor_url"                      => "",
           "enable_clone"                    => false,
           "disable_os_prober"               => false,
+          "enable_local_users"              => true,
           # FATE #304865
           "base_product_license_directory"  => "/etc/YaST2/licenses/base/",
           "full_system_media_name"          => "",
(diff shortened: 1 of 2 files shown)


library/control/src/modules/ProductFeatures.rb
```diff
@@ -241,6 +242,27 @@ def GetBooleanFeature(section, feature)
...
+    # @param [Boolean] fallback
+    #
+    # @return [Boolean] the feature value or fallback if not specified
+    def GetBooleanFeatureWithFallback(section, feature, fallback)
+      value = GetFeature(section, feature)
+      return fallback if value.nil?
+      return value if Ops.is_boolean?(value)
+
+      if value.respond_to?(:downcase)
+        return true  if ["yes", "true"].include?(value.downcase)
```
aschnell: Could this break in strange locales?

shundhammer: No, those are values from the XML file, and this is used in other places in the same code as well.

PR #870, opened 2018-12-05
Sle 15 ga 1113732

schubi2:


lslezak:
See the related https://github.com/yast/yast-services-manager/pull/183/

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -154,7 +154,7 @@ def build(name)
       #
       # @param names [Array<String>] service names to find
       #
-      # @return [Array<SystemService>]
+      # @return [Array<SystemService>] not found: SystemService.service is nil
```
lslezak:
```suggestion
      # @return [Array<SystemService>] when not found then the SystemService.service is nil for the respective service in the list
```

PR #872, opened 2018-12-06
Hardening

jreidinger:
- always use absolute path to binaries (bsc#1118291)
- properly escape shell arguments (bsc#1118291)


library/control/src/modules/ProductProfile.rb
```diff
@@ -99,7 +100,7 @@ def IsCompliant(productId)
           tmp_path = Ops.add(Ops.add(@profiles_dir, name), ".profile")
           SCR.Execute(
             path(".target.bash"),
-            Builtins.sformat("/bin/cp -a '%1' '%2'", profile, tmp_path)
+            Builtins.sformat("/bin/cp -a '%1' '%2'", profile.shellescape, tmp_path.shellescape)
```
shundhammer: Bad idea: This will fail. %1 and %2 are already quoted in that command!
Simply leave it as it is in this particular code location.

library/control/src/modules/InstError.rb
```diff
@@ -83,7 +76,7 @@ def SaveLogs
       cmd = Convert.to_map(
         WFM.Execute(
           path(".local.bash_output"),
-          Builtins.sformat("save_y2logs '%1'", String.Quote(savelogsto))
+          Builtins.sformat("/usr/sbin/save_y2logs '%1'", String.Quote(savelogsto))
```
shundhammer: Bad idea: %1 is already quoted. This will now be double-quoted which is most likely not what you want.

library/control/src/modules/ProductProfile.rb
jreidinger: you are right. I will just remove that quoting. Thanks for catching

library/control/src/modules/InstError.rb
```diff
@@ -45,26 +45,19 @@ def main
...
-          0,
-          "/"
-        )
-        homedir = "/" if homedir == ""
+      homedir = if env_home && !env_home.empty?
+        env_home
       else
         Builtins.y2warning(
           "Unable to find out home dir: %1, using %2",
           cmd,
```
mvidner: `cmd` has been deleted

library/control/src/modules/InstError.rb
```diff
@@ -45,26 +45,19 @@ def main
...
           homedir
         )
+        "/"
       end
-      homedir = Builtins.sformat("%1/y2logs.tgz", homedir)
+      homedir << "/y2logs.tgz"
```
mvidner:
```console
$ ruby -e 'h = ENV["HOME"]; h << "/foo"'
Traceback (most recent call last):
-e:1:in `<main>': can't modify frozen String (FrozenError)
```

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -302,15 +303,15 @@ def create(snapshot_type, description, previous: nil, cleanup: nil, important: f
...
+          root:          target_root.shellescape,
+          snapshot_type: snapshot_type.to_s.shellescape,
+          description:   description.shellescape)
+        cmd << " --pre-num #{previous.number.to_s.shellescape}" if previous
         cmd << " --userdata \"important=yes\"" if important
 
         if cleanup
           strategy = CLEANUP_STRATEGY[cleanup]
-          cmd << " --cleanup \"#{strategy}\"" if strategy
+          cmd << " --cleanup \"#{strategy.shellescape}\"" if strategy
```
shundhammer: Bad idea: That string is already quoted within this command line; that's the `\"` before and after the strategy. If you shell-escape it in addition to that, this will be wrong.

Since the strategies are only a very limited set of fixed strings, this will most likely never cause problems, but it is contrary to the intent of this. Please remove this shellescape call.

library/control/src/modules/InstError.rb
shundhammer: More precisely, the variable now longer exists before the `else` block when it is still referenced, so we will get `nil` there which may or may not result in a nil.NilClass exception (depending how smart our y2warning is).

library/control/src/modules/InstError.rb
```diff
@@ -45,26 +45,19 @@ def main
     end
 
     def SaveLogs
-      cmd = Convert.to_map(
-        WFM.Execute(path(".local.bash_output"), "echo ${HOME}")
-      )
-      homedir = "/"
+      env_home = ENV["HOME"]
```
shundhammer: +=  `dup`  (see mvidner's backtrace below9

library/cwm/src/modules/CWMTsigKeys.rb
```diff
@@ -397,8 +398,8 @@ def Handle(widget, _key, event)
             SCR.Execute(
               path(".target.bash"),
               Builtins.sformat(
-                "rm -rf `ls /etc/named.d/K%1\\.*`",
-                Builtins.tolower(key2)
+                "/usr/bin/rm -rf `ls /etc/named.d/K%1\\.*`",
```
shundhammer: Please also add the path to that embedded backtick call to `ls`.

library/control/src/modules/InstError.rb
jreidinger: well, cmd cause method not found exception. for homedir it will be nil, so no problem for y2warning.

JFYI how it can be verified in irb:
```
irb(main):001:0> test = if false; puts "no"; else puts test.inspect; end
nil
=> nil
```

library/gpg/src/modules/GPG.rb
```diff
@@ -240,22 +241,12 @@ def CreateKey
...
-                Ops.add(Ops.add("LC_ALL=en_US.UTF-8 ", xterm), " -e \""),
-                command
-              ),
-              "; echo $? > "
-            ),
-            exit_file
-          ),
-          "\""
-        )
+        command = "LC_ALL=en_US.UTF-8 #{xterm.shellescape} -e " \
```
shundhammer: That appears overkill to me. That `xterm` variable is really a constant that is set at the start of this function, then it checks if that thing is available (which could also be refactored to make this much better readable and maintainable). No different value is ever used or allowed here. So I'd vote for either simply using a string literal `/usr/bin/xterm` in both of the two (no more!) cases where it's used or at least use a constant `XTERM` instead. This can clearly never have a value that needs to be shell-escaped.

library/log/src/modules/LogViewCore.rb
```diff
@@ -120,19 +121,20 @@ def Start(widget, d)
...
+            file.shellescape,
+            grep.shellescape
           )
 
           if max_lines != 0
             lc_command = Builtins.sformat(
-              "cat '%1' | grep '%2' | wc -l",
-              file,
-              grep
+              "/usr/bin/cat %1 | /usr/bin/grep %2 | /usr/bin/wc -l",
```
shundhammer: Simply use `/usr/bin/grep -c` instad of the `grep | wc -l` pipeline?

shundhammer: ...even worse, that `cat` is completely redundant. Simply use

`/usr/bin/grep -c %2 <%1`

library/log/src/modules/LogViewCore.rb
```diff
@@ -153,10 +155,7 @@ def Start(widget, d)
               lines_count = 0 if Ops.less_than(lines_count, 0)
 
               if Ops.greater_than(lines_count, 0)
-                command = Ops.add(
-                  command,
-                  Builtins.sformat(" | tail -n +%1", lines_count)
-                )
+                command << Builtins.sformat(" | tail -n +%1", lines_count.to_s.shellescape)
```
shundhammer: Do we really need to shell-escape a number? and while we are at it, why not use Ruby variable expansion here and get rid of this `sformat()`?

library/log/src/modules/LogViewCore.rb
jreidinger: even stdin redirect is not needed and simple `/usr/bin/grep -c %2 %1` is enough

library/network/src/lib/network/firewalld.rb
```diff
@@ -140,31 +141,31 @@ def zones
     # @param zone [String] The firewall zone
     # @return [Array<String>] list of zone's interfaces
     def list_interfaces(zone)
-      fwd_result("--permanent --zone=#{zone} --list-interfaces").split(" ")
+      fwd_result("--permanent --zone=#{zone.shellescape} --list-interfaces").split(" ")
```
shundhammer: I know that this is beyond the scope of this PR, but still...

It appears to me that the result of this `fwd_result` function is _always_ sent straight away to `split()` to be split up into pieces. So why not change that function so it returns the array that everybody using it needs?

library/log/src/modules/LogViewCore.rb
jreidinger: @shundhammer it depends how paranoic you are and if you believe that lines_count is always only integer or if you think that attacker can somehow pass malicious string there.

library/network/src/modules/NetworkService.rb
```diff
@@ -93,18 +94,15 @@ def main
     # Helper to run systemctl actions
     # @return exit code
     def RunSystemCtl(service, action)
-      cmd = Builtins.sformat("%1 %2 %3.service", SYSTEMCTL, action, service)
-      ret = Convert.convert(
-        SCR.Execute(path(".target.bash_output"), cmd, "TERM" => "raw"),
-        from: "any",
-        to:   "map <string, any>"
-      )
+      cmd = Builtins.sformat("%1 %2 %3.service", SYSTEMCTL.shellescape,
```
shundhammer: Having a constant SYSTEMCTL in that class and then needing to shell-escape that here is IMHO very much counterproductive and makes the code less readable and maintainable, as we can see here: It does not need to be shell-escaped because it's a constant, and it is used for a grand total of **one** (1) time in the complete code.

Please change this to simply use

```Ruby
cmd = "/usr/bin/systemctl #{action.shellescape} #{service.shellescape}"

and get rid of the constant in the class header.
```

library/network/src/modules/NetworkService.rb
```diff
@@ -93,18 +94,15 @@ def main
...
+        action.shellescape, service.shellescape)
+      ret = SCR.Execute(path(".target.bash_output"), cmd, "TERM" => "raw")
       Builtins.y2debug("RunSystemCtl: Command '%1' returned '%2'", cmd, ret)
       Ops.get_integer(ret, "exit", -1)
     end
 
     def run_wicked(*params)
-      cmd = "#{WICKED} #{params.join(" ")}"
+      cmd = "#{WICKED} #{params.map(&:shellescape).join(" ")}"
```
shundhammer: Same as above: That const is used only once, and it obscurs what's going on here. I suggest to change this to a plain string as well:

```Ruby
cmd = "/usr/bin/wicked..."
```

library/network/src/modules/PortAliases.rb
```diff
@@ -140,29 +141,10 @@ def AllowedPortNameOrNumber
...
-      port_name
-    end
-
     # Internal function for loading unknown ports into memory and returning them as list[string]
     def LoadAndReturnPortToName(port_number)
-      command = Ops.add(
-        Ops.add("grep \"^[^#].*[ \\t]", port_number),
-        "/\" /etc/services | sed \"s/\\([^ \\t]*\\)[ \\t]*.*/\\1/\""
-      )
+      command = "grep \"^[^#].*[ \\t]#{port_number}/\" /etc/services " \
```
shundhammer: /usr/bin/grep and /usr/bin/sed

library/network/src/modules/PortAliases.rb
```diff
@@ -197,10 +179,9 @@ def LoadAndReturnNameToPort(port_name)
         return nil
       end
 
-      command = Ops.add(
-        Ops.add("grep --perl-regexp \"^", QuoteString(port_name)),
-        "[ \\t]\" /etc/services | sed \"s/[^ \\t]*[ \\t]*\\([^/ \\t]*\\).*/\\1/\""
-      )
+      grep_regexp = "^#{port_name}[ \\t]".shellescape
+      command = "grep --perl-regexp #{grep_regexp} /etc/services " \
```
shundhammer: /usr/bin/grep and /usr/bin/sed

library/packages/src/modules/PackageSystem.rb
```diff
@@ -66,7 +67,7 @@ def main
       Yast.include self, "packages/common.rb"
 
       @_rpm_query_binary_initialized = false
-      @_rpm_query_binary = "rpm"
+      @_rpm_query_binary = "/usr/bin/rpm"
```
shundhammer: Again, this is beyond the scope of this PR, but still...

It seems to be very inconsistent to have a member variable (why not a constant? does that ever change?) for the RPM command, yet not use that variable, but the string literal everywhere in the code. That makes me wonder what that member variable is good for at all.

library/system/src/lib/yast2/fs_snapshot.rb
mvidner: Explicit quotes with shellescape are plain wrong.

library/system/src/modules/Initrd.rb
```diff
@@ -400,9 +400,9 @@ def Write
         path(".target.bash"),
         Builtins.sformat(
           "/sbin/mkinitrd %1 %2 >> %3 2>&1",
-          param,
-          @additional_parameters,
-          Ops.add(Directory.logdir, "/y2logmkinitrd")
+          param, # cannot escaped due multiple possible params
```
shundhammer: English: "due **to**"  (also below)

shundhammer: Apart from that, I don't understant that comment at all. Examples might help.

library/system/src/modules/Initrd.rb
```diff
@@ -400,9 +400,9 @@ def Write
...
+          @additional_parameters, # cannot escaped due multiple possible params
+          File.join(Directory.logdir, "y2logmkinitrd")
```
shundhammer: shell-escape the result of this? Are we 100% sure `Directory.logdir` can never contain special characters like, say, a blank?

library/system/src/lib/yast2/fs_snapshot.rb
shundhammer: OK, removing the quotes is also an option; but one way or the other, it needs to be consistent. Right now we have the embedded quotes with `\"` **plus** the shell-escape. That won't work.

library/log/src/modules/LogViewCore.rb
shundhammer: You are right. So let's simplify this to get it better readable and maintainable.

library/log/src/modules/LogViewCore.rb
shundhammer: OK.

library/packages/src/modules/PackageSystem.rb
jreidinger: here it is problematic, as in code this path can change. 
see:
https://github.com/yast/yast-yast2/pull/872/files/7f355a06aa1746e0c11fb8315ca9bf2bd655f296#diff-2a6112a932a3433114e3b20bb80f0c42R323
but it looks quite old feature. So for now I am not sure if I want to drop that feature completely.

library/system/src/modules/Initrd.rb
jreidinger: well, I want to say that param can contant multiple parameters like "-s something" at https://github.com/yast/yast-yast2/pull/872/files/7f355a06aa1746e0c11fb8315ca9bf2bd655f296#diff-585c9223748698b81e5bb470365c90b8R397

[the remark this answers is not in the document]
jreidinger: yes, it is true.

library/system/src/modules/Initrd.rb
```diff
@@ -400,9 +400,9 @@ def Write
         path(".target.bash"),
         Builtins.sformat(
           "/sbin/mkinitrd %1 %2 >> %3 2>&1",
-          param,
-          @additional_parameters,
-          Ops.add(Directory.logdir, "/y2logmkinitrd")
+          param, # cannot escaped due multiple possible params
+          @additional_parameters, # cannot escaped due multiple possible params
```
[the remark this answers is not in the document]
jreidinger: I am not sure if I should as part of this pull request do also drop of unused code as it can grow a lot.

library/network/src/modules/PortAliases.rb
```diff
@@ -197,10 +179,9 @@ def LoadAndReturnNameToPort(port_name)
         return nil
       end
 
-      command = Ops.add(
-        Ops.add("grep --perl-regexp \"^", QuoteString(port_name)),
-        "[ \\t]\" /etc/services | sed \"s/[^ \\t]*[ \\t]*\\([^/ \\t]*\\).*/\\1/\""
-      )
+      grep_regexp = "^#{port_name}[ \\t]".shellescape
```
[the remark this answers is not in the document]
jreidinger: well, my intention here that it is properly passed to grep, which this should ensure. I am not sure if grep can be tweaked to allow getting root with its params.

library/system/src/modules/Initrd.rb
jreidinger: it should not, but to be sure, I will add it, so it is consistent with rule "escape or explain why not possible".

jreidinger:
@mvidner @shundhammer updated. Thanks for review

library/network/src/modules/PortAliases.rb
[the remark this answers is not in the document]
mvidner: You are concerned with proper _shell_ quoting, which is OK, but you have missed the fact that `QuoteString` (undocumented) is all about _regexp_ quoting. You are not adding a root exploit, just silently breaking the functionality.

library/network/src/lib/network/firewalld.rb
mvidner: No. In this PR alone I can see `.split " "`, `.split "\n"`, `.strip` and `.rstrip` applied on it. QED.

library/packages/src/modules/PackageSystem.rb
mvidner: OK. In case you decide to keep `InitRPMQueryBinary`, please fix `/bin` to `/usr/bin` here, twice: https://github.com/yast/yast-yast2/blame/7f355a06aa1746e0c11fb8315ca9bf2bd655f296/library/packages/src/modules/PackageSystem.rb#L330

library/network/src/modules/PortAliases.rb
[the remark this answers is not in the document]
jreidinger: ok, so I bring back that QuoteString and try to improve a name of it and document it.

library/network/test/port_aliases_test.rb
```diff
@@ -0,0 +1,53 @@
...
+#                  10500/tcp    Reserved
+hip-nat-t          10500/udp    # HIP NAT-Traversal  [RFC5770] [Ari_Keranen]
+MOS-lower          10540/tcp    # MOS Media Object Metadata Port  [Eric_Thorniley]
+MOS-lower          10540/udp    # MOS Media Object Metadata Port  [Eric_Thorniley]
+MOS-upper          10541/tcp    # MOS Running Order Port  [Eric_Thorniley]
+MOS-upper          10541/udp    # MOS Running Order Port  [Eric_Thorniley]
+EOS
+
+describe Yast::PortAliases do
+  describe ".LoadAndReturnNameToPort" do
```
mvidner: Yayyy! Thank you!

library/packages/src/modules/PackageSystem.rb
```diff
@@ -327,8 +327,8 @@ def InitRPMQueryBinary
       )
         @_rpm_query_binary = "/usr/bin/rpmqpack "
         # than rpm itself
-      elsif Ops.greater_than(SCR.Read(path(".target.size"), "/bin/rpm"), -1)
-        @_rpm_query_binary = "/bin/rpm -q "
+      elsif Ops.greater_than(SCR.Read(path(".target.size"), "/usr/bin/rpm"), -1)
```
shundhammer: Boy scout rules: use a plain `>` instead of that ycp-killer `Ops.greater_than()` ? That would make it a lot better to read IMHO.

library/network/src/modules/PortAliases.rb
```diff
@@ -175,34 +177,35 @@ def LoadAndReturnPortToName(port_number)
...
-      grep_regexp = "^#{port_name}[ \\t]".shellescape
-      command = "/usr/bin/grep --perl-regexp #{grep_regexp} /etc/services " \
-        "| /usr/bin/sed \"s/[^ \\t]*[ \\t]*\\([^/ \\t]*\\).*/\\1/\""
-      found = Convert.to_map(SCR.Execute(path(".target.bash_output"), command))
-      alias_found = nil
+      path = File.join(WFM.scr_root, "etc/services")
+      content = ::File.read(path)
+      # find line with given port_name. Line looks like:
+      #   http               80/tcp       # World Wide Web HTTP
+      matching_lines = content.lines.grep(/^#{Regexp.escape(port_name)}\s/)
```
shundhammer: :+1:

library/network/test/port_aliases_test.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    it "returns nil if given port name not found" do
+      expect(Yast::PortAliases.LoadAndReturnNameToPort("hellrouting")).to eq nil
+    end
+
+    it "does not interpret regexp characters" do
+      expect(Yast::PortAliases.LoadAndReturnNameToPort("b.*s")).to eq nil
+    end
+
+    it "respect SCR chroot" do
```
shundhammer: English: respect**s** (yes, I know, nitpicking, but it hurts the eyes :smile: )

jreidinger: good catch. I welcome any grammar improvements :wink:

PR #873, opened 2018-12-11
Security Audit Fixes: Removing unused functions with potential security issues

shundhammer:
This is part of the YaST security audit by @jsegitz.

From the audit:

> yast2/yast2-4.1.6/library/general/src/modules/Misc.rb:233
> 
>  RunCommandWithTimeout should be reworked to receive an absolute path and arguments that then can be quoted/escaped

That function is only used once in all of YaST, and that's in another function _RunDumbTimeout()_ that is also not used.

Removing both of them to get rid of potential security hazards _and_ of dead code.


jsegitz:
well that's obviously the best solution :)

PR #874, opened 2018-12-11
Support for FHS-compliant skelcd packages (part of fate#325482)

ancorgs:
See https://trello.com/c/piNC0YZM/529-part-of-fate-325482-make-installationxml-files-fhs-compliant


library/control/src/modules/WorkflowManager.rb
```diff
@@ -478,6 +467,46 @@ def addon_control_dir(src_id, cleanup: false)
...
+
+    # Full name of the control file located directly in the given directory
+    #
+    # The content of the file is not verified to be compliant with the structure
+    # of a control file, this method simply finds the (hopefully only) XML file
+    # in the directory.
+    #
+    # @param dir [String] directory where the control file is expected to be
+    # @return [String, nil] nil if there is no control file
+    def ctrl_file_at(dir)
```
dgdavid: Here, I prefer to simplify the code to something like

```ruby

files = Dir.glob("#{dir}/*.xml") 

log.error "More than one XML file in #{dir}: #{files.inspect}" if files.size > 1

files.first
```

What do you think?

library/control/src/modules/WorkflowManager.rb
```diff
@@ -478,6 +467,46 @@ def addon_control_dir(src_id, cleanup: false)
...
     end
 
+    # Path of the control file contained in the package that has been previously
+    # extracted to the given directory
+    #
+    # @see #control_file
+    #
+    # @param dir [String] directory where the package has been extracted to
+    # @return [String] name of the control file
+    def control_file_at_dir(dir)
```
dgdavid: It seems to me a bit weird (aka confusing) to have `control_file_at_dir` and `control_file_at` methods, both receiving the _dir_ parameter. 

What about changing the first to something similar to `find_control_file_for`... not a great name, but "slightly" different to the other. Thoughts?

library/control/test/workflow_manager_test.rb
```diff
@@ -386,22 +386,59 @@
...
+        let(:product_files) { [] }
+        let(:role_files) { ["/tmp/usr/share/system-roles/superyast.xml"] }
+        let(:installation_xml) { false }
+
+        it "returns the path of the control file" do
+          expect(subject.control_file(repo_id)).to eq "/tmp/usr/share/system-roles/superyast.xml"
+        end
+      end
+
+      context "if the package contains a control file in the installation-products dir" do
```
jreidinger: maybe just for documentation purpose, can you write spec that if rpm contains combination of all methods, that installation-product has precedence?

library/control/test/workflow_manager_test.rb
```diff
@@ -386,22 +386,59 @@
...
-      expect(subject.control_file(repo_id)).to end_with("/installation.xml")
-    end
+        allow(File).to receive(:exist?) do |name|
+          product_files.include?(name) || role_files.include?(name)
+        end
+
+        allow(File).to receive(:exist?).with(/installation\.xml\z/).and_return installation_xml
+      end
+
+      context "if the package contains a control file in the system-roles dir" do
```
dgdavid: Do you think that it worth it to add another scenario with files in both, `product_files` and `system_roles`?

library/control/test/workflow_manager_test.rb
dgdavid: That is what I meant in https://github.com/yast/yast-yast2/pull/874#discussion_r240717962

ancorgs: @jreidinger I was tempted to write such spec... and more. I didn't do it because I wanted to keep the simplicity of the current test suite (which doesn't test/document what happens when we have `/installation.xml` and also a role file or what happens when we have several role files) since you usually complain that I tend to over-complicate tests. 😏

But I will take your request as a green light to write one of my comprehensive tests. Be prepared. :wink:

library/control/src/modules/WorkflowManager.rb
```diff
@@ -478,6 +467,46 @@ def addon_control_dir(src_id, cleanup: false)
...
+    #
+    # @param dir [String] directory where the control file is expected to be
+    # @return [String, nil] nil if there is no control file
+    def find_control_file(dir)
+      # sadly no glob escaping - https://bugs.ruby-lang.org/issues/8258
+      # but as we generate directory, it should be ok
+      files = Dir.glob("#{dir}/*.xml")
+
+      if files.size > 1
+        log.error "more than one file in #{dir}: #{files.inspect}"
```
dgdavid: NP: I guess you are just keeping the log error as it was, but I would change it a little bit:

> **M**ore than one **XML** file in #{dir}: #{files.inspect}

At the end, it is only looking for _.xml_ files.

ancorgs:
Intentionally not merging yet. We want to build SLE-15-SP1 Beta1 first.

dgdavid:
Closing in favor of #877 (the same, without the merge commit)

PR #876, opened 2018-12-13
Extend Yast::Execute API

joseivanlopez:
Related to PBI: https://trello.com/c/AzEDChwZ/520-issues-from-yast-audit-systemlog-backtickslog

`Yast::Execute.locally!.stdout(...)`
`Yast::Execute.on_target!.stdout(...)`
`Yast::Execute.stdout(...)`

### DIFF
--- a/library/system/src/lib/yast2/execute.rb
+++ b/library/system/src/lib/yast2/execute.rb
@@ -22,6 +22,7 @@
 
 require "yast"
 require "cheetah"
+require "forwardable"
 
 module Yast
   # A module for executing scripts/programs in a safe way
@@ -30,12 +31,34 @@ module Yast
   # as the backend, but adds support for chrooting during the installation.
   # It also globally switches the default Cheetah logger to
   # {http://www.rubydoc.info/github/yast/yast-ruby-bindings/Yast%2FLogger Y2Logger}.
+  #
+  # @example Methods of this class can be chained.
+  #
+  #   Yast::Execute.locally!.stdout("ls", "-l")
+  #   Yast::Execute.stdout.on_target!("ls", "-l")
   class Execute
+    include Yast::I18n
+
     # use y2log by default
     Cheetah.default_options = { logger: Y2Logger.instance }
 
-    extend Yast::I18n
-    textdomain "base"
+    class << self
+      extend Forwardable
+
+      def_delegators :new, :on_target, :on_target!, :locally, :locally!, :stdout
+    end
+
+    # Constructor
+    #
+    # @param options [Hash<Symbol, Object>] options to add for the execution. Some of
+    #   these options are directly passed to Cheetah#run, and others are used to control
+    #   the behavior when running commands (e.g., to indicate if a popup should be shown
+    #   when the command fails). See {#options}.
+    def initialize(options = {})
+      textdomain "base"
+
+      @options = options
+    end
 
     # Runs with chroot; a failure becomes a popup.
     # Runs a command described by *args*,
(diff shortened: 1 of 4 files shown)


library/system/src/lib/yast2/execute.rb
```diff
@@ -37,76 +37,120 @@ class Execute
...
-    # {http://www.rubydoc.info/github/yast/yast-ruby-bindings/Yast%2FLogger Y2Logger}.
-    # @param args see http://www.rubydoc.info/github/openSUSE/cheetah/Cheetah.run
-    def self.locally(*args)
-      popup_error { locally!(*args) }
+    def stdout(*args)
+      self.args = args
+
+      add_options(stdout: :capture)
+
+      run
```
jreidinger: just note, maybe it makes sense to also act same as others. so `args.empty? ? self : run`

library/system/src/lib/yast2/execute.rb
```diff
@@ -37,76 +37,120 @@ class Execute
...
-        }
-      )
+  private
+
+    attr_accessor :args
+
+    attr_reader :options
+
+    def add_options(options)
+      @options.merge!(options)
```
jreidinger: example what I mean by allowing to construct it in different order:

```ruby
module Yast
  class ExecuteClass
     def initialize(options = {})
￼       @options = options
￼     end

     def locally!(*args)
        args.empty? ? self.class.new(options) : run(*args)
     end

      def on_target!(*args)
         root = Yast::WFM.scr_root
         target_class = self.class.new(options.merge(chroot: root)
         args.empty? ? target_class : target_class.run(args)
      end
         
       Execute = ExecuteClass.new # this keeps backward compatibility and allows easy shortcut for simply runs
  end
end
```

library/system/src/lib/yast2/execute.rb
```diff
@@ -30,12 +30,21 @@ module Yast
   # as the backend, but adds support for chrooting during the installation.
   # It also globally switches the default Cheetah logger to
   # {http://www.rubydoc.info/github/yast/yast-ruby-bindings/Yast%2FLogger Y2Logger}.
-  class Execute
+  class ExecuteClass
```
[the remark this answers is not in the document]
jreidinger: But it is not possible as Execute is now object and not namespace. Goal is to keep working existing API and allow chaining of commands.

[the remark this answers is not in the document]
mvidner: Well,  `Execute::Result` is not even necessary. We can simply continue with `on_target` and `locally` as class methods, and have them return instances of the same class so that you can call methods on that.

I do feel strongly about not allowing ExecuteClass.

library/system/src/lib/yast2/execute.rb
```diff
@@ -87,11 +104,119 @@ def self.locally(*args)
...
+    def run_or_chain(args)
+      args.none? ? self : run(*args)
+    end
+
+  private
+
+    # Options to add when running a command
+    #
+    # Some options are intended to control the behavior and they are not passed to
+    # Cheetah.run. For example:
```
jreidinger: I think this statement is not true as it is actually passed to cheetah, which ignore it.

jreidinger: ah, now I see it. OK.

library/system/src/lib/yast2/execute.rb
```diff
@@ -87,11 +104,119 @@ def self.locally(*args)
...
+    # @param new_options [Hash<Symbol, Object>]
+    # @return [ExecuteClass]
+    def chaining_object(new_options)
+      self.class.new(options.merge(new_options))
+    end
+
+    # Runs the given command
+    #
+    # It takes into account the object options when running the command.
+    # Note that `yast_popup` takes precedence over `yast_stdout`. So, when both options
```
jreidinger: please mention also that if popup or stdout is used, that `!` variant does not raise. so `locally!.stdout` does not raise exception.

library/system/src/lib/yast2/execute.rb
joseivanlopez: "yast_*" options are filtered out, see `#command_options` method.

PR #879, opened 2018-12-18
Add configuration_management section to control file

imobachgs:
Add a `configuration_management` to the control file. Related to https://github.com/yast/yast-installation-control/pull/76.
https://fate.suse.com/322722

### DIFF
--- a/library/control/src/modules/ProductFeatures.rb
+++ b/library/control/src/modules/ProductFeatures.rb
@@ -49,7 +49,7 @@ module Yast
       # Default values for features
       # two-level map, section_name -> [ feature -> value ]
       @defaults = {
-        "globals"      => {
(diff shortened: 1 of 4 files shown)


library/control/src/modules/ProductFeatures.rb
```diff
@@ -93,7 +93,8 @@ def main
           "addon_selections"                     => [],
           "inform_about_suboptimal_distribution" => false
         },
-        "network"      => { "force_static_ip" => false }
+        "network"                  => { "force_static_ip" => false },
+        "configuration_management" => {}
```
mvidner: Does this mean there are **no** defaults? I doubt that. We should add a comment pointing to the place where the defaults are defined.

PR #880, opened 2018-12-19
Display the license contained in the fallback repo

dgdavid:
Related to fate#325482 (see https://trello.com/c/ish1gWEZ/568-part-of-fate325482-display-the-license-contained-in-the-fallback-repo)

---

As is stated in the above linked trello card, this PR introduce the necessary changes to display properly the license content when the installer is working with the "fallback repo".

Summarizing them,  

* A [new fetcher was introduced](https://github.com/yast/yast-yast2/pull/880/commits/196de7d24b078148a25c651367c6eb55d3965b49) to read the license from the `.rpm` file.
* Also a [specific handler was needed](https://github.com/yast/yast-yast2/pull/880/commits/920fc91c28026205d47f71227a76f18a8c2c5f21) to check if the license confirmation was required.
* Instead of being passed as a parameter, now [the fetcher is inferred](https://github.com/yast/yast-yast2/pull/880/commits/8d7843f660384fd723b1dba193c734c32cd16586), selecting the first one containing a valid license content (not nil, not empty) for the default language.
* In a similar way, [the chosen handler depends on the selected fetcher](https://github.com/yast/yast-yast2/pull/880/commits/920fc91c28026205d47f71227a76f18a8c2c5f21#diff-afc9898eb30e35e92b815b39b4d55aebR28), which is received as param.

---

**Doable improvements**

~~There is a quite similar codebase in the Rpm's [fetchers](https://github.com/yast/yast-yast2/pull/880/files#diff-c8f8b800d5bf8974e9a7611f82f4d7f6R130) and [handlers](https://github.com/yast/yast-yast2/pull/880/files#diff-c778c028b170012e32d7d2818a336ed0R68) for both, licenses and [releases notes](https://github.com/yast/yast-yast2/blob/ba44c0688199b73418602c54184df8cd4b867905/library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb#L112). Maybe we could extract them an build a single common piece, like "PackageFinder" or similar.~~

---

**Pending tasks**

 - [x] Test it manually (again) _see screenshots in the next comment_
 - [ ] Update version and changelog


dgdavid:
### Some screenshots

#### With a single product (requiring license)
| License and language selection  |
|---|
| Installing single product from fallback repo |

#### With multiple products (requiring license)

> <sub>**NOTE** that screenshots were taken before the commit https://github.com/yast/yast-yast2/pull/880/commits/215a9f55860905036e1a3b8e448d4ba41344b972, which fix the correct value selected/displayed in the language selector.</sub>

|  Product and language selection | License  |
|---|---|
| Multiple products - selection | Multiple products -license |

#### With multiple products (some of them requiring license)

> <sub> :warning: **IMPORTANT** The `openSUSE` product was included here only to test that the `no-acceptance-needed` empty file is used correctly to determine if license acceptance should be required when the installer is dealing with the fallback repo. However, this product is not compatible (at the time of writing) with this type of installation.</sub>

| Product and language selection  | License acceptance required  | License acceptance not required |
|---|---|---|
| Multiple products selection | License required | Optional license |

library/packages/src/lib/y2packager/license.rb
```diff
@@ -87,18 +95,19 @@ def cache
...
     # Bear in mind that `fetcher` will be ignored if `content` is specified.
     #
-    # @param fetcher [:libzypp] Fetcher to retrieve licenses information. For the time
-    #   being, only :libzypp is supported.
-    # @param content [String]   License content. If this argument is given, this
-    #   string is used as the license's content (and `source` is ignored).
-    def initialize(fetcher: nil, content: nil)
+    # @param product_name [String] Product name to retrieve license information
+    # @param content      [String] License content. If this argument is given, this
+    #   string is used as the license's content (and `product_name` is ignored).
```
jreidinger: this is a bit confusing for me as from code it looks like it is used only for DEFAULT_LANG, otherwise it is fetched, not?

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
```diff
@@ -0,0 +1,154 @@
...
+      # @return [Array<String, String>, nil] Array containing content and language code
+      def license_content_for(lang)
+        tmpdir = Dir.mktmpdir
+
+        begin
+          package.extract_to(tmpdir)
+          license_file = license_path(tmpdir, lang) || fallback_path(tmpdir)
+
+          if license_file.nil?
+            log.error("License file not found")
```
jreidinger: if you write error to log try to write for which license file was license was not found.

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
```diff
@@ -0,0 +1,154 @@
...
+      # Fallback license file
+      FALLBACK_LICENSE_FILE = "LICENSE.TXT".freeze
+
+      def fallback_path(directory)
+        log.info("Searching for a fallback #{FALLBACK_LICENSE_FILE} file in #{directory}")
+
+        Dir.glob(
+          File.join(directory, "**", FALLBACK_LICENSE_FILE),
+          File::FNM_CASEFOLD
+        ).first
```
jreidinger: this whole Dir.glob looks almost identical to the one above beside one parameter, what about abstracting it.?

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
```diff
@@ -0,0 +1,154 @@
...
+      # Find the latest available/selected product package
+      #
+      # @return [Y2Packager::Package, nil] Package containing licenses; nil if not found
+      def package
+        return nil if package_name.nil?
+
+        @package ||=
+          Y2Packager::Package
+          .find(package_name)
+          .compact
```
jreidinger: so `find` returns array that should be compacted? Strange Package API.

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
```diff
@@ -0,0 +1,154 @@
...
+        Dir.glob(
+          File.join(directory, "**", FALLBACK_LICENSE_FILE),
+          File::FNM_CASEFOLD
+        ).first
+      end
+
+      # Valid statuses for packages containing licenses
+      AVAILABLE_STATUSES = [:available, :selected].freeze
+
+      # Find the latest available/selected product package
```
jreidinger: as seen from implementation, please mention that it is the one with highest version, not the latest one ( even if it is very often true )

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
```diff
@@ -0,0 +1,154 @@
...
+
+      # Find the package name
+      #
+      # @return [String, nil] the package name for the product; nil if not found
+      def package_name
+        return @package_name if @package_name
+
+        package_properties = Yast::Pkg.ResolvableProperties(product_name, :product, "").first || {}
+
+        @package_name = package_properties.fetch("product_package", nil)
```
jreidinger: above it tries only first product with given name. What if more is found and only the second one has defined product_package? If it cannot happen, please write to code, why.

library/packages/src/lib/y2packager/licenses_handlers.rb
```diff
@@ -19,13 +20,16 @@ module Y2Packager
...
   module LicensesHandlers
-    # Return the licenses proper fetcher for a given source
+    # Return the proper license handler based on given fetcher
+    #
+    # @param fetcher      [LicensesFetchers::Base] Fetcher used as source to fetch license
+    # @param product_name [String]                 Product's name
     #
-    # @param source       [:libzypp,nil] Source to fetch license from (only :libzypp is supported)
-    # @param product_name [String]       Product's name
     # @return [Object]
```
jreidinger: Is it really generic object or Fetcher child?

library/packages/src/lib/y2packager/licenses_handlers/rpm.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      # Determine whether the license should be accepted or not
+      #
+      # @return [Boolean] true if the license acceptance is required
+      def confirmation_required?
+        @package ||= find_package
+
+        return false unless @package
+
+        !ignore_acceptance?
```
jreidinger: to be honest, I do not like the flow method have here. It requires finding package and ignore_acceptance already expect that package is found. What about having lazy load of package ( so not attr_reader ) and if it returns nil, then, return false and having package search here ( so ignore_acceptance method is here )?

library/packages/src/lib/y2packager/product_license.rb
```diff
@@ -49,15 +48,14 @@ class << self
       #
       # This method uses a cache to return an already fetched product license.
       #
-      # @param product_name [String]       Product's name
-      # @param source       [:libzypp,nil] Source to get the license from. For the time being,
-      #   only :libzypp is really supported.
+      # @param product_name [String] Product's name
+      #
       # @return [ProductLicense]
-      def find(product_name, source: nil, content: nil)
```
jreidinger: :+1: having automatic logic in source for finding looks like good step for me.

library/packages/test/y2packager/licenses_fetchers/rpm_test.rb
```diff
@@ -0,0 +1,139 @@
...
+      let(:license_path) { "/fake/path/to/#{lang}/LICENSE" }
+      let(:default_license_path) { "/fake/path/to/default/txt/license" }
+      let(:license_file_content) { "" }
+
+      before do
+        allow(Dir).to receive(:glob).and_call_original
+        allow(Dir).to receive(:glob).with(/#{lang}/, anything).and_return([license_path])
+        allow(Dir).to receive(:glob).with(/LICENSE.TXT/, anything).and_return([default_license_path])
+        allow(File).to receive(:read).with(license_path).and_return(license_file_content)
+        allow(File).to receive(:read).with(default_license_path).and_return(license_file_content)
```
jreidinger: this huge amount of mocking a bit scare me. Is possible to just add to data section few examples rpms on which we do tests? This will also allow in future testing potentially not working rpms ( read it as "we change it and expect that yast will work" )

library/packages/test/y2packager/licenses_handlers/rpm_test.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+    end
+
+    context "when 'no-acceptance-neeed' file is not found" do
+      let(:found_paths) { [] }
+
+      it "returns true" do
+        expect(handler.confirmation_required?).to eq(true)
+      end
+    end
```
jreidinger: same here. Real rpm would be nice.

library/packages/src/lib/y2packager/license.rb
dgdavid: Yes, you are right. I will update the documentation to something like

```ruby

@param content [String] License content. If given, this string will be used as the license's content for the default languge, ignoring the `product_name`.
```
Sounds better for you?

library/packages/src/lib/y2packager/licenses_fetchers.rb
```diff
@@ -12,20 +12,34 @@
...
   #
-  # Licenses can be retrieved from different places (libzypp, URLs, etc.). The classes
-  # defined in this module are able to retrieve licenses contents.
+  # Licenses can be retrieved from different places (libzypp, URLs, etc.). The classes defined in
+  # this module are able to retrieve licenses contents.
   module LicensesFetchers
-    # Return the licenses proper fetcher for a given source
+    include Yast::Logger
+
+    KNOWN_SOURCES = [:libzypp, :rpm].freeze
```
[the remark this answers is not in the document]
dgdavid: :+1:

library/packages/src/lib/y2packager/licenses_fetchers/libzypp.rb
```diff
@@ -20,8 +20,11 @@ class Libzypp < Base
       # Return the license text to be confirmed
       #
       # @param lang [String] Language
-      # @return [String,nil] Product's license; nil if the product or the license were not found.
+      #
+      # @return [String, nil] Product's license; nil if the product or the license were not found.
       def content(lang)
+        super
```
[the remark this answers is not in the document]
dgdavid: Again, you are right. Thank you very much! :)

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
```diff
@@ -0,0 +1,154 @@
...
+          return []
+        end
+
+        @locales ||=
+          begin
+            tmpdir = Dir.mktmpdir
+            package.extract_to(tmpdir)
+            # TODO: Use rpm -qpl file.rpm instead?
+            license_files = Dir.glob(File.join(tmpdir, "**", "LICENSE.*.TXT"), File::FNM_CASEFOLD)
+            languages = license_files.map { |path| path[/LICENSE.(\w*).TXT/i, 1] }
```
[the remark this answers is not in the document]
dgdavid: I am not sure if I understand what you meant.

Here the code is simply testing paths ignoring case sensitivity but returning the right value.

See the examples below

```bash
irb(main):001:0> files = [
irb(main):002:1* "/fake/path/to/LiCeNSe.cz_CZ.TXT",
irb(main):003:1* "/fake/path/to/license.cz.txt",
irb(main):004:1* "/fake/path/to/LICENSE.cz_CZ.TXT",
irb(main):005:1* "/fake/path/to/license.CZ.TxT"
irb(main):006:1> ]
=> ["/fake/path/to/LiCeNSe.cz_CZ.TXT", "/fake/path/to/license.cz.txt", "/fake/path/to/LICENSE.cz_CZ.TXT", "/fake/path/to/license.CZ.TxT"]  
irb(main):007:0> files.map { |path| path[/LICENSE.(\w*).TXT/i, 1] }
=> ["cz_CZ", "cz", "cz_CZ", "CZ"]
irb(main):008:0>
```

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
dgdavid: At this point, the neither license file was found, file for requested lang/translation nor fallback file. Do you want to add the `lang` to the error message?

If you take a look to [`license_path`](https://github.com/yast/yast-yast2/blob/ca4665afee0c806fb619b0541c6c51c40b6bd4a8/library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb#L99) and [`fallback_path`](https://github.com/yast/yast-yast2/blob/ca4665afee0c806fb619b0541c6c51c40b6bd4a8/library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb#L115) methods, you can see that we are already loggin infirmation about what we are looking for. Isn't it enough?

jreidinger: well, I think it is better to have all important info in that error call. So when you searching logs for errors, you see it immediatelly. It is important especially when multiple licenses are needed like in SLE with several addons

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
dgdavid: Probably not. It could be an error on my side at the time to write the test.

See https://github.com/yast/yast-yast2/blob/ca4665afee0c806fb619b0541c6c51c40b6bd4a8/library/packages/test/y2packager/licenses_fetchers/rpm_test.rb#L35 and https://github.com/yast/yast-yast2/blob/ca4665afee0c806fb619b0541c6c51c40b6bd4a8/library/packages/test/y2packager/licenses_fetchers/rpm_test.rb#L109

I'll to fix it.

library/packages/src/lib/y2packager/licenses_handlers.rb
dgdavid: A LicensesHandlers child, which at the end is an Object. I'll change the documentation to `LicensesHandlers::Base` :)

library/packages/test/y2packager/licenses_fetchers/rpm_test.rb
dgdavid: I agree. I'll try to do what you are proposing.

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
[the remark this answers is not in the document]
dgdavid: Could you please give me an example of how to document that?

library/packages/src/lib/y2packager/licenses_fetchers/rpm.rb
```diff
@@ -23,7 +23,7 @@ class Rpm < Base
       #
       # @return [String, nil] Product's license; nil if the product or the license were not found
       def content(lang)
-        super
+        return @default_content if default_lang?(lang) && @default_content
```
jreidinger: I am not sure if we understand each other. Super was fine for me, just here you ignore its output so simple change can be
```
result = super
return result if result
```

dgdavid: Ouch! I did it in a first attempt to accomplish your suggestion... but it seems a bit weird to me. Let me to bring it back.

dgdavid: After test a bit more this, I think we could keep this logic here. Moving it again to the base class and using those two additional lines in each child seems weird to me. Too much indirection, isn't it?

jreidinger: fine for me.

library/packages/src/lib/y2packager/licenses_handlers/rpm.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      # Set the license confirmation for the product
+      #
+      # @param confirmed [Boolean] true if it should be accepted; false otherwise
+      def confirmation=(confirmed)
+        if confirmed
+          log.info("License was accepted")
+        else
+          log.info("License was not accepted")
+        end
```
jreidinger: so only goal of this accessor is only to log its value? That looks strange for me.

library/packages/src/lib/y2packager/package.rb
```diff
@@ -43,6 +43,20 @@ def find(name)
         return nil if props.nil?
         props.map { |i| new(i["name"], i["source"], i["version"]) }
       end
+
+      # Find the highest version of available/selected package by
+      #
+      # @return [Y2Packager::Package, nil] Package containing licenses; nil if not found
+      def highest_available_or_selected(name)
```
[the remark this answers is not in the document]
dgdavid: I like the idea :)

library/packages/src/lib/y2packager/licenses_handlers/rpm.rb
dgdavid: Just for consistency

library/packages/src/lib/y2packager/package.rb
```diff
@@ -43,6 +43,23 @@ def find(name)
...
+      # @param statuses [Array<Symbol>] allowed package statuses
+      #
+      # @return [Y2Packager::Package, nil] Highest found version of package; nil if not found
+      def last_version(name, statuses: [:available, :selected])
+        packages = find(name)
+
+        return nil unless packages
+
+        packages
+          .select { |i| [:available, :selected].include?(i.status) }
```
jreidinger: you should use here that `statuses` and when you will do that force push, please also update commit message with new name of method.

dgdavid: Oops! You know what people says: "Speed isn't good" :man_facepalming: I'll fix it in less than a second :running_man: 

:)

package/yast2.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Jan  3 11:37:36 UTC 2019 - dgonzalez@suse.com
+
+- Fetch (and display) properly the license content when using a
+  product definition located in /usr/share/installation-products
```
jreidinger: I think we should mention here, that it is location inside rpm package and not in insts-sys. At least if I get how code works.

dgdavid: You are right. Thank you.

PR #882, opened 2018-12-27
Cleanup

jreidinger:
cleaning of unused stuff and reorganize a bit directory structure.


library/wizard/test/wizard_test.rb
```diff
@@ -0,0 +1,62 @@
+#! /usr/bin/env rspec
```
dgdavid: I guess we need to include here the missing copyright comments, isn't it?

https://github.com/yast/yast-yast2/blob/master/library/systemd/test/yast2/compound_service_test.rb#L4-L21

library/wizard/test/wizard_test.rb
```diff
@@ -0,0 +1,62 @@
...
+      allow(Yast::UI).to receive(:WizardCommand).and_return(false)
+      expect(Yast::UI).to receive(:ChangeWidget).with(anything, :HelpText, "test")
+
+      subject.SetHelpText("test")
+    end
+  end
+
+  describe ".UserInput" do
+    it "returns result of UI UserInput" do
+      allow(Yast::UI).to receive(:UserInput).and_return(:ok)
```
dgdavid: NP: I admit that I prefer to use the allow outside of `it`, usually in a `before` block and then use `context` to set the return value. Something like

```ruby
describe ".UserInput" do
  before do
    allow(Yast::UI).to receive(:UserInput).and_return(user_input)
  end

  context "when user input is :ok" do
    let(:user_input) { :ok }

    it "returns result of UI UserInput" do
      expect(subject.UserInput).to eq :ok
    end
  do

  context "when user accepts" do
    let(:user_input) { :accept }

    it "returns :next" do
      expect(subject.UserInput).to eq :next
    end
  end

  context "when user cancels" do
    let(:user_input) { :cancel }

    it "returns :back" do
      expect(subject.UserInput).to eq :back
    end
  end
end
```

library/wizard/test/wizard_test.rb
jreidinger: :+1:

library/wizard/test/wizard_test.rb
jreidinger: I know that there is a bit disagreement, as I prefer more explicit tests which is from first sight visible what each api call returns and this style make it a less obvious for me. My code is just two related calls, one sets user input and second just check return value without need to check another contexts, lets and other stuff.

package/yast2.spec
```diff
@@ -183,8 +183,8 @@ mkdir -p %{buildroot}%{_sysconfdir}/YaST2
 %{yast_agentdir}/ag_*
 %{_fillupdir}/sysconfig.yast2
 
+%{_datadir}/bash-completion/completions/yast2*.sh
 # configuration files
-%config %{_sysconfdir}/bash_completion.d/yast2*.sh
```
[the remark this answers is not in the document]
jreidinger: rpmlint warning that /etc/ is not proper place for bash completion provided by system. In /etc should be only user config.

jreidinger:
@mvidner yes, I see that. But I am not sure if we should touch it now when wayland comes into play, so hopefully will be killed by that.

PR #883, opened 2018-12-28
Add missing help text for ServiceWidget

dgdavid:
At the time to continue the work (yast/yast-firewall#109) started by @teclator in the yast-firewall module, I realized that the ServiceWidget has not a help text defined yet. In fact, none of modules that already use the new ServiceWidget shows any related help.

That's the reason why I am proposing to add a generic text, vía the `help` widget method, to all of them.

### DIFF
--- a/library/cwm/src/lib/cwm/service_widget.rb
+++ b/library/cwm/src/lib/cwm/service_widget.rb
@@ -56,5 +56,9 @@ module CWM
     def init
       refresh
     end
+
+    def help
+      @service_widget.help
+    end
   end
 end
--- a/library/systemd/src/lib/yast2/service_widget.rb
+++ b/library/systemd/src/lib/yast2/service_widget.rb
@@ -124,6 +124,34 @@ module Yast2
       store_autostart
     end
 
+    # Returns the service widget help text
+    #
+    # @return [String]
+    def help
+      helptext = "<h2>Service configuration</h2>"
+      # TRANSLATORS: helptext for the service current status
+      helptext << _(
+        "<h3>Current status</h3>" \
+        "Displays the curren status of the service."
+      )
+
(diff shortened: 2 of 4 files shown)


dgdavid:
@jreidinger, what do you think about this? If looks good for you, shall I increase the version number of yast-yast2 module before merge this <sub>small</sub> change?

library/cwm/src/lib/cwm/service_widget.rb
```diff
@@ -56,5 +56,30 @@ def store
...
+      )
+
+      # TRANSLATORS: helptext for the "After reboot" service widget option
+      helptext << _(
+        "<h3>After reboot</h3>" \
+        "Let choose if service should be started automatically on boot. Some services could be
+        configured <b>on demand</b>, which means that the associated socket will be running and in
+        charge of start the service if needed."
+      )
+    end
```
jreidinger: service widget itself does not have its help? BTW textdomain should be same in whole yast module. So check what other files in yast2.rpm uses as textdomain.

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -124,6 +124,34 @@ def store
...
+        options depend on the current state. The <b>Keep current state</b> special action leaves the
+        service state untouched."
+      )
+
+      # TRANSLATORS: helptext for the "After reboot" service widget option
+      helptext << _(
+        "<h3>After reboot</h3>" \
+        "Let choose if service should be started automatically on boot. Some services could be
+        configured <b>on demand</b>, which means that the associated socket will be running and in
+        charge of start the service if needed."
```
jreidinger: maybe I would simplify this with `will be running and start the service if needed.`

dgdavid: Done!

PR #886, opened 2019-01-09
Extend installation dialog

joseivanlopez:
Needed for PBI: https://trello.com/c/53IEUxrB/577-susecaas-platform-4-p2-1099485-caasp-specific-partitioning-warning-dialog-missing-in-storage-ng

### DIFF
--- a/library/wizard/src/modules/Wizard.rb
+++ b/library/wizard/src/modules/Wizard.rb
@@ -92,6 +92,42 @@ module Yast
       @have_fancy_ui_cache
     end
 
+    # Returns a button box with buttons "Back", "Cancel", "Next"
+    #
+    # @return [Yast::Term] a widget tree representing the buttons
+    def BackCancelNextButtonBox
+      HBox(
+        HWeight(
+          1,
+          ReplacePoint(
+            Id(:rep_help),
+            PushButton(Id(:help), Opt(:key_F1, :helpButton), Label.HelpButton)
+          )
+        ),
+        HStretch(),
+        HWeight(
+          1,
+          ReplacePoint(
+            Id(:rep_back),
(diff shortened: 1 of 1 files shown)


library/general/src/lib/ui/installation_dialog.rb
```diff
@@ -135,12 +135,26 @@ def create_dialog
...
         help_text,
-        Yast::GetInstArgs.enable_back,
-        Yast::GetInstArgs.enable_next || Yast::Mode.normal
+        enable_back,
+        enable_next
       )
       true
     end
 
+    # Whether the back button is enabled
```
jreidinger: please document behavior in this implementation ( so read options from client arguments ).

library/wizard/src/modules/Wizard.rb
```diff
@@ -92,6 +92,42 @@ def haveFancyUI
...
+        HWeight(
+          1,
+          ReplacePoint(
+            Id(:rep_next),
+            PushButton(Id(:next), Opt(:key_F10, :default), Label.NextButton)
+          )
+        )
+      )
+    end
+
```
jreidinger: hmm, I am not much happy with this additional code increase which result that this ruby file has almost 2k lines. I know that it is a bit more work, but is possible to somehow refactor ( in sense of API and functionality) this module to reduce huge duplication here and decrease size?

joseivanlopez: Yes, maybe we can refac that code, but the problem I see is that thd changes here are for SP0, so I prefer a more conservative change here. I would prefer to do such refactorization as part of another PBI, and not for SP0.

joseivanlopez:
Finally these changes were not needed, see https://github.com/yast/yast-storage-ng/pull/821.

PR #888, opened 2019-01-11
Cwm improvements2

jreidinger:


library/cwm/src/lib/cwm/popup.rb
```diff
@@ -34,6 +36,12 @@ def should_open_dialog?
       true
     end
 
+    # Popup does not allow nil, so overwrite Dialog default value.
+    # @return [String,nil] The dialog title.
```
teclator: If it does not allow `nil`, then the return should be modified accordingly, isn't it?

jreidinger: yes, you are right. Nil cause libyui red warning.

jreidinger:
@mvidner popup is subclass of dialog, so I believe that having it only in dialog is enough. Documentation in dialog is missing. I overlook that I did not write it on friday and forgot to add it today. Will do it. Thanks for notice.

teclator:
LGTM

PR #890, opened 2019-01-16
Master fate 327099

schubi2:


library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -65,10 +66,25 @@ def installation_package_mapping
     #
     # @return [Array<Product>] Available products
     def all_products
+      linuxrc_special_products = if Yast::Linuxrc.InstallInf("specialproduct")
+        Yast::Linuxrc.InstallInf("specialproduct").split(",")
+      else
+        []
+      end
+
       @all_products ||= available_products.map do |prod|
```
lslezak: Functionality OK, I'd just use [each_with_object](https://ruby-doc.org/core-2.5.1/Enumerable.html#method-i-each_with_object) for better readability, it will avoid removing `nil`s at the end, something like (please check, I'm not sure this is actually correct): 
```ruby
@all_products ||= available_products.each_with_object([]) do |prod, all_products|
  ...
  if special_product_tag && ...
    next
  end

  all_products << Y2Packager::Product.new(
    ... )
end
```

PR #891, opened 2019-01-17
Date and time widgets

teclator:
- Trello Card: https://trello.com/c/DS4ULBxR/541-finish-implementation-of-input-widgets

This PR adds CWM widgets for `Date` and `Time` fields.

**Note:** It seems that both are now supported in ncurses and this [check](https://github.com/yast/yast-journal/blob/66f6eebee05903264b9893193a0cf574c4541694/src/lib/y2journal/time_helpers.rb#L57) is not needed probably [since](https://github.com/libyui/libyui-ncurses/commit/e1d3e8043130c64ca2217622f607ff274ab81eb6)

## Example screenshot (Y2STYLE=installation.qss)
<p align="center">
  
</p>

## Ncurses example screenshot
dateandtimefieldsncurses

### DIFF
--- a/library/cwm/examples/date_time.rb
+++ b/library/cwm/examples/date_time.rb
@@ -1,0 +1,69 @@
+# Simple example to demonstrate object API for CWM
+
+require_relative "example_helper"
+
+require "cwm"
+require "cwm/popup"
+
+Yast.import "CWM"
+
+class Name < CWM::InputField
+  def initialize
+    textdomain "example"
+  end
+
+  def label
+    _("Name")
+  end
+end
+
+class EventDate < CWM::DateField
+  def initialize
+    textdomain "example"
+  end
+
+  def init
+    self.value = Time.now.strftime("%Y-%m-%d")
+  end
+
+  def label
+    _("Event date")
+  end
+end
+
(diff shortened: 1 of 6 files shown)


library/cwm/examples/date_time.rb
```diff
@@ -0,0 +1,69 @@
+# Simple example to demonstrate object API for CWM
```
jreidinger: looks like copy and paste

library/cwm/src/lib/cwm/common_widgets.rb
```diff
@@ -366,4 +366,22 @@ class RichText < AbstractWidget
...
+  end
+
+  # Date field widget
+  # The {#label} method is mandatory.
+  class DateField < AbstractWidget
+    self.widget_type = :date_field
+
+    include ValueBasedWidget
+    abstract_method :label
+  end
```
jreidinger: please add also shared rspec specification for them in cwm/rspec

library/cwm/examples/date_time.rb
teclator: It was based on the popup example so, basically copied & modified example, yes :smile:

PR #892, opened 2019-01-18
fix broken yast module startup (bsc#1122473)

wfeldt:
For some reason the constructed command line from the ncurses control center
looks like this:
> y2start bootloader --arg ncurses -name YaST2 -icon yast

with an superfluous '--arg'. I've no idea where that comes from but
the check in this patch is a safe-guard against this anyway.

### DIFF
--- a/scripts/yast2
+++ b/scripts/yast2
@@ -295,8 +295,10 @@ if [ -n "$CMDLINE_HELP" ]; then
 fi
 ARGS=( )
 for i in "$@"; do
-  ARGS+=( "--arg" )
-  ARGS+=( "$i" )
+  if [ -n "$i" ] ; then
+    ARGS+=( "--arg" )
+    ARGS+=( "$i" )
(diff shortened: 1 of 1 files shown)


mvidner:
This looks like a duplicate of https://github.com/yast/yast-yast2/pull/887

scripts/yast2
```diff
@@ -295,8 +295,10 @@ if [ -n "$CMDLINE_HELP" ]; then
 fi
 ARGS=( )
 for i in "$@"; do
-  ARGS+=( "--arg" )
-  ARGS+=( "$i" )
+  if [ -n "$i" ] ; then
```
jreidinger: as @mvidner already pointed out it should be fixed by different patch. This will explicitly filter all intentionally empty parameters, which is quite big change. so :-1: for me

wfeldt:
not needed

PR #894, opened 2019-01-30
save_y2logs: save kernel messages and udev log (bsc#1089647, bsc#1085…

hramrach:
…212)

Kernel messages can show up in /var/log/boot.msg (linuxrc redirects there) and
/var/log/messages (syslog).

Since bsc#1085212 linuxrc can run udev in debug mode. In this case udev logs
into /var/log/udev.log.


jreidinger (approved):

jreidinger:
change is fine, probably not worth release on its own so version change is not needed.

PR #896, opened 2019-02-05
Interface zone mapping improvements

teclator:
- Do not delete the **ZONE** attribute when empty ("default zone")
- Added methods to `Interface` and `Zone` for changing the current map.

### DIFF
--- a/library/network/src/lib/y2firewall/firewalld/interface.rb
+++ b/library/network/src/lib/y2firewall/firewalld/interface.rb
@@ -96,6 +96,15 @@ module Y2Firewall
         fw.zones.find { |z| z.interfaces.include?(name) }
       end
 
+      # Assign the interface to the given zone
+      #
+      # @param zone_name [String] the name of the zone to be assigned to
+      def zone=(zone_name)
+        fw.zones.map { |z| z.remove_interface(name) if z.interfaces.include?(name) }
+        z = fw.find_zone(zone_name)
+        z && z.add_interface(name)
+      end
+
     private
 
       # Return an instance of Y2Firewall::Firewalld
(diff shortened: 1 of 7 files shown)


library/network/src/lib/y2firewall/firewalld/interface.rb
```diff
@@ -96,6 +96,15 @@ def zone
         fw.zones.find { |z| z.interfaces.include?(name) }
       end
 
+      # Assign the interface to the given zone
+      #
+      # @param zone_name [String] the name of the zone to be assigned to
+      def zone=(zone_name)
+        fw.zones.map { |z| z.remove_interface(name) if z.interfaces.include?(name) }
```
dgdavid: Why `.map`? I think `.each` could be enough, isn't it?

dgdavid: Where is `name` defined?

teclator: https://github.com/yast/yast-yast2/blob/f318a9239b0ec015610f75c5c05f85123650222e/library/network/src/lib/y2firewall/firewalld/interface.rb#L78

dgdavid: Oops! Apologies :) I must to wake up as soon as possible. 

Sorry for the noise.

teclator: Yep, it is.

As it was added here:

https://github.com/yast/yast-yast2/pull/896/files#diff-5f18f6aba33e047bd40791544760f288R144

Probably the inertia.. :smile:

Issue #895, opened 2019-02-05
Fails to build with ruby 2.6


coolo:
The test suite fetches the wrong release notes?

This is from https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:F/yast2/standard/i586 and only happens on i586 (but 100% reproducible)

```
[  127s]    yast2 4.1.53: library/packages/test/test-suite.log
[  127s] ========================================================
[  127s] 
[  127s] # TOTAL: 26
[  127s] # PASS:  25
[  127s] # SKIP:  0
[  127s] # XFAIL: 0
[  127s] # FAIL:  1
[  127s] # XPASS: 0
[  127s] # ERROR: 0
[  127s] 
[  127s] .. contents:: :depth: 2
[  127s] 
[  127s] FAIL: y2packager/release_notes_fetchers/rpm_test
[  127s] ================================================
[  127s] 
[  127s] ..F........
[  127s] 
[  127s] Failures:
[  127s] 
[  127s]   1) Y2Packager::ReleaseNotesFetchers::Rpm#release_notes when a full language code is given (xx_XX) and release notes are not available returns product release notes for the short language code (xx)
[  127s]      Failure/Error: expect(rn.content).to eq("Versionshinweise\n")
[  127s] 
[  127s]        expected: "Versionshinweise\n"
[  127s]             got: "Release Notes\n"
[  127s] 
[  127s]        (compared using ==)
[  127s] 
[  127s]        Diff:
[  127s]        @@ -1,2 +1,2 @@
[  127s]        -Versionshinweise
[  127s]        +Release Notes
[  127s]      # ./y2packager/release_notes_fetchers/rpm_test.rb:65:in `block (5 levels) in <top (required)>'
[  127s] 
[  127s] Finished in 0.29556 seconds (files took 0.57383 seconds to load)
[  127s] 11 examples, 1 failure
[  127s] 
[  127s] Failed examples:
[  127s] 
[  127s] rspec ./y2packager/release_notes_fetchers/rpm_test.rb:63 # Y2Packager::ReleaseNotesFetchers::Rpm#release_notes when a full language code is given (xx_XX) and release notes are not available returns product release notes for the short language code (xx)
[  127s] 
[  127s] FAIL y2packager/release_notes_fetchers/rpm_test.rb (exit status: 1)

lslezak:
Fixed in https://github.com/yast/yast-yast2/pull/897

PR #897, opened 2019-02-15
Fixed evaluating release notes

lslezak:
## Problem

- The `./y2packager/release_notes_fetchers/rpm_test.rb` unit test randomly failed in OBS (see #895).

From a debugging session:

```
(byebug) p = File.join(directory, "**", "RELEASE-NOTES.{#{langs.join(",")}}.#{prefs.format}")
"/tmp/d20190215-22316-ra8tss/**/RELEASE-NOTES.{de_DE,de,en}.txt"
(byebug) Dir.glob(p)
["/tmp/d20190215-22316-ra8tss/usr/share/doc/release-notes/dummy/RELEASE-NOTES.en.txt",
"/tmp/d20190215-22316-ra8tss/usr/share/doc/release-notes/dummy/RELEASE-NOTES.de.txt"]
```

So the glob for the `RELEASE-NOTES.{de_DE,de,en}.txt` files returns `"RELEASE-NOTES.en.txt", "RELEASE-NOTES.de.txt"` files (in this order!). So the order of the files might be different than in the input glob, the code does not expect this and uses the first one (via `.first`).

## Additional Problems

- The commandline tests always failed for me when running locally, however it was fine in Travis, Jenkins or OBS so I ignored that. This time I checked the details and the problem is that locally I have installed the byebug debugger. The test raises some exception and expect to print something in the exception handler. However when the debugger is installed YaST displays a popup asking to start the debugger. (In the test mode there is just dummy UI so in reality nothing is displayed in the end.)

- The firewall test failed when started via `rake test:unit` by passed when started via `rspec ./library/network/test/firewalld_wrapper_test.rb`. It turned out that when running in full test the `Yast::NetworkInterfaces.List` method returns some known interfaces, when running standalone the result is empty list.

### DIFF
--- a/library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb
+++ b/library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb
@@ -174,9 +174,15 @@ module Y2Packager
         langs << prefs.user_lang.split("_", 2).first if prefs.user_lang.include?("_")
         langs << prefs.fallback_lang
 
-        path = Dir.glob(
-          File.join(directory, "**", "RELEASE-NOTES.{#{langs.join(",")}}.#{prefs.format}")
-        ).first
+        path = nil
+        langs.any? do |lang|
+          path = Dir.glob(
+            File.join(directory, "**", "RELEASE-NOTES.#{lang}.#{prefs.format}")
+          ).first
+
+          !path.nil?
+        end
+
         return nil if path.nil?
         [path, path[/RELEASE-NOTES\.(.+)\.#{prefs.format}\z/, 1]] if path
       end
--- a/library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb
+++ b/library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb
@@ -175,12 +175,12 @@ module Y2Packager
         langs << prefs.fallback_lang
 
         path = nil
-        langs.any? do |lang|
+        langs.each do |lang|
           path = Dir.glob(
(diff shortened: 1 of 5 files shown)


library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb
```diff
@@ -174,9 +174,15 @@ def release_notes_file(directory, prefs)
         langs << prefs.user_lang.split("_", 2).first if prefs.user_lang.include?("_")
         langs << prefs.fallback_lang
 
-        path = Dir.glob(
-          File.join(directory, "**", "RELEASE-NOTES.{#{langs.join(",")}}.#{prefs.format}")
-        ).first
+        path = nil
+        langs.any? do |lang|
```
jreidinger: it is strange to call any when you ignore return value. So maybe each with break would be better?

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Feb 18 07:47:37 UTC 2019 - lslezak@suse.cz
+
+- Fixed license file ordering issue causing a random test failure
+  (gh#yast-yast2#895)
```
jreidinger: I worry that this jenkins won't accept as valid bugzilla id and neither OBS.

library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb
lslezak: Hm, I was thinking how to implement this, I'd need something similar to `.find` but which return the  evaluated value, not the item from the list. `each` with `break` also does not look nice, but it's probably still a bit better...

PR #898, opened 2019-02-26
Don't use nonexisting programs

mvidner:
-   https://bugzilla.suse.com/show_bug.cgi?id=1127138
-   https://trello.com/c/plarcsbX/770-1-sle15-sp1-no-mkdir-is-not-in-usr-sbin

### DIFF
--- a/library/network/src/modules/NetworkPopup.rb
+++ b/library/network/src/modules/NetworkPopup.rb
@@ -195,7 +195,7 @@ module Yast
         UI.OpenDialog(Label(_("Scanning for hosts on this LAN...")))
         # #71064
         # this works also if ICMP broadcasts are ignored
-        cmd = "/usr/sbin/rpcinfo -b mountd 1 | /usr/bin/cut -d ' ' -f 2 | /usr/bin/sort -u"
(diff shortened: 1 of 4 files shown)


library/network/src/modules/NetworkPopup.rb
```diff
@@ -195,7 +195,7 @@ def NFSServer(selected)
         UI.OpenDialog(Label(_("Scanning for hosts on this LAN...")))
         # #71064
         # this works also if ICMP broadcasts are ignored
-        cmd = "/usr/sbin/rpcinfo -b mountd 1 | /usr/bin/cut -d ' ' -f 2 | /usr/bin/sort -u"
+        cmd = "/sbin/rpcinfo -b mountd 1 | /usr/bin/cut -f 2 | /usr/bin/sort -u"
```
jreidinger: commit message does not explain why also -d is removed.

mvidner: You're right. Explained now.

> The delimiter in its output is a tab, not a space

It's because the two rpcinfo paths are actually different implementations with a slightly different output. See also https://github.com/yast/yast-nfs-client/pull/80 for full history.

PR #899, opened 2019-03-01
Made Report.LongError to use adjustable Popup's *LongError methods

mchf:
Used in https://github.com/yast/yast-ruby-bindings/pull/227

### DIFF
--- a/library/general/src/modules/Report.rb
+++ b/library/general/src/modules/Report.rb
@@ -492,6 +492,8 @@ module Yast
 
     # Store new message text, the text is displayed in a richtext widget - long lines are automatically wrapped
     # @param [String] message_string message text (it can contain rich text tags)
+    # @param width [Integer] width of popup (@see Popup#LongMessageGeometry)
+    # @param height [Integer] height of popup (@see Popup#LongMessageGeometry)
     # @return [void]
     def LongMessage(message_string, width: 60, height: 10)
       Builtins.y2milestone(1, "%1", message_string) if @log_messages
@@ -552,6 +554,8 @@ module Yast
 
     # Store new warning text, the text is displayed in a richtext widget - long lines are automatically wrapped
     # @param [String] warning_string warning text (it can contain rich text tags)
+    # @param width [Integer] width of popup (@see Popup#LongWarningGeometry)
+    # @param height [Integer] height of popup (@see Popup#LongWarningGeometry)
     # @return [void]
     def LongWarning(warning_string, width: 60, height: 10)
       Builtins.y2warning(1, "%1", warning_string) if @log_warnings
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Report.rb
```diff
@@ -493,14 +493,14 @@ def Message(message_string)
     # Store new message text, the text is displayed in a richtext widget - long lines are automatically wrapped
     # @param [String] message_string message text (it can contain rich text tags)
     # @return [void]
-    def LongMessage(message_string)
+    def LongMessage(message_string, width: 60, height: 10)
```
jreidinger: please document new parameters in documentation

library/general/src/modules/Report.rb
```diff
@@ -553,14 +553,14 @@ def Warning(warning_string)
     # Store new warning text, the text is displayed in a richtext widget - long lines are automatically wrapped
     # @param [String] warning_string warning text (it can contain rich text tags)
     # @return [void]
-    def LongWarning(warning_string)
+    def LongWarning(warning_string, width: 60, height: 10)
```
jreidinger: same here, please document parameters

library/general/src/modules/Report.rb
```diff
@@ -595,14 +595,14 @@ def Error(error_string)
     # Store new error text, the text is displayed in a richtext widget - long lines are automatically wrapped
     # @param [String] error_string error text  (it can contain rich text tags)
     # @return [void]
-    def LongError(error_string)
+    def LongError(error_string, width: 60, height: 10)
```
jreidinger: and here

library/general/test/report_test.rb
```diff
@@ -44,15 +44,15 @@
       let(:show) { false }
 
       it "does not show a popup" do
-        expect(Yast::Popup).to_not receive(meth)
+        expect(Yast::Popup).to_not receive("#{meth}Geometry")
```
jreidinger: hmm, I though that receive should get symbol and not string. Do you test if it really fail if it is called?

mchf: well I'm not the author of tests, however it was failing when I patched sources before updating the testsuite ;-)

jreidinger: ok, so looks like it accepts also string. So I also learn something new here :+1:

library/general/src/modules/Report.rb
```diff
@@ -594,15 +598,17 @@ def Error(error_string)
 
     # Store new error text, the text is displayed in a richtext widget - long lines are automatically wrapped
     # @param [String] error_string error text  (it can contain rich text tags)
+    # @param width [Integer] width of popup (@see Popup#LongErrorGeometry)
+    # @param height [Integer] height of popup (@see Popup#LongErrorGeometry)
     # @return [void]
-    def LongError(error_string)
+    def LongError(error_string, width: 60, height: 10)
```
lslezak: Hm, optional parameters....will that work via the YaST component system? The problem is that the `Report` module might be used from Perl code (yast2-users)....

lslezak: For the record: according to @jreidinger it should work across languages.

jreidinger: For the record let me describe it how it works.

Ruby bindings uses `publish` to get API with types. Limitation is that publish support only single signature. So in this case one, two or three params can be used, but only one have to picked. When called this method from e.g. perl, it will use ruby calls, so fulfill all missing params with its default value. But keyword arguments is not supported by ruby bindings and component system, so probably some hash workaround have to be used. In this case publish has only string param and no keyword ones, so it should work without issues.

PR #900, opened 2019-03-04
Fixed command line error reporting, fixed testsuite

lslezak:
- It is related to this change: https://github.com/yast/yast-ruby-bindings/pull/227
- It turned out that the `Report.LongError` call did not support the command line mode (as the `Report.Error` does), so after changing that in the global exception handler the popup was displayed even in the command line mode.
- Added a debugging hint into `commandline_test.rb`, running `byebug` is a bit tricky there.
- Switched from `STDOUT` (the default stdout stream) to `$stdout` (the current stdout stream) (Just a minor thing, but it would handle possible stdout redirection better.)

### DIFF
--- a/library/commandline/test/commandline_test.rb
+++ b/library/commandline/test/commandline_test.rb
@@ -19,18 +19,22 @@ describe Yast::CommandLine do
     allow(Yast::Debugger).to receive(:installed?).and_return(false)
   end
 
+  # NOTE: when using the byebug debugger here temporarily comment out
+  # all "expect($stdout)" lines otherwise the byebug output will be
+  # lost in the rspec mocks and you won't see anything.
+
   it "invokes initialize, handler and finish" do
-    expect(STDOUT).to receive(:puts).with("Initialize called").ordered
-    expect(STDOUT).to receive(:puts).with("something").ordered
-    expect(STDOUT).to receive(:puts).with("Finish called").ordered
+    expect($stdout).to receive(:puts).with("Initialize called").ordered
+    expect($stdout).to receive(:puts).with("something").ordered
+    expect($stdout).to receive(:puts).with("Finish called").ordered
 
     Yast::WFM.CallFunction("dummy_cmdline", ["echo", "text=something"])
   end
 
   it "displays errors and aborts" do
-    expect(STDOUT).to receive(:puts).with("Initialize called").ordered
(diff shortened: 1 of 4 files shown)


ancorgs:
Definitely fixing his test doesn't see to be the answer. We need to fix yast2-ruby-bindings to not completely break the command line mode.

library/commandline/test/commandline_test.rb
```diff
@@ -19,18 +19,22 @@
...
     Yast::WFM.CallFunction("dummy_cmdline", ["echo", "text=something"])
   end
 
   it "displays errors and aborts" do
-    expect(STDOUT).to receive(:puts).with("Initialize called").ordered
-    expect(Yast::CommandLine).to receive(:Print).with(/I crashed/).ordered
-    expect(STDOUT).to_not receive(:puts).with("Finish called")
+    expect($stdout).to receive(:puts).with("Initialize called").ordered
+    expect(Yast::Report).to receive(:LongError).with(/I crashed/, anything).ordered
+    expect($stdout).to_not receive(:puts).with("Finish called")
```
jreidinger: that is strange, how it can work previously? Before that PR we also open Error when program crashes.

jreidinger:
@ancorgs I do not agree to make change in ruby bindings. I try to keep global handler as simple as possible because it is last line of defense and should not fail itself. Ideally  where I see better change is in command line itself which when running in CLI mode have own exception catcher and it will report it as it wants. But I am not sure how it will work when you call another client as it is catched on WFM level.

library/commandline/test/commandline_test.rb
mchf: That's what I would also like to know. Especially when it builds in travis, so may be an inconsistency in build envs?

library/general/src/modules/Report.rb
```diff
@@ -605,7 +605,9 @@ def LongError(error_string, width: 60, height: 10)
       Builtins.y2error(1, "%1", error_string) if @log_errors
 
       if @display_errors
-        if Ops.greater_than(@timeout_errors, 0)
+        if Mode.commandline
```
jreidinger: Well, when you touch it, is possible to fix also longWarning and LongInfo?

lslezak: Done in #901 .

PR #901, opened 2019-03-05
Command line support in Report module (related to bnc#1127685)

lslezak:
- Related to #900, fixing even more `Report` methods to support the command line mode
- 4.1.59

### DIFF
--- a/library/general/src/modules/Report.rb
+++ b/library/general/src/modules/Report.rb
@@ -478,7 +478,9 @@ module Yast
       Builtins.y2milestone(1, "%1", message_string) if @log_messages
 
       if @display_messages
-        if Ops.greater_than(@timeout_messages, 0)
+        if Mode.commandline
+          CommandLine.Print(message_string)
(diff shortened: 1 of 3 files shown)


library/general/src/modules/Report.rb
```diff
@@ -561,7 +569,9 @@ def LongWarning(warning_string, width: 60, height: 10)
       Builtins.y2warning(1, "%1", warning_string) if @log_warnings
 
       if @display_warnings
-        if Ops.greater_than(@timeout_warnings, 0)
+        if Mode.commandline
+          CommandLine.Print("Warning: #{error_string}")
+        elsif Ops.greater_than(@timeout_warnings, 0)
```
jreidinger: thanks for fixing it. BTW it also shows how badly original API was designed by so many code duplications. Maybe something like Yast2::Popup should be done also for Report.

lslezak: Sure, this module deserves a bigger refactoring, but for SP2...

PR #902, opened 2019-03-05
Remove the XVersion API and config file

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=1144627

Back in 2006 SUSE Linux was migrating from XFree86 to X.org which used different filesystem paths, so YaST got an internal config file to manage the transition.
It is time to delete that obsolete code.

Do it only after yast-installation and yast-country stop using this:
- [x] https://github.com/yast/yast-installation/pull/780
- [x] https://github.com/yast/yast-country/pull/223


library/commandline/test/clients/dummy_cmdline.rb
```diff
@@ -2,6 +2,7 @@ module Yast
   class DummyCmdlineClient < Client
     def main
       Yast.import "CommandLine"
+      textdomain "base"
```
lslezak: Please use the "example" text domain, otherwise all texts from this file will end up in the `base.pot` file and the translators will translate them. The `example.pot` file is ignored (not sent to weblate).

```suggestion
      textdomain "example"
```

package/yast2.spec
```diff
@@ -137,6 +137,7 @@ This package contains scripts for handling YAST logs.
 %setup -q
 
 %check
+export Y2STRICTTEXTDOMAIN=1
```
lslezak: We should check all other packages, fix the possible issues and make this the default...

mvidner: Yes, that makes sense. We need to make a staging (local?) build of all packages first.

PR #906, opened 2019-03-08
Do not crash with internal error when activating network service

mchf:
~~bnc#112788~~
bnc#1127798

### DIFF
--- a/library/network/src/modules/NetworkService.rb
+++ b/library/network/src/modules/NetworkService.rb
@@ -47,9 +47,10 @@ require "shellwords"
 
 module Yast
   class NetworkServiceClass < Module
-    # @current_name - current network backend identification, nil is valid value for "no service selected / running"
-    # @cached_name  - the new network backend identification, nil is valid value for "no service selected / running"
-    attr_accessor :current_name, :cached_name
+    # return [String, nil] current network backend identification, nil is valid value for "no service selected / running"
+    attr_accessor :current_name
+    # return [String, nil] new network backend identification, nil is valid value for "no service selected / running"
+    attr_accessor :cached_name
 
     # network backend identification to service name mapping
     BACKENDS = {
--- a/library/network/src/modules/NetworkService.rb
+++ b/library/network/src/modules/NetworkService.rb
@@ -93,6 +93,7 @@ module Yast
     # @param force [Boolean] if action should be forced
     # @return exit code
     def RunSystemCtl(service, action, force: false)
+      raise ArgumentError, "No network service defined." if service.nil?
       cmd = "/usr/bin/systemctl "\
         "#{force ? "--force" : ""} " \
         "#{action.shellescape} " \
--- a/library/network/src/modules/NetworkService.rb
+++ b/library/network/src/modules/NetworkService.rb
@@ -49,6 +49,7 @@ module Yast
   class NetworkServiceClass < Module
     # @current_name - current network backend identification, nil is valid value for "no service selected / running"
     # @cached_name  - the new network backend identification, nil is valid value for "no service selected / running"
+    attr_accessor :current_name, :cached_name
 
     # network backend identification to service name mapping
     BACKENDS = {
@@ -209,12 +210,12 @@ module Yast
     def EnableDisableNow
       return if !Modified()
 
-      if @current_name
-        stop_service(@current_name)
-        disable_service(@current_name)
+      if current_name
+        stop_service(current_name)
+        disable_service(current_name)
       end
 
-      RunSystemCtl(BACKENDS[@cached_name], "enable", force: true) if @cached_name
+      RunSystemCtl(BACKENDS[cached_name], "enable", force: true) if cached_name
 
       @initialized = false
       Read()
--- a/library/network/src/modules/NetworkService.rb
+++ b/library/network/src/modules/NetworkService.rb
@@ -47,8 +47,8 @@ require "shellwords"
 
 module Yast
   class NetworkServiceClass < Module
-    # @current_name - current network backend identification
-    # @cached_name  - the new network backend identification
+    # @current_name - current network backend identification, nil is valid value for "no service selected / running"
+    # @cached_name  - the new network backend identification, nil is valid value for "no service selected / running"
 
     # network backend identification to service name mapping
     BACKENDS = {
@@ -209,10 +209,12 @@ module Yast
     def EnableDisableNow
       return if !Modified()
 
-      stop_service(@current_name)
-      disable_service(@current_name)
+      if @current_name
+        stop_service(@current_name)
+        disable_service(@current_name)
+      end
 
-      RunSystemCtl(BACKENDS[@cached_name], "enable", force: true)
+      RunSystemCtl(BACKENDS[@cached_name], "enable", force: true) if @cached_name
 
       @initialized = false
       Read()
--- a/library/network/test/network_service_test.rb
+++ b/library/network/test/network_service_test.rb
@@ -23,6 +23,13 @@ describe Yast::NetworkService do
         expect { Yast::NetworkService.is_netconfig }.not_to raise_error
       end
     end
+
+    describe "#EnableDisableNow" do
+      it "does not crash when current / cached service is nil" do
+        allow(Yast::NetworkService).to receive(:Modified).and_return(true)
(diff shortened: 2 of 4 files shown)


library/network/test/network_service_test.rb
```diff
@@ -23,6 +23,13 @@
...
+
+    describe "#EnableDisableNow" do
+      it "does not crash when current / cached service is nil" do
+        allow(Yast::NetworkService).to receive(:Modified).and_return(true)
+        expect { Yast::NetworkService.EnableDisableNow }.not_to raise_error
+      end
+    end
   end
 
   describe "#RunSystemCtl" do
```
teclator: please, add also the test for the raised `Execption`

library/network/test/network_service_test.rb
```diff
@@ -23,6 +23,13 @@
         expect { Yast::NetworkService.is_netconfig }.not_to raise_error
       end
     end
+
+    describe "#EnableDisableNow" do
+      it "does not crash when current / cached service is nil" do
```
teclator: Instead of not crashing what about describing the behavior...

```ruby
context "when a network service is running" do
  it "stops the current service"  do
  end
  it "disables the current service" do
  end
end

context "when a network service has been selected to be enabled" do
  it "enables the selected service" do
  end
end
....
```

mchf: it is test which covers the issue from the bug report

teclator: True, but the logic is not so big, so what about improving the test coverage?

mchf: ok, will add something ;-)

teclator: The responsibility of this module / class is well defined and isolated, so it does not need to be rewritten / re-factorized as much as other classes, but the test coverage is very low, then we should improve when we are able to, and I think it will not take so much time.

mchf: yes, I agree and already promised that will add some tests ;-)

library/network/test/network_service_test.rb
```diff
@@ -39,5 +46,52 @@
...
+      expect { Yast::NetworkService.RunSystemCtl(nil, "enable") }.to raise_error
+    end
+  end
+
+  describe "#EnableDisableNow" do
+    subject { Yast::NetworkService }
+
+    context "When changing running service" do
+      before(:each) do
+        allow(subject).to receive(:current_name).and_return(:netconfig)
```
teclator: this before block and allow can be moved up, and then use let(:current_name) in each context

library/network/test/network_service_test.rb
```diff
@@ -39,5 +46,52 @@
...
+
+        # using anything instead of exact service name because of some magic in identifying the service in the system
+        expect(subject).to receive(:RunSystemCtl).with(anything, /stop|kill/)
+        expect(subject).to receive(:RunSystemCtl).with("network", "disable")
+      end
+
+      it "disables old service and enables new one" do
+        allow(subject).to receive(:cached_name).and_return(:wicked)
```
teclator: same with this

library/network/test/network_service_test.rb
```diff
@@ -39,5 +46,52 @@
...
+
+        expect(subject).to receive(:RunSystemCtl).with("wicked", "enable", any_args)
+
+        Yast::NetworkService.EnableDisableNow
+      end
+
+      it "only disables old service when no network service was requested" do
```
teclator: and this could be a context where you declare the cached_name as nil

library/network/test/network_service_test.rb
```diff
@@ -39,5 +46,52 @@
...
+        expect(subject).to receive(:RunSystemCtl).with(anything, /stop|kill/)
+        expect(subject).to receive(:RunSystemCtl).with("network", "disable")
+      end
+
+      it "disables old service and enables new one" do
+        allow(subject).to receive(:cached_name).and_return(:wicked)
+
+        expect(subject).to receive(:RunSystemCtl).with("wicked", "enable", any_args)
+
+        Yast::NetworkService.EnableDisableNow
```
teclator: why not used suject?

library/network/test/network_service_test.rb
```diff
@@ -39,5 +46,52 @@
...
+      end
+
+      it "only disables old service when no network service was requested" do
+        allow(subject).to receive(:cached_name).and_return(nil)
+
+        expect(subject).not_to receive(:RunSystemCtl).with("wicked", "enable", any_args)
+
+        Yast::NetworkService.EnableDisableNow
```
teclator: subject¿

library/network/test/network_service_test.rb
```diff
@@ -39,5 +46,52 @@
...
+    context "When activating a service if none is running" do
+      before(:each) do
+        allow(subject).to receive(:current_name).and_return(nil)
+        allow(subject).to receive(:cached_name).and_return(:wicked)
+      end
+
+      it "activates new service" do
+        expect(subject).to receive(:RunSystemCtl).with("wicked", "enable", any_args)
+
+        Yast::NetworkService.EnableDisableNow
```
teclator: subject

library/network/test/network_service_test.rb
```diff
@@ -39,5 +46,52 @@
...
+        allow(subject).to receive(:cached_name).and_return(nil)
+
+        expect(subject).not_to receive(:RunSystemCtl).with("wicked", "enable", any_args)
+
+        Yast::NetworkService.EnableDisableNow
+      end
+    end
+
+    context "When activating a service if none is running" do
+      before(:each) do
```
teclator: read below

teclator:
In general, looks good, just some np comments about unit tests

library/network/test/network_service_test.rb
mchf: well, matter of preference. And I currently don't see any benefit from switching to your case.

library/network/test/network_service_test.rb
mchf: i believe that there is many ways how to write tests ... and each of us prefer one of them ;-)

library/network/test/network_service_test.rb
mchf: can be

library/network/test/network_service_test.rb
teclator: well, it is a good practice for DRY

teclator:
@mchf just make Travis happy and it will be approved

library/network/test/network_service_test.rb
mchf: i prefer to keep tests short and readable ... e.g. to have mocks as close as possible. I don't like to read whole test when searching how a particular value is mocked.

library/network/test/network_service_test.rb
teclator: Sure, but there are also conventions, style guidelines and good practices that we should follow as good citizens :stuck_out_tongue_winking_eye: 

http://www.betterspecs.org/

library/network/test/network_service_test.rb
teclator: The mocked value is closed to the text as it it declared in the context itself.
```
context.. do
  let(:current_name) { :netconfig }
end
```
and in the text there are mainly expectations.

But it is ok as it is

teclator:
> bnc#112788
> bnc#112798

- [bsc#1127798](https://bugzilla.suse.com/show_bug.cgi?id=1127798)

library/network/test/network_service_test.rb
mchf: I'm not nice guy today ;-) But seriously ... I've even problem with some rubocop checks, so convention is good ... but I break them time to time on purpose.

library/network/src/modules/NetworkService.rb
```diff
@@ -47,8 +47,9 @@
 
 module Yast
   class NetworkServiceClass < Module
-    # @current_name - current network backend identification
-    # @cached_name  - the new network backend identification
+    # @current_name - current network backend identification, nil is valid value for "no service selected / running"
+    # @cached_name  - the new network backend identification, nil is valid value for "no service selected / running"
```
teclator: NP: as you have used an accessor it should be documented as

```
# @return [String, nil] ... and split it in two lines for each variable
```

PR #907, opened 2019-03-12
Fix ProductFeatures#InitIfNeeded

dgdavid:
## Problem

Thanks to a change made in the licenses location (FATE#324053 / SLE-4173), we realized that the [`ProductFeatures#InitIfNeeded`](https://github.com/yast/yast-yast2/blob/2c99f30df4996605c7149fdb6c78c01583f1011c/library/control/src/modules/ProductFeatures.rb#L193) method is not working as expected, since it is not restoring the available features values in the running system.

This could be easily checked fixing by hand the wrong value for `base_product_license_directory` in the `/etc/YaST2/ProductFeatures` file in a SLE-115-SP1, changing it from _/etc/YaST2/licenses/base_ to _/usr/share/licenses/product/base_ and running the yast2-firstboot module: it skips the _License agreement_ step.

## Solution

Add tests and fix the method.

---

Related PR and discussions, 

* https://github.com/yast/skelcd-control-Kubic/pull/31
* https://github.com/yast/skelcd-control-openSUSE/pull/175
* https://github.com/yast/skelcd-control-leanos/pull/36
* https://github.com/yast/skelcd-control-CAASP/pull/79
* https://github.com/yast/yast-installation/pull/784


library/control/test/ProductFeatures_test.rb
```diff
@@ -158,4 +158,95 @@
...
+        subject.InitFeatures(true)
+      end
+
+      it "does not force to initialize them again" do
+        expect(subject).to receive(:InitFeatures).with(false)
+
+        subject.GetFeature("globals", "base_product_license_directory")
+      end
+
+      context "but it is in normal stage" do
```
lslezak: I'd simplify just to `in normal stage`.

library/control/test/ProductFeatures_test.rb
```diff
@@ -158,4 +158,95 @@
...
+        let(:normal_stage) { true }
+
+        it "reads the value from the running system" do
+          # value readed from data/etc/YaST2/ProductFeatures file
+          expect(subject.GetFeature("globals", "base_product_license_directory"))
+            .to eq("/usr/share/licenses/product/base")
+        end
+      end
+
+      context "but it is in firstboot stage" do
```
lslezak: ... and `in firstboot stage` here

library/control/test/ProductFeatures_test.rb
```diff
@@ -158,4 +158,95 @@
...
+        let(:firstboot_stage) { true }
+
+        it "reads the value from the running system" do
+          # value readed from data/etc/YaST2/ProductFeatures file
+          expect(subject.GetFeature("globals", "base_product_license_directory"))
+            .to eq("/usr/share/licenses/product/base")
```
lslezak: I guess the default in https://github.com/yast/yast-yast2/blob/2fe30e8a1435a14da21a9f11dee5d8ceb59979e0/library/control/src/modules/ProductFeatures.rb#L74 will be later changed from `/etc/YaST2/licenses/base/` to `/usr/share/licenses/product/base`. In that case we cannot easily decide if the value was really read from the system. I'd probably use some custom value in the testing data/etc/YaST2/ProductFeatures file just for the testing purposes, so we can be sure it's really different than the default.

dgdavid: > will be later changed from /etc/YaST2/licenses/base/ to /usr/share/licenses/product/base

I am not sure, because the value has changed for some products, but not for all of them. So, shall we really change the current value? I don't think so.

Take a look to https://jira.suse.de/browse/SLE-4173 and https://build.suse.de/request/show/185130

What do you think?

dgdavid: > I'd probably use some custom value in the testing data/etc/YaST2/ProductFeatures file just for the testing purposes

Ok, I'll change to a custom value.

lslezak: > I am not sure, because the value has changed for some products, but not for all of them. So, shall we really change the current value? I don't think so.

Ugh, really? Which product is missing? I'd suppose the same value would be used for all SLE15-SP1 product family. Maybe it is a bug. Ask Behlert if that's correct.

dgdavid: As we were talking in IRC, TW, Leap and others. Not SLE15-SP1 family.

library/control/test/ProductFeatures_test.rb
```diff
@@ -158,4 +158,95 @@
...
+        subject.InitFeatures(true)
+      end
+
+      it "does not force to initialize them again" do
+        expect(subject).to receive(:InitFeatures).with(false)
+
+        subject.GetFeature("globals", "base_product_license_directory")
+      end
+
+      context "in normal stage" do
```
imobachgs: Does it apply only when _features are already initialized_? Because as you stated in the outer context, they will be initialized anyway when calling `GetFeature`. I would move these examples to the outer context.

library/control/test/ProductFeatures_test.rb
```diff
@@ -158,4 +158,95 @@
...
+      allow(Yast::Stage).to receive(:normal).and_return(normal_stage)
+      allow(Yast::Stage).to receive(:firstboot).and_return(firstboot_stage)
+    end
+
+    around do |example|
+      change_scr_root(scr_root_dir, &example)
+    end
+
+    context "when features are not initialized yet" do
+      it "initializes them" do
```
imobachgs: I would 1) something like `expect(subject).to receive(:InitIfNeeded)` or 2) remove this example. See the next comment to understand why.

library/control/test/ProductFeatures_test.rb
```diff
@@ -158,4 +158,95 @@
...
+        expect(subject).to receive(:InitFeatures).with(false)
+
+        subject.GetFeature("globals", "base_product_license_directory")
+      end
+
+      context "in normal stage" do
+        let(:normal_stage) { true }
+
+        it "reads the value from the running system" do
+          # value readed from data/etc/YaST2/ProductFeatures file
```
imobachgs: `readed` does not exist: just use `read`.

dgdavid: :see_no_evil:

PR #912, opened 2019-03-19
Require `tar` for yast2 logs

dgdavid:
Add the `tar` dependency to `yast2-logs`.

_Tested manually_

Screenshot from 2019-03-19 17-32-16

Related to https://bugzilla.suse.com/show_bug.cgi?id=1125142


imobachgs:
@dgdavid Oh, I overlooked that the bug number was not included in the changes file. Please, add it :)

PR #921, opened 2019-04-29
Added Hostname class in y2network

teclator:
- Added a basic Hostname class by now.
- The hostname reader is just the same we had in DNS module.


teclator:
Closed by now as it was added into yast2-network

PR #925, opened 2019-05-15
Make sure the wizard buttons always remain visible in NCurses

shundhammer:
# Trello

https://trello.com/c/kUpqSPGu

# Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1133367

# Problem

In the NCurses UI, the wizard buttons can disappear when the wizard content is too large.

registration-ncurses-80x24-broken
_Notice there are no wizard buttons visible._

# Cause

Unlike in the Qt UI, the NCurses UI does not have a dedicated Wizard widget; a wizard dialog is pieced together from standard widgets. This includes the outer wizard dialog, the wizard buttons and the framework for the wizard contents which is basically a fancy ReplacePoint where the application later puts the dialog content.

Since all those parts were created equal, in case the UI runs out of screen space because a wizard dialog is overcrowded, it has to cut off widgets somewhere. In the standard algorithm it cuts off some part of all the widgets in the dimension where there is not enough space; so the dialog content loses some space, and the wizard buttons also lose some space.

But since an NCurses button is only one line high, it will quickly disappear if that happens in the vertical dimension, making the buttons useless.

# Fix

Give higher layout priority to the wizard buttons, i.e. lower the layout priority for the dialog content in the vertical dimension.

This is commonly done with a VWeight widget: In a VBox, each item gets the space that it reports it needs, and excess space is distributed evenly or, if there are VWeights, according to those VWeights.

In the opposite case when there is not enough space, the widgets _without_ a VWeight get their desired space, and the rest goes to the widgets with a VWeight. If there is only one of those, it gets all the remaining space.

# Screenshots

registration-ncurses-80x24-fixed
_80x24: The problematic case in this bug. Notice the wizard buttons are now there alright._

registration-ncurses-128x40-fixed
_128x40: Works fine with much larger than standard resolution._

registration-ncurses-80x20-fixed
_80x20: Even works fine with much less resolution (but this is really pushing the limits)._

# Effect on other dialogs and modules

They won't ever lose their wizard buttons again. They will be the last screen element to disappear.

So this fix fixes not only this particular case, but all NCurses wizard buttons in all dialogs and modules forevermore.

### DIFF
--- a/library/packages/src/lib/y2packager/product_upgrade.rb
+++ b/library/packages/src/lib/y2packager/product_upgrade.rb
@@ -44,6 +44,18 @@ module Y2Packager
       ["openSUSE"]                                                        => "SLES"
     }.freeze
 
+    # This maps uses a list of installed products as the key and the removed products as a value.
+    # All products in key have to be installed.
+    # It is used in special cases when the removed product is still available on the medium
+    # and it is selected to upgrade automatically by the solver (not by YaST or the user).
+    # However, the required behavior is to *uninstall* the product.
+    UPGRADE_REMOVAL_MAPPING = {
+      # the SLES + SUMA Proxy + SUMA Branch Server is upgraded to SUMA Branch Server
+      # (see the definition above), but we need to explicitly remove the old SUMA Proxy
+      # and the SLES product to avoid automatic upgrade by the solver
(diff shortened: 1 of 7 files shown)


shundhammer:
The diff is best viewed with _Diff Settings_ -> _Hide Whitespace Changes_.

PR #928, opened 2019-05-17
create log directory with control.xml and merged installation.xml

aschnell:
For https://trello.com/c/xKoN23o3/906-2-include-the-controlxml-to-be-collected-by-savey2logs.

### DIFF
--- a/library/general/src/lib/yast2/log_dir_rotator.rb
+++ b/library/general/src/lib/yast2/log_dir_rotator.rb
@@ -1,0 +1,195 @@
+# encoding: utf-8
+
+# ***************************************************************************
+#
+# Copyright (c) 2018-2019 SUSE LLC
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact Novell, Inc.
+#
+# To contact Novell about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com.
+#
+# ***************************************************************************
+
+require "yast"
+require "fileutils"
+
+Yast.import "Mode"
+Yast.import "Directory"
+
+module Yast2
+  # The class LogDirRotator handels creating, rotating and removing
+  # directories for logging in the YaST2 log directory according to
+  # this schema:
+  #
+  #   rm #{NAME}-03
+  #   mv #{NAME}-02 #{NAME}-03
+  #   mv #{NAME}-01 #{NAME}-02
+  #   mv #{NAME}    #{NAME}-01
+  #
+  # System error are logged and no exception is raised by the functions.
+  #
+  # TODO This class is based on dump_manager.rb and could be used there.
+  class LogDirRotator
+    include Yast::Logger
+
+    # The number of old log directories to keep. This is in addition
+    # to the current one. Child classes can override the value. Max
+    # value is 99.
+    KEEP_OLD_LOG_DIRS = 3
+
+    # The name of the directory in /var/log/YaST2 to create, rotate
+    # and delete. Child classes should override the value.
+    NAME = "undefined".freeze
+
+    # Prepare (create, rotate and delete) the log directory.
+    def prepare
+      return if Yast::Mode.test
+      begin
+        log.info "preparing log dir #{self.class::NAME}"
+        rotate_log_dirs
+        FileUtils.mkdir_p(log_dir)
+      rescue SystemCallError => e
+        log.error e.to_s
+      end
+    end
+
+    # Copy a file to the log directory.
+    #
+    # @param src [String] name of source
+    # @param dest [String] basename of destination
+    def copy(src, dest)
+      return if Yast::Mode.test
+      begin
+        FileUtils.cp(src, log_dir + "/" + dest)
+      rescue SystemCallError => e
+        log.error e.to_s
+      end
+    end
+
+    # Write a file in the log directory.
+    #
+    # @param dest [String] basename of destination
+    # @param content [String] content to write
+    def write(dest, content)
+      return if Yast::Mode.test
+      begin
+        File.open(log_dir + "/" + dest, "w") { |file| file.write(content) }
+      rescue SystemCallError => e
+        log.error e.to_s
+      end
+    end
+
+    # Return a suitable name for the devicegraph log directory
+    # depending on the YaST mode (installation / installed system).
+    #
+    # @return [String] directory name with full path
+    def log_dir
+      dir = installation? ? "#{self.class::NAME}-inst" : self.class::NAME
+      base_dir + "/" + dir
+    end
+
+  private
+
+    # Return true if this is some installation mode: installation, update
+    #
+    # @return [Boolean]
+    def installation?
+      Yast::Mode.installation || Yast::Mode.update
+    end
+
+    # Rotate the log directories, depending on current YaST mode:
+    #
+    # During installation (or update or AutoYaST), clear and remove any old
+    # /var/log/YaST2/#{NAME}-inst directory.
+    #
+    # In the installed system, keep a number of old log directories, remove
+    # any older ones in /var/log/YaST2, and rename the ones to keep:
+    #
+    #   rm -rf #{NAME}-03
+    #   mv #{NAME}-02 #{NAME}-03
+    #   mv #{NAME}-01 #{NAME}-02
+    #   mv #{NAME}    #{NAME}-01
+    #
+    # This will NOT create any new log directory.
+    def rotate_log_dirs
+      return unless File.exist?(base_dir)
+      if installation?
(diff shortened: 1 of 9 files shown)


library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,195 @@
...
+    # to the current one. Child classes can override the value. Max
+    # value is 99.
+    KEEP_OLD_LOG_DIRS = 3
+
+    # The name of the directory in /var/log/YaST2 to create, rotate
+    # and delete. Child classes should override the value.
+    NAME = "undefined".freeze
+
+    # Prepare (create, rotate and delete) the log directory.
+    def prepare
```
jreidinger: is separate call needed? Why not have it in initialize? then simple LogDirRotator.new will be enough instead of calling prepare explicitely.

library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    # Write a file in the log directory.
+    #
+    # @param dest [String] basename of destination
+    # @param content [String] content to write
+    def write(dest, content)
+      return if Yast::Mode.test
+      begin
+        File.open(log_dir + "/" + dest, "w") { |file| file.write(content) }
```
jreidinger: It can be written as `File.write(log_dir + "/" + dest, content)` https://ruby-doc.org/core-2.5.0/IO.html#method-c-write

library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,195 @@
...
+    #   mv #{NAME}    #{NAME}-01
+    #
+    # This will NOT create any new log directory.
+    def rotate_log_dirs
+      return unless File.exist?(base_dir)
+      if installation?
+        kill_old_log_dirs([File.basename(log_dir)])
+      else
+        log_dirs = old_log_dirs.sort
+        keep_dirs = log_dirs.shift(self.class::KEEP_OLD_LOG_DIRS)
```
jreidinger: this won;t work when KEEP_OLD_LOG_DIRS will be bigger then 10 as it is string sorting. See this output of irb:
```
irb(main):004:0> a = []; 11.times { |i| a << "t#{i}" }; a.sort
=> ["t0", "t1", "t10", "t2", "t3", "t4", "t5", "t6", "t7", "t8", "t9"]
```

library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    # Kill (recursively remove) old log directories.
+    #
+    # @param log_dirs [Array<String>] directory names (without path) to remove
+    def kill_old_log_dirs(log_dirs)
+      log_dirs.each do |dir|
+        next unless File.exist?(base_dir + "/" + dir)
+        log.info("Removing old log dir #{dir}")
+        FileUtils.remove_dir(base_dir + "/" + dir)
```
jreidinger: this will potentially raise exception.  I suggest to use `FileUtils.remove_dir(base_dir + "/" + dir, true)`. See https://ruby-doc.org/stdlib-2.4.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-remove_dir

library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,195 @@
...
+    def write(dest, content)
+      return if Yast::Mode.test
+      begin
+        File.open(log_dir + "/" + dest, "w") { |file| file.write(content) }
+      rescue SystemCallError => e
+        log.error e.to_s
+      end
+    end
+
+    # Return a suitable name for the devicegraph log directory
```
jreidinger: here it mention devicegraph, which is probably not intended.

library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,195 @@
...
+    #
+    #   mv #{NAME}-02 #{NAME}-03
+    #   mv #{NAME}-01 #{NAME}-02
+    #   mv #{NAME}    #{NAME}-01
+    #
+    # @param old_name [String] old directory name (without path)
+    def rename_old_log_dir(old_name)
+      new_name =
+        if old_name =~ /[0-9]+$/
+          old_name.next
```
jreidinger: nice usage of next :+1:

library/general/src/lib/yast2/log_dir_rotator.rb
aschnell: Yes, one might not need the prepare at other code locations. And the class is not a singleton.

library/general/src/lib/yast2/log_dir_rotator.rb
aschnell: The dirs are t-01, t-02, ... so it does work.

library/general/src/lib/yast2/log_dir_rotator.rb
aschnell: Exceptions are logged as documented.

library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,195 @@
...
+    # Rename an old log directory according to this schema:
+    #
+    #   mv #{NAME}-02 #{NAME}-03
+    #   mv #{NAME}-01 #{NAME}-02
+    #   mv #{NAME}    #{NAME}-01
+    #
+    # @param old_name [String] old directory name (without path)
+    def rename_old_log_dir(old_name)
+      new_name =
+        if old_name =~ /[0-9]+$/
```
jreidinger: I suggest something like `/\d+\z/`. In ruby `$` in regexp means matching also endline. So potentianally if filepath contain newline[1] this won't work as intented. and `\d` is shorter then `[0-9]`. More about ruby regexp at https://ruby-doc.org/core-2.5.1/Regexp.html#class-Regexp-label-Regexp-23match+method

[1] https://superuser.com/questions/129519/which-file-systems-support-newlines-in-file-names

library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,195 @@
...
+          old_name.next
+        else
+          old_name + "-01"
+        end
+      log.info("Rotating log dir #{old_name} to #{new_name}")
+      File.rename(base_dir + "/" + old_name, base_dir + "/" + new_name)
```
jreidinger: NP: this can potentionally can raise SystemCallError. But I do not know what is proper reaction in that case.

package/yast2.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri May 17 11:17:17 CEST 2019 - aschnell@suse.com
+
+- create log directory with control.xml and merged installation.xml
```
jreidinger: I worry that it will need bsc, jira or fate number.

aschnell: Not for factory (according to the openSUSE rules). For SP2 we can assume that it gets accepted together with other changes that do have a number.

library/general/src/lib/yast2/log_dir_rotator.rb
aschnell: Exceptions are logged as documented.

library/general/src/lib/yast2/log_dir_rotator.rb
aschnell: Why should the filename contain newline in the first place. This code is reused and has worked fine for some time.

library/general/src/lib/yast2/log_dir_rotator.rb
jreidinger: ah, right

library/general/src/lib/yast2/log_dir_rotator.rb
jreidinger: Well, reason is security. It is not critical for me, but in general for security reasons it is recommented to use `\A\z` instead `^$` in ruby. The second one worked, but can be tricked to not do what expected.

library/general/src/lib/yast2/log_dir_rotator.rb
```diff
@@ -0,0 +1,196 @@
...
+    # value is 99.
+    KEEP_OLD_LOG_DIRS = 3
+
+    # The name of the directory in /var/log/YaST2 to create, rotate
+    # and delete. Child classes should override the value.
+    NAME = "undefined".freeze
+
+    # Prepare (create, rotate and delete) the log directory.
+    def prepare
+      return if Yast::Mode.test
```
imobachgs: Why is this line needed? If you, for some reason, need to keep it, please document in the method's description.

jreidinger: my guess is that ProductControl is used in old testsuite and you want to skip this file modification in old testsuite.

aschnell: Because otherwise the old testsuite fails. But I can also document that.

imobachgs: Thanks, that's what I wanted to know. Mentioning it would be fine so I do not ask the question again in the future :)

library/general/src/lib/yast2/log_dir_rotator.rb
aschnell: OK, then I suggest you create a bug report so that is also gets fixed in yast2-storage-ng.

PR #932, opened 2019-05-29
Drop old testsuite

jreidinger:
- Drop old testsuite
- Convert from autotools to rake based installation

trello: https://trello.com/c/lmE34oq8/1021-drop-old-testsuite


library/network/test/network_interfaces_test.rb
```diff
@@ -23,7 +23,11 @@ def mock_ppp
     .and_return "512\n"
 end
 
-describe Yast::NetworkInterfaces do
+# FIXME: find why it does not work. It looks like some problem with SCR agent
+# registration as .network.section is empty
+# Y2DEBUG mention "Agent registration: /home/abuild/rpmbuild/BUILD/yast2-4.2.4/library/network/src/scrconf last changed at Thu Jan  1 00:00:00 1970"
```
mvidner: The last_changed code is in fact unused.

mvidner: @jreidinger strace tells me you have removed /usr/share/YaST2/scrconf/network.scr

mvidner:
@jreidinger this shows the problem:

```console
$ diff -u <(rpm -qlp /tmp/yast2-4.2.4-old.x86_64.rpm ) <(rpm -qlp /tmp/yast2-4.2.4-new.x86_64.rpm )
--- /dev/fd/63  2019-05-29 17:42:05.193113470 +0200                                                                                                                                                                
+++ /dev/fd/62  2019-05-29 17:42:05.193113470 +0200
@@ -287,7 +287,7 @@
 /usr/share/YaST2/scrconf/etc_ssh_sshd_config.scr
 /usr/share/YaST2/scrconf/etc_sysctl_conf.scr
 /usr/share/YaST2/scrconf/net_showexports.scr
-/usr/share/YaST2/scrconf/network.scr
+/usr/share/YaST2/scrconf/network.scr.in
 /usr/share/YaST2/scrconf/proc_cmdline.scr
 /usr/share/YaST2/scrconf/proc_cpuinfo.scr
 /usr/share/YaST2/scrconf/proc_meminfo.scr
@@ -313,7 +313,9 @@
 /usr/share/applications/YaST2/messages.desktop
 /usr/share/bash-completion/completions/yast2-completion.sh
 /usr/share/doc/packages/yast2
+/usr/share/doc/packages/yast2/CONTRIBUTING.md
 /usr/share/doc/packages/yast2/COPYING
+/usr/share/doc/packages/yast2/README.md
 /usr/share/fillup-templates/sysconfig.yast2
 /usr/share/icons
 /usr/share/icons/hicolor
```

The good news is that you can blame Oracle for that ;-) f3c2944777b3a473e04b669ab63e1cc503bf9699

library/network/test/network_interfaces_test.rb
jreidinger: @mvidner it is %check section, so it is not installed on system. Check is run before install. so it should use local https://github.com/yast/yast-yast2/blob/master/library/network/src/scrconf/network.scr.in which is probably that issue. WTF is that dynamic scr agent? I have to check how it works in past.

package/yast2.spec
```diff
@@ -139,9 +142,21 @@ mkdir -p %{buildroot}%{yast_schemadir}/control/rnc
 mkdir -p %{buildroot}%{yast_schemadir}/autoyast/rnc
 mkdir -p %{buildroot}%{_sysconfdir}/YaST2
 
+rake install DESTDIR="%{buildroot}"
+
+# symlink to call yast in ncurses
+mkdir -p %{buildroot}%{_sbindir}
+ln -snf yast2 %{buildroot}%{_sbindir}/yast
```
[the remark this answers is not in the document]
jreidinger: sadly currently not. See https://github.com/yast/yast-rake/blob/master/lib/tasks/install.rake#L81
What can be possible is to create symlink in current directory and then later install it, but we need to also clean it, which is bad. Or we can add it as symlink already to git repo? Is it acceptable?

package/yast2.spec
```diff
@@ -139,9 +142,21 @@ mkdir -p %{buildroot}%{yast_schemadir}/control/rnc
...
+# backward compatibility
+mkdir -p %{buildroot}/sbin
+ln -snf yast2 %{buildroot}/sbin/yast2
+ln -snf yast2 %{buildroot}/sbin/yast
+ln -snf yast2.8 %{buildroot}/usr/share/man/man8/yast.8
 # symlink the yardoc duplicates, saves over 2MB in installed system
 # (the RPM package size is decreased just by few kilobytes
 # because of the compression)
+# remove local .rubocop
+rm %{buildroot}/usr/share/YaST2/lib/network/.rubocop.yml
```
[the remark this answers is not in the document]
jreidinger: hmmm, sadly not, see link above. Only if I make regexp more precise. Or another option is to move these exception to global .rubocop.yml ( global in sense of /.y2rubocop in yast2 git repo ). instead of local file.

jreidinger:
@mvidner so can we just drop that oracle patch? I do not like this build magic stuff. It also means that local run is problematic, as it does not use sources from git, but from system.

mvidner:
Yes, the Oracle patch is a lone piece that was not followed by a better Fedora support.

package/yast2.spec
```diff
@@ -139,9 +142,21 @@ mkdir -p %{buildroot}%{yast_schemadir}/control/rnc
...
+rake install DESTDIR="%{buildroot}"
+
+# symlink to call yast in ncurses
+mkdir -p %{buildroot}%{_sbindir}
+ln -snf yast2 %{buildroot}%{_sbindir}/yast
+# backward compatibility
+mkdir -p %{buildroot}/sbin
+ln -snf yast2 %{buildroot}/sbin/yast2
+ln -snf yast2 %{buildroot}/sbin/yast
+ln -snf yast2.8 %{buildroot}/usr/share/man/man8/yast.8
```
mvidner: We could avoid **all** these symlinks by using application specific links:
```sh
echo 'exec /usr/sbin/yast2'         > /sbin/yast2
echo 'exec -a yast /usr/sbin/yast2' > /sbin/yast
echo 'exec -a yast /usr/sbin/yast2' > /usr/sbin/yast
```

(the `-a` argument sets argv[0] so that yast2 knows to use ncurses)

`man` can also do it (although I can't find the reference docs for it)
```sh
echo '.so yast2.8' > /usr/share/man/man8/yast.8
```

I write them as `echo` but in fact we would have them in Git and install them to their places.

jreidinger: OK, I try it and it complain ( but build pass ):

```
[   54s] yast2.x86_64: W: duplicate-executable yast ['/sbin/yast', '/usr/sbin/yast']
[   54s] yast2.x86_64: W: duplicate-executable yast2 ['/sbin/yast2', '/usr/sbin/yast2']
[   54s] This executable file exists in more than one standard binary directories. It
[   54s] can cause problems when dirs in $PATH are reordered.

[   54s] yast2.x86_64: W: script-without-shebang /sbin/yast
[   54s] yast2.x86_64: W: script-without-shebang /sbin/yast2
[   54s] yast2.x86_64: W: script-without-shebang /usr/sbin/yast
[   54s] This text file has executable bits set or is located in a path dedicated for
[   54s] executables, but lacks a shebang and cannot thus be executed.  If the file is
[   54s] meant to be an executable script, add the shebang, otherwise remove the
[   54s] executable bits or move the file elsewhere.
```

So looks like symlink is preferred over this exec stuff. BTW why you think the file is better?

mvidner: It's not really better, it just avoids symlinks, but in the end symlinks seem to be less trouble than this.

PR #933, opened 2019-05-31
Support new desktop file structure

hellcp:


library/general/src/modules/Directory.rb
```diff
@@ -147,7 +143,6 @@ def find_data_file(relative_path)
     publish variable: :moduledir, type: "string"
     publish variable: :scrconfdir, type: "string"
     publish variable: :desktopdir, type: "string"
-    publish variable: :icondir, type: "string"
```
lslezak: We cannot remove the `Directory.icondir` call, it is [still used by some YaST code](https://github.com/search?q=org%3Ayast+%22Directory.icondir%22&type=Code)!

lslezak:
BTW the correct procedure for removing `Directory.icondir` is to first mark it as deprecated (now for Leap 15.2/SP2) and remove it in the next release (Leap 15.3/SP3). So the 3rd party developers have some time to adapt their code.

library/general/src/modules/Directory.rb
```diff
@@ -86,6 +86,10 @@ def main
       # Directory for desktop files
       @desktopdir = "/usr/share/applications/YaST2"
 
+      # Base directory for icons
+      #
```
lslezak: I'd add a comment like 
```ruby
# @deprecated Use just a plain icon name without suffix, the UI will automatically find the best suitable icon.
```

(Is that actually correct?)

library/general/src/modules/Directory.rb
```diff
@@ -143,6 +147,7 @@ def find_data_file(relative_path)
     publish variable: :moduledir, type: "string"
     publish variable: :scrconfdir, type: "string"
     publish variable: :desktopdir, type: "string"
+    publish variable: :icondir, type: "string"
```
lslezak: I'd add a comment like 
```ruby
# @deprecated Use just a plain icon name without suffix,
#    the UI will automatically find the best suitable icon.
```

(Is that actually correct?)

hellcp: In most widgets, yes, not in RichText yet though

PR #935, opened 2019-06-05
Deprecate itanium

jreidinger:
Itanium is last build for SLE11 code stream, so it is no longer used and should be removed from code.


package/yast2.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Wed Jun  5 13:19:01 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
+
+- deprecate Arch.ia64 and drop all support for ia64
```
mvidner:
```suggestion
- deprecate Arch.ia64 and drop all support for ia64 (last seen in SLE 11)
```

PR #936, opened 2019-06-09
Change the way groups are defined 

hellcp:
https://specifications.freedesktop.org/menu-spec/latest/


library/desktop/yast-settings.menu
```diff
@@ -0,0 +1,85 @@
...
+			<Category>X-SuSE-YaST-Support</Category>
+		</Include>
+	</Menu>
+	<Menu>
+		<Name>Miscellaneous</Name>
+		<Directory>X-SuSE-YaST-Misc.directory</Directory>
+		<Include>
+			<Category>X-SuSE-YaST-Misc</Category>
+		</Include>
+	</Menu>
```
jreidinger, line 84: I have question about this approach. We now have some yast modules that defines own category and we are not aware about it ( at least almost not aware as it is private yast modules, just see some bug reports ). How they can add new categories to yast menus?

jreidinger:
changed it back to draft as change to ycc still missing. See https://github.com/yast/yast-yast2/issues/937

PR #938, opened 2019-06-13
PackageSystem.Installed: Improve loging details

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1137992#c8

### DIFF
--- a/library/packages/src/modules/PackageSystem.rb
+++ b/library/packages/src/modules/PackageSystem.rb
@@ -348,9 +348,8 @@ module Yast
       # so we must avoid it.
       # added --whatprovides due to bug #76181
       rpm_command = "/usr/bin/rpm -q --whatprovides #{package.shellescape}"
-      log.debug("Starting RPM query: #{rpm_command}")
       output = SCR.Execute(path(".target.bash_output"), rpm_command)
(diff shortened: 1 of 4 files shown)


library/packages/src/modules/PackageSystem.rb
```diff
@@ -335,22 +337,23 @@ def InitRPMQueryBinary
...
-        Convert.to_integer(
-          SCR.Execute(
-            path(".target.bash"),
-            Ops.add("/usr/bin/rpm -q --whatprovides ", package.shellescape)
-          )
-        )
+      rpm_command = "/usr/bin/rpm -q --whatprovides #{package.shellescape}"
+      log.debug("Starting RPM query: #{rpm_command}")
+      output = SCR.Execute(path(".target.bash_output"), rpm_command)
+      log.debug("result of the query: #{output}")
```
jreidinger: well, I would make it non-debug ( info ) and also compress in into one line (rpm -q does not take so long and it is already logged in debug level) like:
```
log.info "Query installed package with '#{rpm_command}' and result #{output.inspect}"
```

teclator: I was not sure about loging always.

PR #942, opened 2019-06-19
PackageSystem.Installed: Use Yast::Execute instead of SCR

teclator:
Use Yast::Execute as suggested in https://bugzilla.suse.com/show_bug.cgi?id=1137992#c16

### DIFF
--- a/library/packages/src/modules/PackageSystem.rb
+++ b/library/packages/src/modules/PackageSystem.rb
@@ -348,12 +348,15 @@ module Yast
       # Unfortunately, initializing Pkg reads the RPM database...
       # so we must avoid it.
       # added --whatprovides due to bug #76181
-      rpm_command = "/usr/bin/rpm -q --whatprovides #{package.shellescape}"
-      output = Yast::Execute.on_target.stdout(rpm_command.split(" "))
-      log.info "Query installed package with '#{rpm_command}' and result #{output.inspect}"
+      # Use Yast::Execute to prevent false positives (boo#1137992)
+      rpm_command = ["/usr/bin/rpm", "-q", "--whatprovides", package.shellescape]
+      # We are not raising exceptions in case of return codes different than
+      # 0 or 1. So do not plan to modify the current behavior.
+      output, return_code = Yast::Execute.stdout.on_target!(rpm_command, allow_exitstatus: 0..1)
+      log.info "Query installed package with '#{rpm_command.join(" ")}' and result #{output}"
 
       # return Pkg::IsProvided (package);
-      !!output
+      return_code == 0
     end
 
     # Is a package installed? Checks only the package name in contrast to Installed() function.
--- a/library/packages/src/modules/PackageSystem.rb
+++ b/library/packages/src/modules/PackageSystem.rb
@@ -349,7 +349,7 @@ module Yast
       # so we must avoid it.
       # added --whatprovides due to bug #76181
       # Use Yast::Execute to prevent false positives (boo#1137992)
-      rpm_command = ["/usr/bin/rpm", "-q", "--whatprovides", package.shellescape]
+      rpm_command = ["/usr/bin/rpm", "-q", "--whatprovides", package]
       # We are not raising exceptions in case of return codes different than
       # 0 or 1. So do not plan to modify the current behavior.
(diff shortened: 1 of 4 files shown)


library/packages/src/modules/PackageSystem.rb
```diff
@@ -348,11 +349,11 @@ def Installed(package)
       # so we must avoid it.
       # added --whatprovides due to bug #76181
       rpm_command = "/usr/bin/rpm -q --whatprovides #{package.shellescape}"
-      output = SCR.Execute(path(".target.bash_output"), rpm_command)
+      output = Yast::Execute.on_target.stdout(rpm_command.split(" "))
```
jreidinger: that split is very bad idea. As it kills shellescape and e.g. "pkg\ a" is then splitted. So please adapt rpm command above to array.

jreidinger: and also it will popup if rpm failed, so not found package. That is wrong.

teclator: Yep, should use on_target! instead

teclator: Something like:

```ruby
      output = Yast::Execute.stdout.on_target!(rpm_command)
      log.info "Query installed package with '#{rpm_command.join(" ")}' and result #{output}"

      # return Pkg::IsProvided (package);
      !output.empty?
```

Should be good enough, do you have another suggestion?

jreidinger: no, you need to specify allowed_exitstatus to avoid exception when rpm is not found as it is common case.

teclator: But when no found it is empty as force_stdout is used

library/packages/src/modules/PackageSystem.rb
```diff
@@ -347,12 +348,15 @@ def Installed(package)
       # Unfortunately, initializing Pkg reads the RPM database...
       # so we must avoid it.
       # added --whatprovides due to bug #76181
-      rpm_command = "/usr/bin/rpm -q --whatprovides #{package.shellescape}"
-      output = SCR.Execute(path(".target.bash_output"), rpm_command)
-      log.info "Query installed package with '#{rpm_command}' and result #{output.inspect}"
+      # Use Yast::Execute to prevent false positives (boo#1137992)
+      rpm_command = ["/usr/bin/rpm", "-q", "--whatprovides", package.shellescape]
```
jreidinger: here shellescape is wrong as cheetah does not use shell for command. So wrong package name will be passed.

library/packages/src/modules/PackageSystem.rb
```diff
@@ -33,6 +33,7 @@
 # The documentation is maintained at
 # <a href="../index.html">.../docs/index.html</a>.
 require "yast"
+require "yast2/execute"
 require "shellwords"
```
jreidinger: should no longer be needed if it is used only in that one instance for shellescape

teclator: Well, it is used also in `PackageInstalled` method. So we could consider adapt it to use Yast::Execute in case we want to completely remove the 'shellwords' requirement

library/packages/src/modules/PackageSystem.rb
```diff
@@ -347,12 +348,15 @@ def Installed(package)
...
       # so we must avoid it.
       # added --whatprovides due to bug #76181
-      rpm_command = "/usr/bin/rpm -q --whatprovides #{package.shellescape}"
-      output = SCR.Execute(path(".target.bash_output"), rpm_command)
-      log.info "Query installed package with '#{rpm_command}' and result #{output.inspect}"
+      # Use Yast::Execute to prevent false positives (boo#1137992)
+      rpm_command = ["/usr/bin/rpm", "-q", "--whatprovides", package]
+      # We are not raising exceptions in case of return codes different than
+      # 0 or 1. So do not plan to modify the current behavior.
+      output, return_code = Yast::Execute.stdout.on_target!(rpm_command, allow_exitstatus: 0..1)
```
mvidner: @teclator Luckily I was reviewing a Cheetah PR yesterday so I know that the correct name of the option is `allowed_exitstatus`

teclator: Yep :(, It was already fixed in https://github.com/yast/yast-yast2/pull/944

PR #946, opened 2019-06-25
New rubocop

jreidinger:
Initial pull request with the rubocop version 0.71.0 that include shared rubocop rules for whole yast in `rubocop_yast_style.yml` and module specific in `.rubocop.yml`.
Changes are done per commit. Commits with NO-AUTO in name is done manually.
Travis is not yet adapted to run newer version of rubocop.

# TODO

- [x] Package the new Rubocop as RPM in [YaST:Head](https://build.opensuse.org/project/show/YaST:Head), make sure both (the old and the new one) can be installed in parallel (maybe the old package will need to be adapted to this...)
- [x] Share the new Rubocop configuration file
- [x] Ensure that the right Rubocop is used when running the `rake check:rubocop` task, make sure the old Rubocop is still used in the old (maintenance) branches


.rubocop.yml
```diff
@@ -83,11 +90,21 @@ Style/TrivialAccessors:
...
   Include:
     - "library/*/src/lib/**/*.rb"
   Exclude:
     - "library/network/src/lib/network/susefirewall.rb"
     - "library/network/src/lib/network/susefirewalld.rb"
     - "library/network/src/lib/network/susefirewall2.rb"
     - "library/network/src/lib/network/susefirewall2services.rb"
+
+# Some tests and examples mixin to global, where it is not so harmful
+# (not so sure about tests as it modify global namespace for multiple tests in paralel )
```
mvidner: We should fix this, simply by moving the `include` to a `describe` block. My own HackWeek project finds it a blocker for further checking.

PR #947, opened 2019-07-01
DnsServerAPI: skip 'bind' absence in mode test

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1138668

## Problem
- Failing old testsuite in yast2-dns-server
- It depends on bind rpm installed.

## Fix

- Do not depend on the package installed at all in Mode.test()
- 4.2.11

### DIFF
--- a/library/network/src/modules/DnsServerAPI.pm
+++ b/library/network/src/modules/DnsServerAPI.pm
@@ -61,6 +61,7 @@ YaST::YCP::Import ("Report");
 YaST::YCP::Import ("IP");
 # for syntax checking
 YaST::YCP::Import ("Hostname");
+YaST::YCP::Import ("Mode");
 
 our %TYPEINFO;
 my $package_installed = -1;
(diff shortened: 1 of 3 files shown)


library/network/src/modules/DnsServerAPI.pm
```diff
@@ -654,6 +655,11 @@ sub UnquoteString {
 }
 
 sub Init {
+    if (Mode->test ()) {
```
imobachgs: I must admit that I do not like this approach, although I am not sure if we have any other posibility here. If it is not possible to mock the calls to the bind rpm, I guess it is good enough.

teclator: In the past we were just mocking the bash_output return, but now it is harder to mock the call to Yast2::Execute.

Following same approach done in https://github.com/yast/yast-users/pull/209 && https://github.com/yast/yast-squid/pull/28

lslezak: Unfortunately the old testsuite can only mock the `SCR` calls, it cannot mock the Ruby [Yast::Execute](https://github.com/yast/yast-yast2/blob/master/library/packages/src/modules/PackageSystem.rb#L350-L353). Moreover this is Perl code...

This is the only way for the old tests... :worried:

imobachgs: Ok, then I am fine with the changes :) Thanks!

PR #948, opened 2019-07-03
Fix desktop modules reading

imobachgs:
Fixes https://bugzilla.opensuse.org/show_bug.cgi?id=1140233.

### DIFF
--- a/library/desktop/src/modules/Desktop.rb
+++ b/library/desktop/src/modules/Desktop.rb
@@ -180,7 +180,7 @@ module Yast
           ret = ReadLocalizedKey(filename, filepath, value)
           Ops.set(filemap, value, ret) if !ret.nil? && ret != ""
         end
-        name2 = Builtins.regexpsub(file, "^.*/(.*).desktop", "\\1")
+        name2 = desktop_file_to_module(file)
         if name2 != "" && !name2.nil?
           Ops.set(@Modules, name2, filemap)
           group = Ops.get_string(filemap, "X-SuSE-YaST-Group", "")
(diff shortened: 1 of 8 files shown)


library/desktop/src/modules/Desktop.rb
```diff
@@ -400,6 +400,27 @@ def ParseSingleDesktopFile(file)
...
+    # In 2019, desktop files were renamed to org.opensuse.yast.MODULE.desktop. This method extracts
+    # and normalizes the *MODULE* part of the filename.
+    #
+    # @param [String] filename
+    # @return [String,nil] Module name or nil if it cannot be inferred.
+    # @see https://github.com/yast/yast-yast2/issues/934
+    def desktop_file_to_module(filename)
+      basename = File.basename(filename)
+      name = basename[FILE_REGEXP, 1]
+      return if name.nil?
```
jreidinger: logging that file is not desktop?

library/desktop/src/modules/Desktop.rb
```diff
@@ -400,6 +400,27 @@ def ParseSingleDesktopFile(file)
...
+
+      name.gsub(/([[:lower:]])([[:upper:]]+)/, '\1-\2').downcase
```
jreidinger: can you add to comment example what will change to what?

library/desktop/test/desktop_test.rb
```diff
@@ -0,0 +1,40 @@
...
+    before do
+      Yast::Desktop.Read(["Name"])
+    end
+
+    it "returns modules information using the short name as key" do
+      expect(Yast::Desktop.Modules).to eq(
+        "add-on"           => { "Name" => "YaST Add-On Products" },
+        "lan"              => { "Name" => "YaST Network" },
+        "services-manager" => { "Name" => "YaST Services Manager" }
+      )
```
jreidinger: :+1: for tests.

PR #957, opened 2019-08-30
yast completions have to be named after their respective command name…

wfeldt:
## Problem

https://bugzilla.suse.com/show_bug.cgi?id=1122259

yast's bash completion does not work.

## See also

This is a backport of https://github.com/yast/yast-yast2/pull/955 for SLE-15-SP1.

Note: `autotools` instead of `rake`.


ancorgs:
So this was fixed first for master and then backported to SLE-15-SP1 on the same day. Any reason for not following the standard procedure at [1]? That is, fixing first the older branch and then merging forward.

Moreover, if this was done via backport, I miss the merge from SP1 to master described in that document when such thing is done instead of the standard procedure (subsection "Backporting fix").

[1] https://yastgithubio.readthedocs.io/en/latest/maintenance-branches/#maintenance-work-flow

PR #962, opened 2019-09-19
Added support for reading products from control.xml file

lslezak:
- Added support for reading the products from the `control.xml` file
- Needed for a repo-less online installation medium, that will not contain any repository (so no product there)
- https://jira.suse.com/browse/SLE-7104
- To share some parts I have moved some code to the `ProductLicenseMixin` module
- This code will be used by some other modules, it was added here into the base YaST package to avoid some circular dependencies between `yast2-packager` and `yast2-installation`.

The base product definition in the `control.xml` file would look like:

```xml
<base_products config:type="list">
  <base_product>
    <label>SUSE Linux Enterprise Server 15 SP2</label>
    <name>SLES</name>
    <version>15.2</version>
    <register_target>sle-15-$arch</register_target>
  </base_product>
</base_products>
```

Optionally a product might be limited to a specific architecture (`archs` key) or can be hidden by default (`special_product` key) like the "SLES BCL" special product.

### DIFF
--- a/library/packages/src/lib/y2packager/product_control_product.rb
+++ b/library/packages/src/lib/y2packager/product_control_product.rb
@@ -1,0 +1,102 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2019 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, contact SUSE.
+#
+# To contact SUSE about this file by physical or electronic mail, you may find
+# current contact information at www.suse.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "y2packager/product_license_mixin"
+
+Yast.import "Arch"
+Yast.import "Linuxrc"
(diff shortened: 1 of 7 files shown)


library/packages/src/lib/y2packager/product_control_product.rb
```diff
@@ -0,0 +1,102 @@
...
+
+Yast.import "Arch"
+Yast.import "Linuxrc"
+Yast.import "ProductFeatures"
+
+module Y2Packager
+  # This class implements a base product read from the control.xml file.
+  class ProductControlProduct
+    # initialize the selected base product
+    @selected = nil
```
jreidinger: is it needed? it should be nil by default, not?

library/packages/src/lib/y2packager/product_control_product.rb
```diff
@@ -0,0 +1,102 @@
...
+    include ProductLicenseMixin
+
+    attr_reader :name, :version, :arch, :label, :license_url, :register_target
+
+    #
+    # Read the base products from the control.xml file. The products for the incompatible
+    # machine architecture and the hidden products are filtered out.
+    #
+    # @return [Array<Installation::ProductControlProduct>] List of the products
+    def self.products
```
jreidinger: why not having this code in above `self << class` ?

library/packages/src/lib/y2packager/product_control_product.rb
lslezak: Ah, I originally used `@@` in the code and Ruby complained about missing initialization, after changing it that's not needed anymore.

library/packages/src/lib/y2packager/product_control_product.rb
lslezak: That was actually added/changed by Rubocop, I'll add it there as well.

PR #965, opened 2019-10-03
Add documentation about future /etc and /usr/etc support

imobachgs:


doc/etc-and-usr-etc.md
```diff
@@ -0,0 +1,75 @@
...
+  codebase (search for `.scr`) files.
+
+### Extending CFA
+
+CFA offers an object-oriented way to read and write configuration files and, nowadays, it is used
+for `zypp.conf`, `chrony.conf`, `ifroute-*` and `ifcfg-*` files, etc. CFA is built around these abstractions:
+
+* **File handlers* provide access to files. By default, it simply uses the `File` class, but it can be replaced
+  with other mechanisms to allow, e.g., accessing over the network. See the discussion about supporting agents
+  to find another use case.
```
jreidinger: I would mention here the main extension done in YaST which is support for configuration in chroot - https://github.com/yast/yast-yast2/blob/master/library/system/src/lib/yast2/target_file.rb#L51

doc/etc-and-usr-etc.md
```diff
@@ -0,0 +1,75 @@
...
+* *Parsers* analyze and extract configuration from files. Usually, CFA parsers use Augeas under the
+  hood.
+* *Models* offer an object-oriented API to a access a configuration file.
+
+Usually, a model is meant to represent a configuration file, but in a layout where the configuration
+is spread through several files, this approach could be pretty inconvenient. So, in order to support
+the new layout, we are introducing a new layer that abstracts the details of merging and building
+the model instance.
+
+The new `loader` classes offer and API to `#load` and `#save` the configuration files. On the one
```
jreidinger: typo `and` instead of `an`

doc/etc-and-usr-etc.md
```diff
@@ -0,0 +1,75 @@
...
+hand, the default loader (`Loader` class) would read the information from a single file. On the
+other hand, an alternative loader (`VendorLoader`) would build the configuration model by reading
+vendor and user settings.
+
+When it comes to writing the changes, the `VendorLoader` class will write changes to the `.d`
+directory if it exists.
```
jreidinger: under which name it will write to that .d directory?

imobachgs: For the time being, `50-yast.conf`, but it should be configurable.

doc/etc-and-usr-etc.md
```diff
@@ -0,0 +1,75 @@
...
+
+Agents are used through all YaST code to read/write configuration settings. In order to support the
+new layout, we could follow (at least) these approaches:
+
+* Extend [agents](https://github.com/yast/yast-core/) to support reading/writing from/to different
+  files. At least `any`, `ini` and `modules` would need to be adapted.
+* Extend CFA to support reading/writing information from agents. It can be done by creating a
+  specific *file handler* for agents. See [yast/yast-yast2 usr-etc-support
+  branch](https://github.com/yast/yast-yast2/compare/usr-etc-support?expand=1) for a
+  proof-of-concept.
```
jreidinger: maybe add third option to not change agent and instead add agents for other locations and then merge it in ruby code that works with agent. It can be useful if merging rules are tricky.

imobachgs: That's basically the idea behind the CFA-based implementation, as we register agents as needed on demand.

PR #966, opened 2019-10-03
Add a CFA class to handle sysctl configuration

imobachgs:
This PR adds an small class to handle `sysctl` settings:

* Values are now written to `/etc/sysctl.d/50-yast.conf`.
* Old values are imported from `/etc/sysctl.conf`.
* Old values are removed from `/etc/sysctl.conf`.

## Packages to adapt

- [x] yast2-network (https://github.com/yast/yast-network/pull/991)
- [x] yast2-security (https://github.com/yast/yast-security/pull/59)
- [x] yast2-vpn (https://github.com/yast/yast-vpn/pull/15)
- [x] yast2-tune (https://github.com/yast/yast-tune/pull/39)


doc/etc-and-usr-etc.md
```diff
@@ -27,6 +27,24 @@ When it comes to reading or writing configuration files, YaST uses mainly two di
...
+## A Simple Solution for `sysctl.conf`
+
+In a nutshell, jsc#SLE-9077 states that `/etc/sysctl.conf` should not be modified.  So if you want
+to modify any `sysctl` setting, you should drop a file in `/etc/sysctl.d` containing the new values.
+
+As a first step, we have added a {Yast2::CFA::Sysctl} class which offers an API to sysctl settings.
+This new class uses `/etc/sysctl.d/50-yast.conf` instead of `/etc/sysctl.conf` to write the configuration.
+Moreover, it removes known keys from the original `/etc/sysctl.conf` to avoid confusion.
+
+The main limitation of this approach is that YaST will only read a set of known keys but, as it uses
```
jreidinger: I do not get why it is limitation. It is in fact whole idea of CFA to read/modify only parts that yast need.

library/general/src/lib/yast2/cfa/sysctl.rb
```diff
@@ -0,0 +1,154 @@
...
+          add_key(key)
+
+          define_method meth do
+            self[key]
+          end
+
+          define_method "#{meth}=" do |value|
+            self[key] = value
+          end
+        end
```
jreidinger: well, what about just redefine attributes method from base model to also remember defined attributes? https://github.com/config-files-api/config_files_api/blob/master/lib/cfa/base_model.rb#L98

library/general/src/lib/yast2/cfa/sysctl.rb
```diff
@@ -0,0 +1,154 @@
...
+        data[key]
+      end
+
+      # Sets the value for a given key
+      #
+      # @param key   [String] Name of the key to set the value
+      # @param value [Object] New value
+      def []=(key, value)
+        data[key] = value
+      end
```
jreidinger: if this method is needed, I would mark it private to use accessor instead.

library/general/src/lib/yast2/cfa/sysctl.rb
```diff
@@ -0,0 +1,154 @@
...
+
+    private
+
+      # SCR paths IPv4 / IPv6 Forwarding
+      SYSCTL_AGENT_PATH = ".etc.sysctl_conf".freeze
+
+      # Remove values from `/etc/sysctl.conf` to reduce noise and confusion
+      def clean_old_values
+        self.class.known_keys.each do |key|
+          Yast::SCR.Write(key_path(key), nil)
```
jreidinger: well, what code simplify is if you have path as constant and here only append - 
so
```
SYSCTL_AGENT_PATH = Yast::Path.new(".etc.sysctl_conf")
```

and here
```
Yast::SCR.Write(SYSCTL_AGENT_PATH + key, nil)
```

see https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/path.rb#L31

Path is object and contain a bit of smartness :)

library/general/src/lib/yast2/cfa/sysctl.rb
imobachgs: The only purpose of this module was to open a "backdoor" to set any sysctl value. But after all the keys/attributes handling, now it looks like a bad idea. Removed.

library/general/src/lib/cfa/sysctl.rb
```diff
@@ -0,0 +1,123 @@
...
+  #   sysctl.load
+  #   sysctl["net.ipv4.ip_forward"] = "1"
+  #   sysctl.save
+  #
+  # @example Setting a value using an accessor
+  #   sysctl = Sysctl.new
+  #   sysctl.forward_ipv4 = "1"
+  #   sysctl.save
+  class Sysctl < ::CFA::BaseModel
+    PARSER = ::CFA::AugeasParser.new("sysctl.lns")
```
jreidinger: those CFA prefixes are no longer needed as it is now in CFA namespace.

imobachgs:
Finally I decided to do not remove the old values from `/etc/sysctl.conf` because it may happen that the file gets emptied (losing the comment that explains the difference between `/etc/sysctl.conf` and `/etc/sysctl.d`).

library/general/src/lib/cfa/sysctl.rb
```diff
@@ -0,0 +1,155 @@
...
+  class Sysctl < BaseModel
+    PARSER = AugeasParser.new("sysctl.lns")
+    PATH = "/etc/sysctl.d/50-yast.conf".freeze
+
+    class << self
+      # Modifies default CFA methods to handle boolean values
+      #
+      # When getting or setting the value, a boolean value will be expected. Under the hood, it will
+      # be translated into "1" or "0". Additionally, to access to the raw value ("1", "0" or +nil+),
+      # just prepend "raw_" to the name of the method.
```
jreidinger: maybe write there that if value is not defined it is "false".

PR #968, opened 2019-10-08
Remove old sysctl.conf values

imobachgs:
In #967 we decided to simply update old values in `/etc/sysctl.conf` because the `.sysctl_conf` agent could empty the file when trying to remove the last entry. To work around this problem, this change simply uses a regular expression to drop the unwanted lines, keeping the comments.

### DIFF
--- a/library/general/src/lib/cfa/sysctl.rb
+++ b/library/general/src/lib/cfa/sysctl.rb
@@ -17,6 +17,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "yast"
 require "cfa/base_model"
 require "yast2/target_file"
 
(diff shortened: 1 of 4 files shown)


library/general/src/lib/cfa/sysctl.rb
```diff
@@ -118,39 +124,27 @@ def load
...
-
-    # Returns the list of known attributes
-    #
-    # Just a helper method to get the list of defined attributes
-    #
-    # @return [String<Symbol>]
-    def known_keys
-      @known_keys ||= ATTRIBUTES.values.uniq
+    MAIN_SYSCTL_CONF_PATH = "/etc/sysctl.conf".freeze
+    KNOWN_KEYS_REGEXP = /^(#{KNOWN_KEYS.join("|")})/.freeze
```
jreidinger: I think if we use regexp approach we need to ensure it does not fetch e.g. partial keys like net.ipv4.conf.all and net.ipv4.conf.all.forwarding. Another issue here is special regexp characters like `.` in keys which can result in unwanted match. In general using regexp for manipulating config files where we are not so sure how it can look like is dangerous (e.g. can be whitespace between start of line and key? ).
If you want cleanly remove some key I think that CFA approach is better. Just load directly that file and remove given key from it. It will keep comments.

PR #972, opened 2019-10-21
NetworkService: Check if systemd is running for determining the network backend

teclator:
## Problem

In general, YaST assumes wicked as the default network service for any kind of installation. But that is not the case for a live installation.

In a live installation systemd is already running and we can determine the backend checking the network.service link.

- https://trello.com/c/qvY7Oygp/1375-3-osdistribution-p1-1151291-build-1330-openqa-test-fails-in-awaitinstall-yast-breaks-network-during-install

## Solution

Check whether `systemd` is running or not for determining the current network backend.

### DIFF
--- a/library/network/src/modules/NetworkService.rb
+++ b/library/network/src/modules/NetworkService.rb
@@ -187,18 +187,30 @@ module Yast
       Read()
     end
 
+    # Determines which backend is in use based on the network service. In an
+    # (auto)installation it returns the default backend except if systemd is
+    # running (live) where the systemd service can be checked.
+    #
+    # @return [Symbol,nil] backend in use or nil
+    def backend_in_use
+      backend = nil
+
+      if Stage.initial && !Systemd.Running
+        backend = DEFAULT_BACKEND
+        log.info "Running in installer/AutoYaST, use default: #{backend}"
+      else
+        service = Yast2::Systemd::Service.find("network")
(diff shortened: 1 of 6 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Mon Oct 21 06:59:33 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
+- Network: During an installation, check which backend is in use
+  when Systemd is running. (bsc#1151291)
+
```
teclator:
```suggestion
```

library/network/src/modules/NetworkService.rb
```diff
@@ -190,7 +191,7 @@ def disable
     def Read
       return if @initialized
 
-      if Stage.initial
+      if Stage.initial && !Systemd.Running
```
jreidinger: I expect here comment that in live system it can run.

library/systemd/test/yast2/systemd_unit_test.rb
```diff
@@ -71,32 +72,24 @@ def trigger_reloading_properties(command)
     end
 
     describe "#properties" do
+      let(:properties_methods) do
```
teclator: Just adapted for reducing the number of lines (reported by rubocop)

PR #973, opened 2019-10-24
read products from control file in online medium

jreidinger:
related pr https://github.com/yast/yast-registration/pull/454

### DIFF
--- a/library/packages/src/lib/y2packager/product_reader.rb
+++ b/library/packages/src/lib/y2packager/product_reader.rb
@@ -72,49 +72,49 @@ module Y2Packager
     # Available products
     #
     # @return [Array<Product>] Available products
-    def all_products
+    def all_products(force_repos: false)
       linuxrc_special_products = if Yast::Linuxrc.InstallInf("specialproduct")
         linuxrc_string(Yast::Linuxrc.InstallInf("specialproduct")).split(",")
       else
         []
       end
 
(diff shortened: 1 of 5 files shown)


library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -78,30 +79,45 @@ def all_products
...
+      @all_products = []
 
-          # Evaluating display order
-          prod_pkg["deps"].find { |dep| dep["provides"] =~ /\Adisplayorder\(\s*([0-9]+)\s*\)\z/ }
-          displayorder = Regexp.last_match[1].to_i if Regexp.last_match
+      if Yast::Stage.initial && Y2Packager::MediumType.online?
+        Y2Packager::ProductControlProduct.products.each do |p|
+          @all_products << Y2Packager::Product.new(name: p.name, display_name: p.label,
+            version: p.version,
+            installation_package: "Test" # just hack as we do not know current package name yet
```
jreidinger: sadly, this does not work, as it is needed later when trying to merge workflows and package is needed.
In general problem is that until we migrate, we do not have base products and after that step we need to load it from real repo from scc, but at many places there is caching.

lslezak: NP: maybe something like "Not known yet" could be better here. But it does not matter much...

PR #974, opened 2019-10-25
Use snapper machine-readable output

joseivanlopez:
## Problem

Class `FsSnapshot` is used to retrieve filesystem snapshots information. To achieve that, some *snapper* cli commands are directly used. Till now, *snapper* commands only generated a human-readable output. That output might change when some new information is displayed. And such changes might ruin scripts relying on the *snapper* cli output.

But now, *snapper* is also offering a machine-readable output that should be used by scripts. 

* Part of https://trello.com/c/YAhIvrHz/417-5-optional-non-default-machine-readable-output-from-snapper.

* Requires https://github.com/openSUSE/snapper/pull/494

## Solution

Class `FsSnapshot` is adapted to use *snapper* machine-readable output (*CSV* output format in this case).

## Tests

* Unit tests have been adapted.

### DIFF
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -94,11 +94,13 @@ module Yast2
     #
     # @param number          [Fixnum]        Snapshot's number.
     # @param snapshot_type   [Symbol]        Snapshot's type: :pre, :post or :single.
-    # @param previous_number [Fixnum, nil]   Previous snapshot's number.
-    # @param timestamp       [DateTime, nil] Timestamp
-    # @param user            [String, nil]   Snapshot's owner username.
-    # @param cleanup_algo    [Symbol, nil]   Clean-up algorithm.
-    # @param description     [String, nil]   Snapshot's description.
+    # @param previous_number [Fixnum, nil]   Previous snapshot's number; nil if the snapshot has no pre
+    #                                        snapshot associated to it.
+    # @param timestamp       [DateTime, nil] Timestamp; nil if the datetime is unknown.
+    # @param user            [String, nil]   Snapshot's owner username; nil if the owner is unknown.
+    # @param cleanup_algo    [Symbol, nil]   Clean-up algorithm; nil if the algorithm is unknown.
+    # @param description     [String, nil]   Snapshot's description; nil if the snapshot has no
+    #                                        description.
     # @return [FsSnapshot] New FsSnapshot object.
     def initialize(number, snapshot_type, previous_number, timestamp, user, cleanup_algo, description)
       @number = number
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -274,6 +274,11 @@ module Yast2
 
           fields = row.fields
 
+          if !fields[3].is_a?(DateTime)
+            log.warn("Error when parsing date/time: #{fields[3]}")
+            fields[3] = nil
+          end
+
           fields[1] = fields[1].to_sym # type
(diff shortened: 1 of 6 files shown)


joseivanlopez:
> The PR looks good to me. But just a question, what will happen if the system does not have the proper Snapper version, i.e., > 0.8.6? Will it crash?

That was my last question on IRC last Friday. I don't see any snapper dependency in spec file. Should we add it?

dgdavid:
> That was my last question on IRC last Friday.

I didn't see it :)

>  I don't see any snapper dependency in spec file. 

Exactly.

> Should we add it?

I think so, although `yast2` does not strictly depend on snapper. Let's ask for other POVs.

dgdavid:
As discussed in IRC, adding such dependency is not necessary and wrong since

> a system can be installed without snapshots

and

> the installation runs from the ISO and there we control the versions.

Even more, the [`#configured?`](https://github.com/yast/yast-yast2/pull/974/files#diff-42c17994232fca0fc845aa14c1fc9d6aR130-R140) method is already checking the exit code from the snapper command, and [it will not crash](https://github.com/yast/yast-yast2/pull/974/files#diff-42c17994232fca0fc845aa14c1fc9d6aR259) if it is not configured... or using an old Snapper version.

Thanks so much @aschnell for the help :)

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -88,11 +94,11 @@ class FsSnapshot
...
-    # @param previous_number [Fixnum]        Previous snapshot's number.
-    # @param timestamp       [DateTime]      Timestamp
-    # @param user            [String]        Snapshot's owner username.
-    # @param cleanup_algo    [String]        Clean-up algorithm.
-    # @param description     [String]        Snapshot's description.
+    # @param previous_number [Fixnum, nil]   Previous snapshot's number.
+    # @param timestamp       [DateTime, nil] Timestamp
+    # @param user            [String, nil]   Snapshot's owner username.
+    # @param cleanup_algo    [Symbol, nil]   Clean-up algorithm.
+    # @param description     [String, nil]   Snapshot's description.
```
jreidinger: please try to document what means if nil is passed as it is special case, so it deserve explanation.

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -256,21 +262,20 @@ def all
...
-            timestamp = DateTime.parse(data[3])
-          rescue ArgumentError
-            log.warn("Error when parsing date/time: #{timestamp}")
-            timestamp = nil
-          end
-          previous_number = (data[2] == "") ? nil : data[2].to_i
-          snapshots << new(data[0].to_i, data[1].to_sym, previous_number, timestamp,
-            data[4], data[5].to_sym, data[6])
+
+        csv = CSV.parse(out["stdout"], headers: true, converters: [:date_time, :numeric])
```
jreidinger: what about errors? E.g. old code handle data_time conversion error, what happen here if something is malformed?

package/yast2.spec
```diff
@@ -119,6 +119,8 @@ Conflicts:      yast2-installation < 4.2.9
 Conflicts:      yast2-mail < 3.1.7
 # Older packager use removed API
 Conflicts:      yast2-packager < 4.0.33
+# Older snapper does not provide machine-readable output
+Conflicts:	snapper < 0.8.6
```
jreidinger: :+1:

library/system/src/lib/yast2/fs_snapshot.rb
joseivanlopez: Oh yes. That check is missing now. Thanks!

PR #975, opened 2019-10-29
Short name control product

jreidinger:


library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -85,6 +85,7 @@ def all_products(force_repos: false)
         return Y2Packager::ProductControlProduct.products.each_with_object([]) do |p, result|
           result << Y2Packager::Product.new(name: p.name, display_name: p.label,
               version: p.version, arch: p.arch,
+              short_name: p.name, # TODO: verify that name and shortname are same for control products
```
teclator: @jreidinger , @lslezak a TODO could be ignored, shall we create a PBI for it?, Why the short name is not part of the control product definition? Even a comment in the control file could help

lslezak: To make the maintenance easier I want to keep only minimal set of data in `control.xml`. I omitted the short name because it is not needed for registration. Of course, if we need more data we can add it there.

For the current products actually `name` != `short_name`, e.g. for SLE15-SP2 it's `SLES` vs. `SLES15-SP2`. I do not know if this could make some problems, but maybe it's OK in the early stage before we add the real repositories with real products...

PR #976, opened 2019-10-31
Decreasing required memory

schubi2:
Replacing old Pkg.ResolvableProperties()and Pkg.ResolvabeDependencies() calls by Y2Packager::Resolvable.any? and Y2Packager::Resolvable.find in order to save memory.


library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -145,21 +146,31 @@ def installed_base_product
       base = base_product
       return nil unless base
 
-      Y2Packager::Product.from_h(base)
+      Y2Packager::Product.new(
```
imobachgs: You could have implemented a `Y2Packager::Product.from_resolvable` method, similar to `from_h`. That method should receive an additional `installation_package` argument too, which is not included in the resolvable.

library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -184,14 +195,14 @@ def zypp_products
     # @return [Array<Hash>] pkg-bindings data structures
     def available_products
       # select only the available or to be installed products
-      zypp_products.select { |p| p["status"] == :available || p["status"] == :selected }
+      zypp_products.select { |p| p.status == :available || p.status == :selected }
     end
 
     # read the installed products
     # @return [Array<Hash>] pkg-bindings data structures
```
imobachgs: Please, adjust the `@return` class (`Array<Y2Packager::Resolvable>`).

library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -168,14 +179,14 @@ def product_package(name, _repo_id = nil)
     # read the available products, remove potential duplicates
     # @return [Array<Hash>] pkg-bindings data structure
```
imobachgs: Please, adjust the `@return` class (`Array<Y2Packager::Resolvable>`). Actually, that makes me think whether we could transform these resolvable into `Product` instances at this point (or if we are going to miss some information later).

If that's not the case, I am not sure about keeping the `*_products` suffix when we are working with resolvables. It is confusing.

library/packages/src/lib/y2packager/product_reader.rb
```diff
@@ -145,21 +146,31 @@ def installed_base_product
...
+      )
     end
 
     # All installed products
     # @return [Array<Y2Packager::Product>] the product list
     def all_installed_products
-      installed_products.map { |p| Y2Packager::Product.from_h(p) }
+      installed_products.map do |p|
+        Y2Packager::Product.new(
+          name: p.name, short_name: p.short_name, display_name: p.display_name,
```
imobachgs: A `from_resolvable` method would save you from repeating this code :)

library/packages/src/modules/Product.rb
```diff
@@ -211,11 +224,11 @@ def fill_up_relnotes(products)
...
 
       products.map do |p|
-        next if p["relnotes_url"] == ""
+        next if p.relnotes_url == ""
 
-        url = p["relnotes_url"]
+        url = p.relnotes_url
         all_release_notes << url
-        release_notes_to_product[url] = (p["display_name"] || "")
+        release_notes_to_product[url] = (p.display_name || "")
```
imobachgs: You do not need the parenthesis.

package/yast2.spec
```diff
@@ -52,6 +52,7 @@ BuildRequires:  yast2-core >= 2.18.12
 BuildRequires:  yast2-devtools >= 3.1.10
 # Pkg.Resolvables()
 BuildRequires:  yast2-pkg-bindings >= 4.2.0
+BuildRequires:  yast2-packager
```
imobachgs: Warning: circular dependency: https://github.com/yast/yast-packager/blob/master/package/yast2-packager.spec#L39. I think you should omit it or move the needed classes (from `Y2Packager`) to this repository.

library/packages/src/lib/y2packager/product.rb
```diff
@@ -65,6 +65,22 @@ def from_h(product)
         Y2Packager::Product.new(params)
       end
 
+      # Create a product from Y2Packager::Resolvable
+      # @param product [Y2Packager::Resolvable] product
+      # @param installation_package [String] installation package name
+      # @param displayorder [Integer] display order from the package provides
+      # @return [Y2Packager::Product] converted product
+      def from_resolvable(product, installation_package = "not defined",
```
imobachgs: why "not defined"?

imobachgs: Discussed in IRC: we considered using `nil` but @schubi2 was afraid of possible side-effects so we decided to go with an empty string. Good enough IMHO.

PR #977, opened 2019-10-31
Add linuxrc option reboot_timeout to configure before reboot (bsc#1122493)

okurz:


library/general/src/modules/Linuxrc.rb
```diff
@@ -286,6 +290,7 @@ def disable_remote(services)
     publish function: :keys, type: "list <string> ()"
     publish function: :value_for, type: "string (string)"
     publish function: :disable_remote, type: "list <string> ()"
+    publish function: :reboot_timeout, type: "string (string)"
```
jreidinger: signature looks really strange, it does not have any parameter and also it returns boolean, not string.

library/general/src/modules/Linuxrc.rb
```diff
@@ -141,6 +141,10 @@ def text
       InstallInf("Textmode") == "1"
     end
 
+    def reboot_timeout
+      !InstallInf("Reboot_timeout").nil?
```
jreidinger: if reboot_timeout is integer with value of timeout, then this code works strangely. I expect it will return number or default if not specified? So if default is 0, I expect code like
```
InstallInf("Reboot_timeout").to_i
```

library/general/test/linuxrc_test.rb
```diff
@@ -46,7 +46,8 @@
   "XBusID"           => "0:2:0",
   "XkbRules"         => "xfree86",
   "XkbModel"         => "pc104",
-  "umount_result"    => "0"
+  "umount_result"    => "0",
+  "reboot_timeout"   => "10"
```
jreidinger: I expect here also test for that value, that it really return 10 from method ( which is not case now ).

library/general/src/modules/Linuxrc.rb
okurz: I must admit, I don't really know what I am doing here :) I was merely copying what I found for display_ip assuming that I want something comparable here. What I wanted to achieve is that the user can specify a parameter "reboot_timeout" with an integer value that is used in https://github.com/yast/yast-installation/pull/823 to override the default of "10" when specified. Can you suggest how I can do that?

jreidinger: ok, so let me write it here. Here you should add that method does not have params and returns interger `type: "integer ()"`

library/general/src/modules/Linuxrc.rb
jreidinger: ok, so when checking another pr, you want to return nil if it is not specified. So code should look like:

```
return nil unless InstallInf("Reboot_timeout")

InstallInf("Reboot_timeout").to_i
```

library/general/test/linuxrc_test.rb
jreidinger: and here it means add some meaningful test below so something like:

```
  describe "#reboot_timeout" do
    it "returns integer value if 'reboot_timeout' is found in install.inf" do
      load_install_inf("reboot_timeout" => "15")
      expect(subject.reboot_timeout).to eq(15)
    end

    it "returns nil if 'reboot_timeout' is not found in install.inf" do
      load_install_inf({})
      expect(subject.reboot_timeout).to eq(nil)
    end
  end
```

add it to other test below that describe other methods.

okurz: thx, done

jreidinger:
now it looks good, just rubocop failing. But it has autocorrection, so you should be fine when running `rake check:rubocop:auto_correct` or you can do it manually by adding line after return at https://github.com/yast/yast-yast2/pull/977/files#diff-b8b68eb2ba2e5a979f2a541dff73c387R145

okurz:
Thank you for mentoring me in the process :) All green now

jreidinger:
@okurz you are welcome and thanks for patch.

PR #980, opened 2019-11-15
Improve the #wrap_text helper

dgdavid:
## Problem

The [`wrap_text`](https://github.com/yast/yast-yast2/blob/43753ef77843e1a43a691b59154e52efb966785c/library/general/src/lib/ui/text_helpers.rb#L31-L53) helper method [does not respect new lines already present](https://github.com/yast/yast-installation/pull/827#issuecomment-553808305) in the given text.

## Solution

Just to preserve the already present new lines.

Now the method,  _slightly_ inspired in the [Rails' `word_wrap`](https://github.com/rails/rails/blob/dcaa388badf49423fb8613bc02652e448890f879/actionview/lib/action_view/helpers/text_helper.rb#L260-L264), is also a bit more efficient.

Additionally, the logic to extract only a few lines after wrap the text has been moved to a new [`head`](https://github.com/yast/yast-yast2/blob/ea2d2d304b234721aae82b91547a1593f50fee68/library/general/src/lib/ui/text_helpers.rb#L50-L67) method.

## Some considerations

* No version bump is needed. Just wait until next one.
* Please, see also [my self review below](#pullrequestreview-317581410) :point_down: 

## Tests

* Unit tests adapted
* Tested manually via dud

## Screenshots

<details>
<summary>Click to toggle the screenshots</summary>

<hr/>

<p align="center">Before</p>

  using wrap_text

   
<p align="center">After</p>
  
  fixed_wrap_text
</details>

---

Related to https://github.com/yast/yast-installation/pull/827#issuecomment-553808305

### DIFF
--- a/library/general/src/lib/ui/text_helpers.rb
+++ b/library/general/src/lib/ui/text_helpers.rb
@@ -27,17 +27,14 @@ module UI
     #
     # @param text [String] text to be wrapped
     # @param line_width [Integer] max line length
-    # @option preprend_text [String] the text to prepended
     # @param n_lines [Integer, nil] the maximum number of lines
     # @param cut_text [String] the omission text to be used when the text should be cut
     #
     # @return [String]
-    def wrap_text(text, line_width = 76, prepend_text: "", n_lines: nil, cut_text: "")
-      full_text = prepend_text + text
+    def wrap_text(text, line_width = 76, n_lines: nil, cut_text: "")
+      return text if line_width > text.length
 
-      return full_text if line_width > full_text.length
-
-      wrapped_text = full_text.lines.collect! do |line|
+      wrapped_text = text.lines.collect! do |line|
         l = (line.length > line_width) ? line.gsub(/(.{1,#{line_width}})(?:\s+|$)/, "\\1\n") : line
         l.strip
       end
--- a/library/general/src/lib/ui/text_helpers.rb
+++ b/library/general/src/lib/ui/text_helpers.rb
@@ -43,18 +43,20 @@ module UI
       end
 
       result = wrapped_text.join("\n")
-      result = excerpt(result, n_lines, omission: cut_text) if n_lines
+      result = head(result, n_lines, omission: cut_text) if n_lines
       result
     end
 
-    # Extract an excerpt of given text if it has more lines than the maximum allowed
+    # Returns only the first requested lines of the given text
+    #
+    # If the omission param is given, an extra line holding it will be included
     #
     # @param text [String]
     # @param max_lines [Integer]
     # @param omission [String] the text to be added
     #
-    # @return [String] an excerpt if given text had more lines than requested; full text otherwise
-    def excerpt(text, max_lines, omission: "")
+    # @return [String] the first requested lines if the text has more; full text otherwise
+    def head(text, max_lines, omission: "")
       lines = text.lines
 
       return text if lines.length <= max_lines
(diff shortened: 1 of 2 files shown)


jreidinger:
@dgdavid prepend_text variant was used in yast2-network pre network-ng. Maybe time to get rid of that option if it is not used and reintroduce only when requirement appear?

library/general/src/lib/ui/text_helpers.rb
```diff
@@ -20,36 +20,48 @@
...
+    end
+
+    # Extract an excerpt of given text if it has more lines than the maximum allowed
+    #
+    # @param text [String]
+    # @param max_lines [Integer]
+    # @param omission [String] the text to be added
+    #
+    # @return [String] an excerpt if given text had more lines than requested; full text otherwise
+    def excerpt(text, max_lines, omission: "")
```
jreidinger: name is really hard to remember. What about max_lines method name?

dgdavid: What do you think about `#fragment` or `#passage` (https://www.thesaurus.com/browse/excerpt)

dgdavid:
> @dgdavid prepend_text variant was used in yast2-network pre network-ng. Maybe time to get rid of that option if it is not used and reintroduce only when requirement appear?

Fine with me.

@teclator, what do you think?

library/general/src/lib/ui/text_helpers.rb
jreidinger: well, my issue is that it is not real fragment or passage. It is just limiting text to have at max X lines. You cannot have senteces or stuff like this, so it can looks like it can do more then it really do.

dgdavid: Ah! Now I got it :)

Ok, what about [`head`](https://github.com/yast/yast-yast2/pull/980/commits/ea2d2d304b234721aae82b91547a1593f50fee68)? Just a suggestion. I can change it to `max_lines` if you think it's not a good idea to use that word in a helper or still being not completely accurate.

Thanks!

jreidinger: head is fine for me.

teclator:
> > @dgdavid prepend_text variant was used in yast2-network pre network-ng. Maybe time to get rid of that option if it is not used and reintroduce only when requirement appear?
> 
> Fine with me.
> 
> @teclator, what do you think?

If the options are not needed anymore, then it is also fine with me to get rid of them. BTW, it also LGTM as it is.

dgdavid:
Since it seems it's pretty sure that `pretend_text` option is not being used, it has been removed. At the end, to prepend a text before wrapping it is as easy as 

```ruby
wrap("Please read carefully before continue: #{warning_text}")
```

Regarding the other options, as there is not so much information I am going to keep them for now, but ideally it would be nice to remove them and also [to decouple `wrap_text` and `head`](#pullrequestreview-317581410) methods. BTW, [I did a benchmark](https://gist.github.com/dgdavid/4bea9bbd4b60d213c2c55f3af051b04a) and the performance penalty seems to be not significant enough.

dgdavid:
> Since it's pretty sure that pretend_text option is not being used —@dgdavid, [Nov 18, 2019](#issuecomment-555096900)

As it can be seen, yast2-network was using it. So, both lessons learned

* Nothing is _pretty sure_
* It worth to re-check things more than twice :)

PR #981, opened 2019-11-18
Network: drop unsupported technologies

teclator:
## Problem

There are many network interfaces that are rarely used or the support in SUSE has been deprecated. As we do not plant to support them in network-ng we should remove them in order to not be offered when adding a new connection.

The current list is obtained by the 'Yast::NetworkInterfaces.GetDeviceTypes' method

https://github.com/yast/yast-network/blob/90bf250c3cd8becf7f7196dea1e83449adaac900/src/lib/y2network/widgets/interface_type.rb#L51

AddConnection

- https://trello.com/c/qg8k5n8y/1379-3-feature-jscsle-7753-drop-deprecated-network-technologies-in-yast2

## Solution

By now, the interfaces requested in the feature have been removed from the list used for configuring a new connection.

The `NetworkInterfaces.GetDeviceTypes` dependency has been also removed from yast2-network but without depending on the specific yast2 version used by this PR. For that reason, we have not dropped the methods yet and just adapted them.

### DIFF
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -100,7 +100,7 @@ module Yast
       @CardRegex =
         # other: irlan|lo|plip|...
         {
-          "netcard" => "ath|bnep|ci|ctc|slc|dummy|bond|escon|eth|fddi|ficon|hsi|qeth|lcs|iucv|myri|tr|usb|wlan|xp|vlan|br|tun|tap|ib|em|p|p[0-9]+p",
+          "netcard" => "ath|ci|ctc|slc|dummy|bond|eth|ficon|hsi|qeth|lcs|iucv|wlan|xp|vlan|br|tun|tap|ib|em|p|p[0-9]+p",
           "modem"   => "ppp|modem",
           "isdn"    => "isdn|ippp",
           "dsl"     => "dsl"
@@ -1013,18 +1013,14 @@ module Yast
     # for this hardware
     def GetDeviceTypes
       # common linux device types available on all architectures
-      common_dev_types = ["eth", "tr", "vlan", "br", "tun", "tap", "bond"]
+      common_dev_types = ["eth", "vlan", "br", "tun", "tap", "bond"]
 
       # s390 specific device types
-      s390_dev_types = ["hsi", "ctc", "escon", "ficon", "iucv", "qeth", "lcs"]
+      s390_dev_types = ["hsi", "ctc", "ficon", "iucv", "qeth", "lcs"]
 
       # device types which cannot be present on s390 arch
       s390_unknown_dev_types = [
-        "bnep",
         "dummy",
-        "fddi",
-        "myri",
-        "usb",
         "wlan",
         "ib"
       ]
@@ -1067,16 +1063,9 @@ module Yast
 
       device_types = {
         # Device type label
-        "arc"   => [_("ARCnet"), _("ARCnet Network Card")],
-        # Device type label
         "atm"   => [
           _("ATM"),
           _("Asynchronous Transfer Mode (ATM)")
-        ],
-        # Device type label
-        "bnep"  => [
-          _("Bluetooth"),
-          _("Bluetooth Connection")
         ],
         # Device type label
         "bond"  => [_("Bond"), _("Bond Network")],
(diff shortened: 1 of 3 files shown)


library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -100,7 +100,7 @@ def main
       @CardRegex =
         # other: irlan|lo|plip|...
         {
-          "netcard" => "ath|bnep|ci|ctc|slc|dummy|bond|escon|eth|fddi|ficon|hsi|qeth|lcs|iucv|myri|tr|usb|wlan|xp|vlan|br|tun|tap|ib|em|p|p[0-9]+p",
+          "netcard" => "ath|ci|ctc|slc|dummy|bond|eth|ficon|hsi|qeth|lcs|wlan|xp|vlan|br|tun|tap|ib|em|p|p[0-9]+p",
```
jreidinger: I am not sure if we should do it before we have ready that unsupported card type. Also some other entries here looks a bit strange like ath, ci, xp, p and pXp.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1169,22 +1142,13 @@ def GetDevTypeDescription(type, longdescr)
...
         "wlan"  => [
           _("Wireless"),
           _("Wireless Network Card")
         ],
         # Device type label
-        "xp"    => [_("XPNET"), _("XP Network")],
-        # Device type label
         "vlan"  => [_("VLAN"), _("Virtual LAN")],
         # Device type label
         "br"    => [_("Bridge"), _("Network Bridge")],
```
jreidinger: I see in this list more opurtinities for cleaning :)

library/network/src/modules/NetworkInterfaces.rb
teclator: Yep, I was not sure about some of the because of not mentioned in the feaure request. 'ath' should be recognized as wlan if I'm not wrong. 'xp' removed, 'ppp' we already have a request for adding them back... will check all of them carefully

library/network/src/modules/NetworkInterfaces.rb
teclator: I'm not sure about getting rid of the descriptions at all. The important part is to determining which are not supported for skipping them from the interfaces list as well for preventing them to be chosen when configuring the network.

library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1013,18 +1013,14 @@ def Import(devregex, devices)
     # for this hardware
     def GetDeviceTypes
```
teclator: The GetDeviceTypes is which really show us the options available:

https://github.com/yast/yast-network/blob/90bf250c3cd8becf7f7196dea1e83449adaac900/src/lib/y2network/widgets/interface_type.rb#L52

mchf:
We should cut NetworkInterfaces.GetType* from yast-network completely as we already have required functionality implemented there in TypeDetector. Quick grep showed that only one place remained where GetType is used in yast2-network. So we should check whether this code is still in use and replace it if yes or remove the code completely. Otherwise we are in risk of inconsistency of type detection ... again.

Otherwise lgtm.

library/network/src/modules/NetworkInterfaces.rb
mchf: In my POV it should, or already is, be replaced by InterfaceType in the yast2-network. I really don't see (in fact have never seen) a reason to have the description in NetworkInterfaces.

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Sun Nov 17 23:21:31 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
+- Network: dropped support for 'bne', 'escon', 'fddi', 'tr',
+  'myrinet', 'xp' and 'usb' netcards (jsc#SLE-7753)
```
mchf: well I'd simply wrote "for obsolete network device types" ... but that is only NP

teclator:
> We should cut NetworkInterfaces.GetType* from yast-network completely as we already have required functionality implemented there in TypeDetector. Quick grep showed that only one place remained where GetType is used in yast2-network. So we should check whether this code is still in use and replace it if yes or remove the code completely. Otherwise we are in risk of inconsistency of type detection ... again.
> 
> Otherwise lgtm.

I agree, and there are also some leftovers which uses NetworkInterfaces.Filter... which we should try to get rid of ASAP.

library/network/src/modules/NetworkInterfaces.rb
teclator: See https://github.com/yast/yast-network/pull/1009

PR #958, opened 2019-09-05
support reading licenses from tar archive (jsc#SLE-7214)

wfeldt:
## Task

Support reading licenses from a tar archive in the repository.

Plus quite some license fetcher refactoring:

- rpm and tar share the same basics
- license archive is now fetched only once and then cached
- moved `#confirmation_required?` from `LicensesHandlers` to `LicensesFetchers`

## See also

This is based on @lslezak's work in

https://github.com/yast/yast-yast2/compare/new_media

## Todo

- unit tests :heavy_check_mark:
- rubocop :heavy_check_mark:
- integration tests :heavy_check_mark:
- proper yardoc comments :heavy_check_mark:

PR #982, opened 2019-11-19
Do no try to fetch the license when it's already provided

dgdavid:
## Problem

Changes made in #880 try to ~~fetch the license content~~ look for a license fetcher even when ~~it~~ the content for default language was already provided ~~, which is wrong although it will not fail most of times.~~

* Trello: https://trello.com/c/ORqk8sk7/1444-1-p1-yast-asks-for-media-because-of-a-license-to-show-but-the-license-is-missing
* Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1153326

## Solution

Don't try to fetch the license text when it is given.

---

Kudos to @lslezak

### DIFF
--- a/library/packages/src/lib/y2packager/license.rb
+++ b/library/packages/src/lib/y2packager/license.rb
@@ -54,7 +54,7 @@ module Y2Packager
         log.info "Searching for a license for product #{product_name}"
         return cache[product_name] if cache[product_name]
 
-        fetcher = LicensesFetchers.for(product_name)
+        fetcher = LicensesFetchers.for(product_name) unless content
         handler = LicensesHandlers.for(fetcher, product_name) if fetcher
 
         license = License.new(product_name: product_name, fetcher: fetcher,
--- a/library/packages/test/y2packager/license_test.rb
+++ b/library/packages/test/y2packager/license_test.rb
@@ -34,14 +34,16 @@ describe Y2Packager::License do
 
   describe ".find" do
     context "when some content is given" do
-      before do
-        allow(Y2Packager::License).to receive(:new)
-          .with(product_name: "SLES", fetcher: fetcher, handler: handler, content: content)
-          .and_return(license)
+      it "does not look for a license fetcher" do
+        expect(Y2Packager::LicensesFetchers).to_not receive(:for)
+
+        described_class.find("SLES", content: content)
       end
 
-      it "uses the content as license's content" do
-        expect(described_class.find("SLES", content: content)).to be(license)
+      it "uses the content as license's text" do
+        product_license = described_class.find("SLES", content: content)
+
+        expect(product_license.content_for).to eq(content)
       end
     end
 
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Nov 19 09:51:48 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Do not try to fetch the license text when it is already provided
+  (bsc#1153326).
+- 4.1.75
+
+-------------------------------------------------------------------
 Fri Sep  6 11:15:40 UTC 2019 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Fix a problem when long warnings reports in command line
--- a/package/yast2.spec
+++ b/package/yast2.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2
-Version:        4.1.74
+Version:        4.1.75
 
 Release:        0
 Summary:        YaST2 - Main Package


dgdavid:
Actually, this bug was reported after introduce [the tarball fetcher](https://github.com/yast/yast-yast2/pull/958), which will be available only from SLE-15-SP2 on. But _backporting_ the fix to SLE-15-P1 and, thanks to @imobachgs comments in the office, I realize that changes related to avoid look for a fetcher are not really needed, or even undesired. Let me explain it.

The original changes also includes [the fix](https://github.com/yast/yast-yast2/compare/skip_license_fetcher?expand=1#diff-a7c827929d4041ee31eb3befc039d79eR61) for mentioned fetcher, which is _simply_ about using `Yast::Pkg.SourceProvideOptionalFile` instead of `Yast::Pkg.SourceProvideFile`. 

But we must keep the code to look for a fetcher because, although the content for default language could be already given, the fetcher is necessary to [retrieve the license translations](https://github.com/yast/yast-yast2/blob/0f84dfbd306fb392af6d1f84191cc96a937416a0/library/packages/src/lib/y2packager/license.rb#L136).

@lslezak, @teclator, friendly ping :)

lslezak:
> But we must keep the code to look for a fetcher because, although the content for default language could be already given, the fetcher is necessary to [retrieve the license translations](https://github.com/yast/yast-yast2/blob/0f84dfbd306fb392af6d1f84191cc96a937416a0/library/packages/src/lib/y2packager/license.rb#L136).

Well in this case the license translations are also downloaded from the outside, the license fetchers cannot download the translations from the SCC anyway.

The problem is that the `License` module API does not allow passing the translations, it only has a single `content` parameter.

From y2log:
```
registration/eula_downloader.rb:82 Downloading license index from https://updates.suse.com/SUSE/Products/SLE-Product-HA/15-SP2/x86_64/product.license/directory.yast...
registration/eula_downloader.rb:87 Downloaded license index: ["license.de.txt", "license.es.txt", "license.fr.txt", "license.it.txt", "license.ja.txt", "license.ko.txt", "license.pt_BR.txt", "license.ru.txt", "license.txt", "license.zh_CN.txt", "license.zh_TW.txt"]
registration/eula_downloader.rb:58 Downloading license from https://updates.suse.com/SUSE/Products/SLE-Product-HA/15-SP2/x86_64/product.license/license.de.txt...
registration/eula_downloader.rb:60 Downloaded license: "SUSE(r)-Endbenutzer-Lizenzvereinb"... (21580 bytes)
registration/eula_downloader.rb:65 Saving the license to file: /tmp/extension-eula-20191115-11979-1x2oehz/license.de.txt
...
registration/eula_reader.rb:72 EULA files in /tmp/extension-eula-20191115-11979-1x2oehz: {"zh_TW"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.zh_TW.txt", "zh_CN"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.zh_CN.txt", "en_US"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.txt", "ru"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.ru.txt", "pt_BR"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.pt_BR.txt", "ko"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.ko.txt", "ja"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.ja.txt", "it"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.it.txt", "fr"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.fr.txt", "es"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.es.txt", "de"=>"/tmp/extension-eula-20191115-11979-1x2oehz/license.de.txt"}
```

But the code passes the temporary directory with all translations to the license dialog which shows them correctly.

dgdavid:
@lslezak, @imobachgs 

I tested the patch[1] in an openSUSE-20191116 there is no problem with license translations. Which makes sense because none `content` is given to the `License.rb`. [See logs](https://github.com/yast/yast-yast2/files/3868006/y2log.log)

---


[1] Actually, I edited the `/usr/share/YaST2/lib/y2packager/license.rb` file and added a `log.info` line and the `unless content` condition by hand, because I had problems using `yupdate` (already talked with @lslezak)

Screenshot from 2019-11-20 08-01-31

lslezak:
I'd just fix it in the `master` branch (for SP2), the bug might be already present in SP1 but probably does not make troubles there, I haven't seen anything like this in SP1. Moreover it would have to be released as a self-update...

Let's fix it in SP2 for now and do the backport only if we get a bug report for SP1.

PR #984, opened 2019-11-21
Read installinf instsys

teclator:
## Problem

Linuxrc maintains the state of what was read from '/etc/install.inf' file. If for any reason the internal state is reset and the SCR root was already modified then, it is probably that we try to read the file from the target system instead of the local one, and if already unmounted that could be like shooting ourselves in the foot.

- https://bugzilla.suse.com/show_bug.cgi?id=1157476 (it also contains a long history of the problem itself)

## Solution

For avoiding that problem we ensure to use "/" as SCR root before reading "/etc/install.inf"


teclator:
Public job failed..

PR #987, opened 2019-11-22
Do not modify /etc/sysctl.conf in inst-sys

mchf:
Found by QA
Tested manually

~~TODO: bug number + changes + version~~

### DIFF
--- a/library/general/src/lib/cfa/sysctl.rb
+++ b/library/general/src/lib/cfa/sysctl.rb
@@ -46,6 +46,8 @@ module CFA
   class Sysctl < BaseModel
     include Yast::Logger
 
+    Yast.import "Stage"
+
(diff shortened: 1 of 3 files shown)


library/general/src/lib/cfa/sysctl.rb
```diff
@@ -128,7 +130,9 @@ def load
     # +/etc/sysctl.conf+ to avoid confusion.
     def save
       super
-      clean_old_values
+
+      # we cannot update /etc/sysctl.conf in first stage as it is on ro filesystem
+      clean_old_values if Yast::Stage.normal || Yast::Stage.cont
```
jsrain: Would it be better    ... unless YaST::Stage.initial ? Just thinking about firstboot... And I cannot remember what  hardware_probed stage is for :-/

PR #989, opened 2019-11-26
Fix syntax for Ruby 2.5

imobachgs:
Omitting the end of the range works in Ruby 2.6 but it looks like it is not allowed in 2.5.


imobachgs:
Me neither :) Thanks.

PR #985, opened 2019-11-21
add is_wsl function to detect wsl

lnussel:

PR #990, opened 2019-11-26
Add is_wsl function to detect wsl

imobachgs:
This is a rebased and slightly improved version of #985 (it adds some simple unit tests and drops the use of `Builtins`).

The PR adds an `is_wsl` function to detect whether the system is running on WSL. See [boo#1154962](https://bugzilla.suse.com/show_bug.cgi?id=1154962) for further details.

### DIFF
--- a/library/general/src/modules/Arch.rb
+++ b/library/general/src/modules/Arch.rb
@@ -494,8 +494,11 @@ module Yast
       true
     end
 
+    # Determines whether the system is running on WSL
+    #
+    # @return [Boolean] true if runs on WSL; false otherwise
     def is_wsl
-      Builtins.regexpmatch(SCR.Read(path(".target.string"), "/proc/sys/kernel/osrelease"), "-Microsoft")
+      SCR.Read(path(".target.string"), "/proc/sys/kernel/osrelease").to_s.include?("-Microsoft")
     end
 
     publish function: :architecture, type: "string ()"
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Arch.rb
```diff
@@ -528,6 +535,7 @@ def x11_setup_needed
     publish function: :is_zkvm, type: "boolean ()"
     publish function: :has_smp, type: "boolean ()"
     publish function: :x11_setup_needed, type: "boolean ()"
+    publish function: :is_wsl, type: "boolean ()"
```
teclator: NP: Not sure if we are still publishing new functions, but is ok if not removed completely.

library/general/src/modules/Arch.rb
```diff
@@ -494,6 +494,13 @@ def x11_setup_needed
       true
     end
 
+    # Determines whether the system is running on WSL
+    #
+    # @return [Boolean] true if runs on WSL; false otherwise
+    def is_wsl
+      SCR.Read(path(".target.string"), "/proc/sys/kernel/osrelease").to_s.include?("-Microsoft")
```
teclator: Now is better as previous implementation (PR #985) could return `nil`

library/general/src/modules/Arch.rb
imobachgs: Yes, you are right. I guess it is needed if you want to call the function from Perl or something like this. Anyway, I am keeping it just for consistency reasons.

teclator:
@imobachgs @lnussel thnx for the pull requests

PR #995, opened 2019-12-02
Use 70-yast.conf instead of 30-yast.conf

imobachgs:
When writing sysctl related settings, YaST will use `70-yast.conf` instead of `30-yast.conf` to reduce the chance of getting its values overridden.

During the installation/upgrade of `yast2.rpm`, if a `30-yast.conf` is found, it will be renamed to `70-yast.conf`. If the `70-yast.conf` already exists, the `30-yast.conf` will be simply removed (this case should not happen).

### DIFF
--- a/package/yast2.spec
+++ b/package/yast2.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2
-Version:        4.2.45
+Version:        4.2.46
 Release:        0
 Summary:        YaST2 Main Package
 License:        GPL-2.0-only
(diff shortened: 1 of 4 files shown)


package/yast2.spec
```diff
@@ -177,6 +177,14 @@ mkdir -p %{buildroot}%{_sysconfdir}/YaST2
 %post
 %{fillup_only -n yast2}
 
+if [ -f "/etc/sysctl.d/30-yast.conf" ]; then
+    if [ -f "/etc/sysctl.d/70-yast.conf" ]; then
+        rm /etc/sysctl.d/30-yast.conf
```
dgdavid: From the PR's description 

> If the 70-yast.conf already exists, the 30-yast.conf will be simply removed **(this case should not happen)**.

If we are sure that it will not happen, I'd simplify this.

Anyway, it looks good to me as it is.

PR #996, opened 2019-12-04
Backward compatible resolvable

jreidinger:
Related to changes introduced in https://github.com/yast/yast-update/pull/143, which raises the below internal error

Snímek_obrazovky_2019-12-04_13-56-32

Bugzilla: [#1157636](https://bugzilla.suse.com/show_bug.cgi?id=1157636) and [#1158056](https://bugzilla.suse.com/show_bug.cgi?id=1158056)

Trello: https://trello.com/c/pyqFurT6 and https://trello.com/c/Rhvu29XB

### DIFF
--- a/library/packages/src/lib/y2packager/resolvable.rb
+++ b/library/packages/src/lib/y2packager/resolvable.rb
@@ -80,6 +80,17 @@ module Y2Packager
       from_hash(hash)
     end
 
+    # Backward compatibility method to access resolvable like hash.
+    def [](key)
+      log.info "Calling [] with #{key}. Deprecated. Use method name directly"
+
+      public_send(key.to_sym)
+    # key not found, so return nil to be compatible
+    rescue NameError
+      log.info "attribute #{key} not defined, returning nil."
+      nil
+    end
+
     #
     # Dynamically load the missing attributes from libzypp.
     #
--- a/library/packages/src/lib/y2packager/resolvable.rb
+++ b/library/packages/src/lib/y2packager/resolvable.rb
@@ -82,12 +82,13 @@ module Y2Packager
 
     # Backward compatibility method to access resolvable like hash.
     def [](key)
-      log.info "Calling [] with #{key}. Deprecated. Use method name directly"
+      log.info "Calling [] with #{key}. It is deprecated. Use method name " \
+        "directly. Called from #{caller(1).first}"
 
       public_send(key.to_sym)
     # key not found, so return nil to be compatible
     rescue NameError
-      log.info "attribute #{key} not defined, returning nil."
+      log.warn "attribute #{key} not defined, returning nil."
       nil
     end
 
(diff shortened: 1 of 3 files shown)


library/packages/src/lib/y2packager/resolvable.rb
```diff
@@ -80,6 +80,17 @@ def initialize(hash)
       from_hash(hash)
     end
 
+    # Backward compatibility method to access resolvable like hash.
+    def [](key)
+      log.info "Calling [] with #{key}. Deprecated. Use method name directly"
```
lslezak: Maybe also log the caller so we can easily find the place where it was called from?

library/packages/src/lib/y2packager/resolvable.rb
```diff
@@ -80,6 +80,17 @@ def initialize(hash)
...
+
+      public_send(key.to_sym)
+    # key not found, so return nil to be compatible
+    rescue NameError
+      log.info "attribute #{key} not defined, returning nil."
```
lslezak: I think this should be a warning, the list of the attributes is more or less fixed, using a missing attribute is really wrong.

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Dec  4 14:26:35 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
+
+- Add backward compatible hash accessors to Resolvable which solve
+  several crashes ( related to bsc#1132650 and bsc#1140037)
```
lslezak: Extra space
```suggestion
  several crashes (related to bsc#1132650 and bsc#1140037)
```

library/packages/src/lib/y2packager/resolvable.rb
jreidinger: good idea.

library/packages/src/lib/y2packager/resolvable.rb
jreidinger: hmm, so using release_type is error there at https://github.com/yast/yast-registration/blob/c9b2889d13857ebba404e9db00cecf27cb58ffdb/src/lib/registration/registration.rb#L118 ?

lslezak: Hm,`release_type` has never been supported in [Pkg.ResolvableProperties](https://github.com/yast/yast-pkg-bindings/blob/master/src/Resolvable_Properties.cc), but there is [register_release](https://github.com/yast/yast-pkg-bindings/blob/master/src/Resolvable_Properties.cc#L385) value.

It seems that the SUSE Connect also uses that (with some additional magic): https://github.com/SUSE/connect/blob/7f743bc184a87b995ea2f02008cdb3b22387b839/lib/suse/connect/zypper/product.rb#L38

But AFAIK that value is `nil` in most (all?) cases so far, so that was not a problem.

PR #997, opened 2019-12-05
Fallback to ::FileUtils those Yast::FileUtils missing methods

dgdavid:
## :construction:  WIP

Alghouth it works, this PR still WIP waiting to

* Feedback from yast-devel: https://lists.opensuse.org/yast-devel/2019-12/msg00003.html
* Document the methods (at least, why they are there)
* Maybe a small unit test?

## Problem

Sometimes [`Yast::FileUtils`](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/FileUtils.rb) is taking the place of [`::FileUtils`](https://ruby-doc.org/stdlib-2.6.5/libdoc/fileutils/rdoc/FileUtils.html)

---

See also, 

* Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1158421
* Trello: https://trello.com/c/qpFItjng
* PR related to the bug report: https://github.com/yast/yast-ftp-server/pull/54

## Solution

Send the missing `Yast::FileUtils` methods to `::FileUtils` (and log the action)

### DIFF
--- a/library/general/src/modules/FileUtils.rb
+++ b/library/general/src/modules/FileUtils.rb
@@ -33,12 +33,32 @@ require "shellwords"
 
 module Yast
   class FileUtilsClass < Module
+    include Yast::Logger
+
     def main
       textdomain "base"
       Yast.import "Popup"
       Yast.import "String"
 
       @tmpfiles = []
+    end
+
+    def method_missing(method_name, *args, &block)
+      if ::FileUtils.respond_to?(method_name)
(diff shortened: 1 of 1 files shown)


dgdavid:
According to the feedback given in https://lists.opensuse.org/yast-devel/2019-12/msg00003.html, we're going to use the scope resolution operator `::` whenever is needed instead.

PR #998, opened 2019-12-12
Avoid a potential crash when calling Y2Packager::Resolvable#deps

imobachgs:
This problem was spotted while running the test suite. Perhaps it won't happen in a real scenario, but I prefer to prevent such a crash.

```ruby
resolvable = Y2Package::Resolvable.new("name" => "foobar")
resolvable.deps #=> RuntimeError (Missing attributes for identifying the resolvable.)
```

Why am I fixing just the `#deps` case and not doing anything about `#name` or `#arch`? Well, the reason is that reading the `yast-pkg-bindings` code I guess [dependencies can be missing](https://github.com/yast/yast-pkg-bindings/blob/6abebc1ed823ccd06495ad39a19a4f5b89279bcb/src/Resolvable_Properties.cc#L529). But I would expect the name or the arch to be present always.

However, I would prefer to rethink the whole approach behind the `Resolvable` class. I prefer to have real instance methods to rely on `#method_missing`. If desired, we could even use metaprogramming to reduce the overhead (and setting default values like the `#deps` case) when creating those methods. But that's out of scope.

### DIFF
--- a/library/packages/src/lib/y2packager/resolvable.rb
+++ b/library/packages/src/lib/y2packager/resolvable.rb
@@ -77,6 +77,7 @@ module Y2Packager
     #
     # @param hash [Hash<Symbol,Object>] A pkg-bindings resolvable hash.
     def initialize(hash)
+      @deps = []
       from_hash(hash)
     end
 
--- a/library/packages/test/y2packager/resolvable_test.rb
+++ b/library/packages/test/y2packager/resolvable_test.rb
@@ -106,4 +106,26 @@ describe Y2Packager::Resolvable do
       expect { res.vendor("dummy") }.to raise_error(ArgumentError)
     end
   end
+
+  describe "#deps" do
+    subject(:resolvable) { Y2Packager::Resolvable.new(pkg_data) }
+
+    let(:pkg_data) do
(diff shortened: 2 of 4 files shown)


lslezak:
Using `Y2Package::Resolvable.find` ensures that all needed attributes for lazy loading the other values are present in the returned objects.

If you use `Y2Package::Resolvable.new` manually you should prefill all attributes used later, it's the caller's responsibility.

If you use
```ruby
r = Y2Packager::Resolvable.new(name: "foo")
```

then anything like `r.arch`, `r.license`, `r.version` will fail the same way. It probably does not make sense to handle this single attribute specifically.

We should rather fix pkg-bindings to return an empty list if the dependencies are requested. 

E.g. [here](https://github.com/yast/yast-pkg-bindings/blob/master/src/Resolvable_Properties.cc#L574) we should change it

```diff
-		if (rawdeps.size() > 0)
+		if (rawdeps.size() > 0 || attrs->contains(YCPSymbol("deps")))
			info->add (YCPString ("deps"), rawdeps);
```

imobachgs:
OK, fine with me. I still do not like relying in `#method_missing` in this case to "automagically" implement these methods, but that's a different topic.

PR #999, opened 2019-12-18
Service widget default

teclator:
## Problem

The new service widget was [introduced]((https://trello.com/c/uAe4i9Ru/107-5-fate319428-allow-socket-activation-widget)) in SLE-15-SP1, and by default, the option **keep_current_state** is selected. Which means that, although the user made same changes in the configuration, the changes were not reflected in the service as the service itself was untouched. 

Reported in: https://bugzilla.suse.com/show_bug.cgi?id=1158946

The main problem is that the option is usually overlooked by the user, and it does not have the same `default` as it had in previous SLE versions. At least it is the case of yast2 dns-server as we can see in the images below.

| **SLE-12-SP4 dns** | **SLE-15-SP1** |
|       :---:                    |      :---:               |
| SLE-12-SP4 |  SLE-15-SP1 |

In the SLE-12-SP4 image, we can see that the checkbox for **reloading** the service in case of modified configuration is **checked** by default.

- https://trello.com/c/q5DPg1Vh/1515-sles15-sp1-l3-important-1158946-namedservice-is-not-restarted-after-change-in-yast

## Solution

When the widget is initialized it checks if the service is running or not. 

- In case of `active`, it will set the default action as **reload** when supported or **restart** when not. 
- In case of 'inactive`, it will not propose any action per default, that is, the option selected will continue being the **keep_current_state** one.

Also, as applying the changes will modify the action items, the widget will try to select the default action when refreshing if the action is available. **UPDATE:** As this could be problematic (changing from restart to reload or other action), this option has been redefined and now the selected option will be remembered.

Last but not least, the option can be modified per module, ie:

```ruby
     # Example modifying the code in dns-server
     def service_widget
       return if @service_widget

       @service_widget = CWM::ServiceWidget.new(service)
       @service_widget.default_action = :restart if service.running?
       @service_widget
     end
```

DnsServerServiceWidget

This new GIF shows an updated version which uses a default option for initializing the widget but then stores the selected option in case of applying changes. That is not a problem in common dialogs where the refresh method is not called at all.

```ruby
     # Example modifying the code in dns-server
     def service
       return if @service

       @service = CWM::ServiceWidget.new(service)
       @service.restart if service.running?
       @service
     end
```
DnsServerServiceWidgetRememberAction

### DIFF
--- a/library/cwm/src/lib/cwm/service_widget.rb
+++ b/library/cwm/src/lib/cwm/service_widget.rb
@@ -23,6 +23,8 @@ require "cwm/custom_widget"
 module CWM
   # CWM wrapper for Yast2::ServiceWidget
   class ServiceWidget < CustomWidget
+    extend Forwardable
+
     # creates new widget instance for given service
     # @param service [Yast2::SystemService,Yast2::CompoundService] service to be configured
     def initialize(service)
@@ -30,16 +32,10 @@ module CWM
       self.handle_all_events = true
     end
 
-    def default_action=(value)
-      @service_widget.default_action=(value)
-    end
+    def_delegators :@service_widget, :refresh, :store, :help, :default_action=
 
     def contents
       @service_widget.content
-    end
-
-    def refresh
-      @service_widget.refresh
     end
 
     def handle(event)
@@ -49,18 +45,10 @@ module CWM
       @service_widget.handle_input(event["ID"])
     end
 
-    def store
-      @service_widget.store
-    end
-
     # The widget needs to be refreshed each time it is rendered. Otherwise, cached
     # service values would not be selected (e.g., when switching in a DialogTree)
     def init
       refresh
-    end
-
-    def help
-      @service_widget.help
     end
   end
 end
--- a/library/cwm/src/lib/cwm/service_widget.rb
+++ b/library/cwm/src/lib/cwm/service_widget.rb
@@ -30,6 +30,10 @@ module CWM
       self.handle_all_events = true
     end
 
+    def default_action=(value)
+      @service_widget.default_action=(value)
+    end
+
     def contents
       @service_widget.content
     end
--- a/library/systemd/src/lib/yast2/service_widget.rb
+++ b/library/systemd/src/lib/yast2/service_widget.rb
@@ -64,6 +64,8 @@ module Yast2
     include Yast::Logger
     include Yast::UIShortcuts
 
+    attr_reader :default_action
+
     # creates new widget instance for given service
     # @param service [Yast2::SystemService,Yast2::CompoundService] service
     def initialize(service)
@@ -72,6 +74,11 @@ module Yast2
       # When the service is active, by default, propose to reload or restart
       # it after writting the configuration (bsc#1158946)
       init_default_action
+    end
+
+    def default_action=(value)
+      @default_action = value
+      service.public_send(default_action) if valid_action?(value)
     end
 
     # gets widget term
@@ -99,6 +106,7 @@ module Yast2
     #
     # @return [nil]
     def refresh
+      service.public_send(default_action) if valid_action?(default_action)
       Yast::UI.ChangeWidget(Id(:service_widget_status), :Value, status)
       Yast::UI.ChangeWidget(Id(:service_widget_action), :Items, action_items)
       Yast::UI.ChangeWidget(Id(:service_widget_autostart), :Items, autostart_items)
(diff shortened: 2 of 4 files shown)


library/cwm/src/lib/cwm/service_widget.rb
```diff
@@ -30,6 +30,10 @@ def initialize(service)
       self.handle_all_events = true
     end
 
+    def default_action=(value)
+      @service_widget.default_action=(value)
```
dgdavid: This looks weird. Likely a copy and paste leftover.

teclator:
It does not cover the 100% of the cases, as refresh is problematic. 

That is, for example when a user modifies the action from **reload** to **restart** and `apply the changes`, it will set the default action again to reload, that is maybe something we could handle in the store method but I don't want to over-engineering

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -199,6 +218,12 @@ def action_widget
       )
     end
 
+    def valid_action?(value)
```
imobachgs: I would document this method too. What is a valid action? (I know :smiley:, but I would like to be it documented).

library/cwm/src/lib/cwm/service_widget.rb
imobachgs: +1

library/systemd/src/lib/yast2/service_widget.rb
teclator: Yep, in general will give it a little more love, wanted to discuss the approach in general, it makes sense or not :)

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -64,11 +64,21 @@ class ServiceWidget
...
       textdomain "base"
       @service = service
+      # When the service is active, by default, propose to reload or restart
+      # it after writting the configuration (bsc#1158946)
+      init_default_action
+    end
+
+    def default_action=(value)
+      @default_action = value
+      service.public_send(default_action) if valid_action?(value)
```
teclator: I do not like at all to already modify the service action before calling the store method.

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -64,11 +64,21 @@ class ServiceWidget
     include Yast::Logger
     include Yast::UIShortcuts
 
+    attr_reader :default_action
```
joseivanlopez: Please, document this.

library/systemd/src/lib/yast2/service_widget.rb
joseivanlopez: The logic of this method is kind of duplicated in `#action_items`. I would try to unify it.

library/systemd/src/lib/yast2/service_widget.rb
jreidinger: so why you do not have it in store method?

library/cwm/src/lib/cwm/service_widget.rb
jreidinger: when I see so much redirection, maybe we should consider using Forwardable here.

library/systemd/src/lib/yast2/service_widget.rb
jreidinger: also why it is needed?

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -96,6 +106,7 @@ def content
     #
     # @return [nil]
     def refresh
+      service.public_send(default_action) if valid_action?(default_action)
```
jreidinger: refresh should not change default action for sure.

library/cwm/src/lib/cwm/service_widget.rb
teclator: I already have it in the changes I'm preparing

library/systemd/src/lib/yast2/service_widget.rb
teclator: I plan to add it, and modify the defaut_action per current action, it needs to modify also the action_items, will commit the changes ASAP

library/systemd/src/lib/yast2/service_widget.rb
```diff
@@ -60,15 +65,32 @@ module Yast2
...
       textdomain "base"
       @service = service
+      # When the service is active, by default, propose to reload or restart
+      # it after writting the configuration (bsc#1158946)
+      init_default_action
+    end
+
+    # Set the given action as the current action selected by the widget
+    #
+    # @param value [Symbol, nil]
```
jreidinger: please document what means nil here. As it can be also interpreted as reset to default.

PR #1003, opened 2019-12-19
Create the snapshot store directory if needed

imobachgs:
Create the directory to save the snapshot IDs if needed. See [bsc#1159562](https://bugzilla.suse.com/show_bug.cgi?id=1159562) for further information about the problem.

This PRs should solve the issue, although I have not tested it yet. On the other hand, perhaps a more general approach might be preferred (like creating all those directories when YaST starts).


imobachgs:
I do not think it is the best approach. After all, we might need this directory to exist in other parts of the code. So perhaps creating the directories in advance could be preferred.

PR #1004, opened 2020-01-06
Propagate an error status when a CommandLine module gets an   unknown command

mvidner:
prompted by the testing discussion in https://github.com/yast/yast-ruby-bindings/pull/237
because I wanted to test that fix with `yast2 lan unknowncommand` and it still did not work

### DIFF
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -360,7 +360,7 @@ module Yast
         # translators: error message in command line interface
         Error(Builtins.sformat(_("Unknown Command: %1"), command))
 
-        return { "command" => command }
+        return {}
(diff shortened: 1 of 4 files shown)


library/commandline/src/modules/CommandLine.rb
```diff
@@ -360,7 +360,7 @@ def Parse(arguments)
         # translators: error message in command line interface
         Error(Builtins.sformat(_("Unknown Command: %1"), command))
 
-        return { "command" => command }
+        return {}
```
mvidner: This fix works because our caller actually obeys our API documented above

```rb
#  @return [Hash{String => Object}]  containing the command and it's option. In case of
#        error it is an empty map.
```

PR #1006, opened 2020-01-08
Add missing require clause (bsc#1160362)

ancorgs:
See https://bugzilla.suse.com/show_bug.cgi?id=1160362 for details.

Some things have changed in the `ProductReader` class recently. We don't test the Live Installation (is not a supported scenario), but it looks like it follows a different workflow that ends up with an exception because `ProductReader` contains a call to `Y2Packager::ProductControlProduct.products` bit it does not require `y2packager/product_control_product`.

Likely, in the normal workflow it works by pure luck (because some other class requires `y2packager/product_control_product`).

This pull request adds the missing require together with a smoke test to prove it was indeed missing.

If fact, the whole situation is even worse at this moment.  The current code looks like this:

```ruby
if Yast::Stage.initial && Y2Packager::MediumType.online? && !force_repos
  return Y2Packager::ProductControlProduct.products.each_with_object([]) {..}
end
```

Which means that there is another missing "require" for `Y2Packager::MediumType`. But that's harder to fix and, thus, I'm leaving it out of the scope of this pull request. That class lives in yast2-packager, not in yast2.rpm. To achieve a 100% correct solution, the file (and all its dependencies) should be moved. This is currently not failing (again, by pure luck), so we can ignore the problem for the time being.

### DIFF
--- a/library/packages/src/lib/y2packager/product_reader.rb
+++ b/library/packages/src/lib/y2packager/product_reader.rb
@@ -14,6 +14,7 @@ require "yast"
 require "y2packager/product"
 require "y2packager/product_sorter"
 require "y2packager/resolvable"
+require "y2packager/product_control_product"
 
 Yast.import "Pkg"
 Yast.import "Linuxrc"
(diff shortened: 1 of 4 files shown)


lslezak:
Thanks!

PR #1007, opened 2020-01-09
Do not fail during upgrade if `/var/lib/YaST2` is missing (bsc#1159562)

ancorgs:
See https://bugzilla.suse.com/show_bug.cgi?id=1159562 for details.

When upgrading a system, YaST takes a previous snapshot and stores its id in the `/var/lib/YaST2` directory of the system being upgraded. That directory will always be there, as long as YaST is installed in the system being upgraded.

But when using YaST to upgrade a system in which YaST is not installed, the process fails because the directory is not there.

It's debatable whether upgrading a non-YaST system with YaST is a valid scenario, but the small change introduced in this pull request should be enough to make it work.

### Testing

Tested manually. With this patch, the snapshot is correctly created in both situations (with and without a preexisting `/var/lib/YaST2`).

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan 10 13:43:20 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Fixed error during upgrade if Btrfs is used and '/var/lib/YaST2'
+  is missing (bsc#1159562)
+- 4.2.53
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Jan 10 13:43:20 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Fixed error during upgrade if Btrfs is used and '/var/lib/YaST2`
```
joseivanlopez: NP: wrong character? YaST2`.

PR #1008, opened 2020-01-09
Do not refresh package installation overview if the medium has changed.

schubi2:
**Problem**
Reading Release Notes during installation is impossible because the installation switch to the
"Details" tab when the medium has been changed.
**Solution**
Do not refresh package installation overview if the medium has been changed and the user has switched to the release notes tab.

### DIFF
--- a/library/packages/src/modules/SlideShow.rb
+++ b/library/packages/src/modules/SlideShow.rb
@@ -743,7 +743,7 @@ module Yast
         end
 
         # Don't override explicit user request!
-        SwitchToSlideView() if !(@user_switched_to == :details)
+        SwitchToSlideView() if @user_switched_to != :details
       elsif !ShowingDetails()
         # (true) : Showing release tab if needed
         RebuildDialog(true)
--- a/library/packages/src/modules/SlideShow.rb
+++ b/library/packages/src/modules/SlideShow.rb
@@ -110,6 +110,8 @@ module Yast
   class SlideShowClass < Module
     include Yast::Logger
 
+    attr_accessor :user_switched_to
+
     module UI_ID
       TOTAL_PROGRESS = :progressTotal
       CURRENT_PACKAGE = :progressCurrentPackage
@@ -741,7 +743,7 @@ module Yast
         end
 
         # Don't override explicit user request!
-        SwitchToSlideView() if !@user_switched_to == :details
+        SwitchToSlideView() if !(@user_switched_to == :details)
       elsif !ShowingDetails()
         # (true) : Showing release tab if needed
         RebuildDialog(true)
--- a/library/packages/src/modules/SlideShow.rb
+++ b/library/packages/src/modules/SlideShow.rb
@@ -805,11 +805,12 @@ module Yast
       elsif button == :showSlide && !ShowingSlide()
         if Slides.HaveSlides
           if @user_switched_to == :release_notes
-            # The user is switching from release notes to slide show.
-            # In order to not disturb the user while reading the release notes
-            # we are not updatding the tabs although the slide show has been
-            # changed meanwhile. So we are updating it now before switching to
-            # slide show.
+            # The user is currently in the release notes tab.
+            # In order to not disturb him while reading the release notes
+            # we are not updating the tabs although the slide show has been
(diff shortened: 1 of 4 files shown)


library/packages/src/modules/SlideShow.rb
```diff
@@ -778,18 +780,29 @@ def HandleInput(button)
       button = deep_copy(button)
       if button == :showDetails && !ShowingDetails()
         Builtins.y2milestone("User asks to switch to details")
+        @user_switched_to_release_notes = false
         @user_switched_to_details = true
         SwitchToDetailsView()
       elsif button == :showSlide && !ShowingSlide()
         if Slides.HaveSlides
+          if @user_switched_to_release_notes
+            # The user is switching from realease notes to slide show.
```
jreidinger: typo realease

library/packages/src/modules/SlideShow.rb
```diff
@@ -778,18 +780,29 @@ def HandleInput(button)
...
+            # In order to not disturbe the user while reading the release notes
```
jreidinger: typo disturbe

library/packages/src/modules/SlideShow.rb
```diff
@@ -778,18 +780,29 @@ def HandleInput(button)
...
+            # we are not updatding the tabs although the slide show has been
```
jreidinger: typo updatding

library/packages/src/modules/SlideShow.rb
```diff
@@ -778,18 +780,29 @@ def HandleInput(button)
...
+            # changed meanwhile. So we are updating it now before switching to
+            # slide show.
```
jreidinger: to be honest I do not get what this comment trying to say. User switched to release notes and now user click on show slide show. So it show release tab and then load slides? I probably missing something.

library/packages/src/modules/SlideShow.rb
```diff
@@ -139,6 +139,7 @@ def main
       @language = "en"
       @widgets_created = false
       @user_switched_to_details = false
+      @user_switched_to_release_notes = false
```
jreidinger: to be honest I do not like this "flags" they are exclusive ( details and release notes ) and they are also related.
What about having just one variable `@user_switched_to` with values `:details`, `:slides` and `release_notes`? so we know where user want to be?

library/packages/src/modules/SlideShow.rb
```diff
@@ -1029,6 +1042,7 @@ def GetSetup
     publish variable: :language, type: "string"
     publish variable: :widgets_created, type: "boolean"
     publish variable: :user_switched_to_details, type: "boolean"
+    publish variable: :user_switched_to_release_notes, type: "boolean"
```
jreidinger: Is it needed to have those variable publicly published? It looks to me like internal variable.

jreidinger: ah, I just found it is used in PackageSlideShow where is another logic when you change CD. :man_facepalming:

library/packages/test/slide_show_test.rb
```diff
@@ -146,4 +147,107 @@
...
+      it "switches to realease notes view" do
+        expect(Yast::SlideShow).to receive(:SwitchToReleaseNotesView)
+        Yast::SlideShow.HandleInput(:rn_Foo)
+      end
+      it "saves user selection" do
+        Yast::SlideShow.HandleInput(:rn_Foo)
+        expect(Yast::SlideShow.user_switched_to).to eq(:release_notes)
+      end
+    end
+  end
```
jreidinger: thanks for tests :+1:

package/yast2.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2
-Version:        4.2.53
+Version:        4.2.54
 Release:        0
 Summary:        YaST2 Main Package
 License:        GPL-2.0-only
```
jreidinger: as change is backward incompatble for packager you need conflict here with older packager. Otherwise `zypper up yast2` can result in crash as old packager willbe used.

PR #1009, opened 2020-01-13
added textdomains

schubi2:


library/packages/test/slide_show_test.rb
```diff
@@ -150,6 +150,7 @@
 
   describe "#Redraw" do
     before do
+      textdomain "base"
```
jreidinger: why it is needed? It should be not needed as there is not localized string ( and should not in test ).

schubi2: RuntimeError:
No textdomain configured in Yast::SlideShowClass, cannot translate "%s Release Notes"
in OSC. Up to now I have not found the right way.....

jreidinger: this is for sure wrong, as it is already in main https://github.com/yast/yast-yast2/blob/master/library/packages/src/modules/SlideShow.rb#L123

jreidinger: let me check it.

jreidinger: can you point me to log where it failed? or can I reproduce it locally in your branch with rake osc:build?

PR #1010, opened 2020-01-14
New classes for storing the old repositories (bsc#1159433)

lslezak:
## Details

- Related to https://bugzilla.suse.com/show_bug.cgi?id=1159433
- We need to remember the original repositories just after mounting the partition selected to upgrade. This will allow us later to distinguish the new SP2 repositories from the old ones and do not display them in the "Previously used repositories" dialog.
- The old repository setup will be saved in the `yast2-update` package and used in `yast2-installation`, this just prepares the infrastructure for that.
- The second part of the move is in this PR: https://github.com/yast/yast-packager/pull/502

### Notes

- New classes:
  - The `OriginalRepositorySetup` is a singleton object holding the original configuration, then we can compare the original and the current repositories to find the old ones.
  - The new `Y2Packager::Service` is similar to the already existing `Y2Packager::Repository`
- Other updates:
  - The `Y2Packager::Repository` class has been moved from `yast2-packager` here and extended a bit (added `repo_alias` attribute, added `url=` and `delete!` methods)
  - Also it's unit test has been moved
- Moved files (do not need detailed review):
  - `library/packages/src/lib/y2packager/repository.rb`
  - `library/packages/test/repository_test.rb`
  - `library/packages/test/product_factory.rb`

### DIFF
--- a/library/packages/src/lib/y2packager/repository.rb
+++ b/library/packages/src/lib/y2packager/repository.rb
@@ -1,0 +1,257 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2017 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+require "uri"
+require "y2packager/product"
+require "y2packager/resolvable"
+
+module Y2Packager
+  # This class represents a libzypp repository
+  #
+  # It offers a simple API to list them, query basic attributes,
+  # find out the products they offer and enabling/disabling them.
+  #
+  # @example Get all repositories
+  #   all_repos = Y2Packager::Repository.all     #=> [#<Y2Packager::Repository>, ...]
+  #   enabled = Y2Packager::Repository.enabled   #=> [#<Y2Packager::Repository>]
+  #
+  # @example Get a repository using a repo ID
+  #   repo = Y2Packager::Repository.find(1) #=> #<Y2Packager::Repository>
+  #   repo.autorefresh?                   #=> true
+  #   repo.url                            #=> "http://download.opensuse.org..."
+  #
+  # @example Disabling a repository
+  #   repo = Y2Packager::Repository.find(1) #=> #<Y2Packager::Repository>
+  #   repo.enabled?                       #=> true
+  #   repo.disabled!
+  #   repo.enabled?                       #=> false
+  class Repository
+    Yast.import "Pkg"
+
+    # Repository schemes considered local (see #local?)
+    # https://github.com/openSUSE/libzypp/blob/a7a038aeda1ad6d9e441e7d3755612aa83320dce/zypp/Url.cc#L458
+    LOCAL_SCHEMES = [:cd, :dvd, :dir, :hd, :iso, :file].freeze
+
+    # @return [Fixnum] Repository ID
+    attr_reader :repo_id
+    # @return [String] Repository name
+    attr_reader :name
+    # @return [URI::Generic] Repository URL (expanded)
+    attr_reader :url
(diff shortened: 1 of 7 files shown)


library/packages/src/lib/y2packager/repository.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      @alias = repo_alias
+      @auto_refresh = auto_refresh
+      @enabled = enabled
+      @name = name
+      @raw_url = raw_url
+      @url = url
+    end
+
+    def find_all
```
jreidinger: why not `self.find_all?` I expect it class method and not instance one.

library/packages/src/lib/y2packager/service.rb
```diff
@@ -0,0 +1,65 @@
...
+      @enabled = enabled
+      @file = file
+      @name = name
+      @repos_to_disable = repos_to_disable
+      @repos_to_enable = repos_to_enable
+      @type = type
+      @url = url
+    end
+
+    def find_all
```
jreidinger: same here

library/packages/src/lib/y2packager/repository.rb
```diff
@@ -0,0 +1,66 @@
...
+  #
+  # @since 4.2.54
+  class Repository
+    include Yast::Logger
+
+    attr_reader :alias, :name, :url, :raw_url, :id, :product_dir, :priority,
+      :enabled, :auto_refresh, :keep_packages, :service
+
+    def initialize(repo_alias: nil, auto_refresh: nil, enabled: nil, name: nil,
+      url: nil, raw_url: nil)
```
jreidinger: why not having there id? and neither in find_all. There is reader, but nothing write it.

library/packages/src/lib/y2packager/repository.rb
```diff
@@ -0,0 +1,257 @@
...
+      @url = url
+      @raw_url = raw_url || url
+      @product_dir = product_dir
+    end
+
+    # Return repository scheme
+    #
+    # The scheme is determined using the URL
+    #
+    # @return [Symbol,nil] URL scheme
```
jreidinger: Please document what means nil here.

library/packages/src/lib/y2packager/repository.rb
```diff
@@ -0,0 +1,257 @@
...
+      url.scheme ? url.scheme.to_sym : nil
+    end
+
+    # Return products contained in the repository
+    #
+    # @return [Array<Y2Packager::Product>] Products in the repository
+    #
+    # @see Y2Packager::Product
+    def products
+      return @products unless @products.nil?
```
jreidinger: it can be easier `return @products if @products`

library/packages/src/lib/y2packager/repository.rb
```diff
@@ -0,0 +1,257 @@
...
+    #
+    # @see #products
+    def addons
+      products.select { |p| p.category == :addon }
+    end
+
+    # Enable the repository
+    #
+    # The repository status will be stored only in memory. Calling to
+    # Yast::Pkg.SourceSaveAll will make it persistent.
```
jreidinger: @return is not documented even when it can return false.

library/packages/src/lib/y2packager/repository.rb
```diff
@@ -0,0 +1,257 @@
...
+    # Disable the repository
+    #
+    # The repository status will be stored only in memory. Calling to
+    # Yast::Pkg.SourceSaveAll will make it persistent.
+    #
+    # @see Yast::Pkg.SourceSetEnabled
+    # @see Yast::Pkg.SourceSaveAll
+    def disable!
+      success = Yast::Pkg.SourceSetEnabled(repo_id, false)
+      success && self.enabled = false
```
jreidinger: so this always return false, even when failed?

library/packages/src/lib/y2packager/repository.rb
```diff
@@ -0,0 +1,257 @@
...
+
+    # Remove the repository
+    #
+    # The repository will be removed only in memory. Calling to
+    # Yast::Pkg.SourceSaveAll will make the removal persistent.
+    #
+    # @see Yast::Pkg.SourceDelete
+    # @see Yast::Pkg.SourceSaveAll
+    def delete!
+      @repo_id = nil if Yast::Pkg.SourceDelete(repo_id)
```
jreidinger: error handling?

PR #1011, opened 2020-01-14
Add a new text helper to strip HTML tags

dgdavid:
## Problem

While updating the [addons selection dialog](https://github.com/yast/yast-registration/blob/master/src/lib/registration/ui/addon_selection_registration_dialog.rb) to make use of the new [`CustomStatusItemSelector`](https://github.com/libyui/libyui/pull/152) widget ([bsc#1157780](https://bugzilla.suse.com/show_bug.cgi?id=1157780 )), it was found that their descriptions contain HTML tags, which are

* not compatible with the `#wrap_text` helper, which could break the tags and/or ends with lines of different widths.
* not fully supported by the `CustomStatusItemSelector` since in ncurses they are not _rendered_ but displayed as part of the description.

## Solution

The simple solution would strip the HTML tags before wrapping the text. 

However, before realizing that the HTML was not being rendered in ncurses, the approach was only to replace some of them by line breaks. That's the reason why this new helper exists now.

Additionally, as a result of the first PR review made by @jreidinger, the solution also includes

* [Accepting a block for this new helper](https://github.com/yast/yast-yast2/pull/1011/commits/a53773a067dbd3e9e8608e72b9906103bbd273bf), which make it more flexible and powerful
* [Converting all text helpers as refinements of the `String` class](https://github.com/yast/yast-yast2/pull/1011/commits/1da6752238f2c66e7401b39e23503d0162e797d7), which makes easier to use them (of course, keeping backward compatibility)

## Test

* Unit tests added (refinements are tested via their _equivalent_ helper methods)
* Also tested manually

##  Notes

[YARD](https://yardoc.org/features.html) does not generate the documentation for refinements.There is a workaround, but it doesn't like to me. See https://github.com/lsegal/yard/issues/780

### DIFF
--- a/library/general/src/lib/ui/text_helpers.rb
+++ b/library/general/src/lib/ui/text_helpers.rb
@@ -17,160 +17,31 @@
 # current contact information at www.novell.com.
 # ------------------------------------------------------------------------------
 
+require "yast2/refinements/string_manipulations"
+
 module UI
   # Provides a set of methods to manipulate and transform UI text
   module TextHelpers
-    # Default HTML tags replacements
-    #
-    # @see #plain_text
-    # @return [Hash<String, String>]
-    DEFAULT_HTML_TAGS_REPLACEMENTS = {
-      "<br>"  => "\n",
-      "<br/>" => "\n",
-      "<br >" => "\n",
-      "</li>" => "\n",
-      "<ol>"  => "\n\n",
-      "<ul>"  => "\n\n",
-      "</p>"  => "\n\n"
-    }.freeze
+    using ::Yast2::Refinements::StringManipulations
 
-    # Get a new text version after ridding of HTML tags
-    #
-    # By default, a fully plain text will be returned, since some tags (see
-    # {DEFAULT_HTML_TAGS_REPLACEMENTS}) are going to be replaced by line breaks while the rest of
-    # them will be removed.
-    #
-    # The _tags_ param allows specifying which tags should be replaced or removed (keeping the rest
-    # of them untouched).
-    #
-    # Replacements can be provided
-    #   * as  tag => replacement collection via _replacements_ param, or
-    #   * using a block to perform desired changes for each match (more flexible and powerful)
-    #
-    # Note that all matched tags without a replacement are going to be replaced with `nil`. In other
-    # words, they will be deleted.
-    #
-    # @example Remove all HTML tags
-    #   text = "<p>YaST:</p><p>a <b>powerful</b> installation and configuration tool.</p>"
-    #   plain_text(text) #=> "YaST:\n\na powerful installation and configuration tool."
-    #
-    # @example Removing only the <p> tags
-    #   text = "<p>YaST:</p><p>a <b>powerful</b> installation and configuration tool.</p>"
-    #   plain_text(text, tags: ["p"])
-    #   #=> "YaST:\n\na <b>powerful</b> installation and configuration tool."
-    #
-    # @example Using custom replacements via _replacements_ param
-    #   text = "<p>YaST:</p><p>a <b>powerful</b> installation and configuration tool.</p>"
-    #   plain_text(text, replacements: { "<p>" => "\n- ", "<b>" => "*", "</b>" => "*" })
-    #     #=> "- YaST:\n- a *powerful* installation and configuration tool."
-    #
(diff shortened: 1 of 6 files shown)


library/general/src/lib/ui/text_helpers.rb
```diff
@@ -20,6 +20,59 @@
...
+    #     #=> "YaST:\n\na <b>powerful</b> installation and configuration tool."
+    #
+    # @example Using custom replacements
+    #   text = "<p>YaST:</p><p>a <b>powerful</b> installation and configuration tool.</p>"
+    #   plain_text(text, replacements: { "<p>" => "\n- ", "<b>" => "*", "</b>" => "*" })
+    #     #=> "- YaST:\n- a *powerful* installation and configuration tool."
+    #
+    # @param text [String] text to be processed
+    # @param tags [Array<String>] specific tags to be replaced or deleted
+    # @param replacements [Hash<String, String>] the replacements collection, tag => replacement
```
jreidinger: what about using block version instead of hash parameter? it is more flexible. so then you have something like
```
block ? block.call(match.to_s.downcase) : replacements[match.to_s.downcase]
```

jreidinger: so it is more like gsub API and definitivelly more flexible one if you want to use some kind of state machine like replacing li by number or unicode dot depending if it is ol or ul.

library/general/src/lib/ui/text_helpers.rb
```diff
@@ -20,6 +20,59 @@
...
+    #
+    # @return [String] the new version after ridding of undesired tags.
+    def plain_text(text, tags: nil, replacements: nil)
+      regex = tags ? Regexp.union(tags.map { |t| /<[^>]*#{t}[^>]*>/i }) : /<.+?>/
+      replacements ||= DEFAULT_HTML_TAGS_REPLACEMENTS
+
+      text.gsub(regex) { |match| replacements[match.to_s.downcase] }.strip
+    end
+
```
jreidinger: BTW I think that it would be nice to create refinement for plain_text and wrap_text so you can have really `string_a.plain_text.wrap_text` which looks more nature.

library/general/src/lib/ui/text_helpers.rb
dgdavid: Sorry, I'm not sure if I got what you mean.

Could you please elaborate it a little bit more?

library/general/src/lib/ui/text_helpers.rb
dgdavid: Definitely, that's a great idea. Let's try it.

However, we should keep the current #wrap_text version since it has been used in several places.

jreidinger: yep, of course backward compatibility is needed.

library/general/src/lib/ui/text_helpers.rb
jreidinger: well, what I mean is that instead of having replacements as hash parameter, you can accept block in same way like gsub allows it ( or allow both same as gsub ).
so something like
```
def plain_text(text, tags: nil, replacements: nil, &block)
....
text.gsub(regex) do |match|
  block ? block.call(match.to_s.downcase) : replacements[match.to_s.downcase]
end
```

This way, you can have something like
```
  text.plain_text do |param|
    case param
    when "<ol>" 
      @ordered = true
       index = 0
    when "<ul>" then @ordered = false
    when "<li>" then @ordered ? "#{index += 1}. " : "•"
    end
  end
```

which also create nicer plain text lists.

library/general/src/lib/ui/text_helpers.rb
dgdavid: @jreidinger 

Refinements have been added. See https://github.com/yast/yast-yast2/pull/1011/commits/d43265799d1b4231f6c1a2a60f56745e31046240

However, below problems were found

* [YARD](https://yardoc.org/features.html) does not generate the documentation. See this old and closed [issue](https://github.com/lsegal/yard/issues/780). I did tests and the workaround kind of work, but the file looks weird.
* I don't know what to do with tests. Those refinements are already tested via the helpers.
* I'm not sure if the _library/general/src/lib/yast2/refinements/_ was a good choice.

Could you give me a hand, please?

Thanks a lot!

PR #1013, opened 2020-01-17
 [SLE-9109] Packages online search

imobachgs:
Please, see the whole picture in https://github.com/yast/yast-registration/pull/467.

This PR adds support to `PackagesUI.RunPackageSelector` to enable the online search feature in the UI.


library/packages/src/modules/PackagesUI.rb
```diff
@@ -299,6 +300,8 @@ def RunPackageSelector(options)
 
       widget_options = Builtins.add(widget_options, :confirmUnsupported) if !display_support_status.nil? && display_support_status
 
+      widget_options = Builtins.add(widget_options, :onlineSearch) if options.fetch("enable_online_search", false)
```
lslezak: Why do we need `fetch` here? Would just a simple `if options["enable_online_search"]` be enough?

PR #1014, opened 2020-01-21
Cli read only

jreidinger:
trello: https://trello.com/c/SHhhC8TH/1598-8-wrong-error-codes-in-cli

goal: add option readonly to actions. If specified then finish action is skipped ( as finish is almost always write of modification ). Code is not straightforward due to interactive mode ( anyone using it? :man_facepalming: )

Documentation written to yardoc. This is how it is rendered:

Snímek obrazovky_2020-01-22_16-08-17


library/commandline/src/modules/CommandLine.rb
```diff
@@ -49,33 +51,38 @@ def main
...
+            "readonly" => true
           },
           "interactive" => {
             # translators: help for 'interactive' option on command line
-            "help" => _(
+            "help"     => _(
               "Start interactive shell to control the module"
-            )
+            ),
+            "readonly" => true
```
mvidner: Please comment why `interactive` is read-only, and why `abort` is not. I would think they are the opposite.

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1477,6 +1441,7 @@ def Run(commandline)
       return !Aborted() if !Init(commandline, WFM.Args)
 
       ret = true
+      read_only = true # do finish action only if module is not read only
```
mvidner: This seems to mean that the default is read-only, which will IMHO result in bug reports "why is my config not changed?!".

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1556,6 +1521,7 @@ def Run(commandline)
           # there is a handler, execute the action
           if !exec.nil?
             res = exec.call(options)
+            read_only = false unless commandline["actions"][command]["readonly"]
```
mvidner: This means that a missing value (-> `nil`) is the same as `false`. Sounds right but why default to `true` above?

library/commandline/src/modules/CommandLine.rb
jreidinger: well, reasoning is if you start interactive mode it does not mean you do operation that need write. E.g. simple interactive and then exit it is readonly.

For abort it can be read-only, but as it is abort it always skip write. So I can add it if you think it increase readability.

library/commandline/src/modules/CommandLine.rb
jreidinger: no, default for any operation that does not define key is write. So this is default if there is no operation, it is read_only and if there is operation that does not have readonly flag set to true, it is set to false.

library/commandline/src/modules/CommandLine.rb
jreidinger: as I said, if there is no operation, then it is readonly, but I am not sure if it can happen. And also it is easier logic for me in interactive mode, otherwise I need tri-state - not set, read_only commands only and at least one write command.

jreidinger:
@mvidner regarding documentation it is very good question. Problem is that there is no nicely suitable place. documentation is xml, examples does not contain init and finish operation ( examples is quite obsolete and also wrong regarding write ).

So I see two options:

1. I can add documentation to module header as yardoc
2. I can convert old xml documentation to markdown

and in both case add note about readonly flag.

library/commandline/MAINTAINER
```diff
@@ -1 +0,0 @@
-Ladislav Slezak <lslezak@suse.cz>
```
lslezak: Thanks for removing this! :smiley:

library/commandline/src/modules/CommandLine.rb
lslezak: Yes, a comment would be nice. It's pretty strange, this seems as if in the interactive mode you cannot change anything...

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1391,69 +1390,34 @@ def Done
...
-      unique_options = deep_copy(unique_options)
+      return nil if options.nil? || unique_options.nil?
+
       # sanity check
-      if Builtins.size(unique_options) == 0
-        Builtins.y2error(
-          "Unique test of options required, but the list of the possible options is empty"
-        )
+      if unique_options.empty?
+        log.error "Unique test of options required, but the list of the possible options is empty"
```
lslezak: Typo? 

```suggestion
        log.error "Unique list of options required, but the list of the possible options is empty"
```

mvidner:
> documentation is xml,

OMG, now I remember, it's the docbook [CommandLine-section.xml](https://github.com/yast/yast-yast2/blob/master/library/commandline/doc/CommandLine-section.xml) which is rendered at [doc.o.o](https://doc.opensuse.org/projects/YaST/openSUSE11.3/tdg/Commandline-library.html), 9 years ago. What a perfectly preserved specimen of a skeleton in a closet!

> So I see two options:
> 
> 1. I can add documentation to module header as yardoc
> 2. I can convert old xml documentation to markdown

I think having the docs close to the code is preferred, so YARD is better if you can make it work.

library/commandline/src/modules/CommandLine.rb
mvidner: No operation? But there is always an operation, otherwise I get the GUI!?

library/commandline/src/modules/CommandLine.rb
mvidner: > well, reasoning is if you start interactive mode it does not mean you do operation that need write. E.g. simple interactive and then exit it is readonly.

Well, the question is "How do you describe what does readonly mean". I thought it means "always skip the writing step" which would break the interactive mode.

The interactive mode has `exit` (with writing) and `abort` (without writing). I am still confused.

> For abort it can be read-only, but as it is abort it always skip write. So I can add it if you think it increase readability.

I see. What about this comment then:

```rb
"abort"       => {
  # aborting is handled before writing, so "readonly" does not matter here
```

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1427,9 +1430,53 @@ def UniqueOption(options, unique_options)
...
+    #        action without readonly is called, it will run finish handler.
+    #        Default value is `false`.
+    #
+    # @option commandline [Hash<String, Object>] "options" definition of options. Hash has action
+    #   name as key and value is hash with following keys:
+    #
+    #      - **"help"** _String|Array<String>_ mandatory action specific help text.
+    #        If array is passed it will be printed indended on multiple lines.
+    #      - **"type"** _String_ optional type check for option parameter. By default no
+    #        type checking is done. It abort if no checking is done and value is passed on CLI.
```
lslezak:
```suggestion
    #        type checking is done. It aborts if no checking is done and a value is passed on CLI.
```

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1427,9 +1430,53 @@ def UniqueOption(options, unique_options)
...
+    #        Possible values are ycp types and additionally enum and regex. For enum additional
+    #        key **"typespec"** with array of values have to be specified. For regex additional
+    #        key **"typespec"** with string containing ycp regexp is required. For integer it
+    #        do conversion of string value to integer value.
```
lslezak:
```suggestion
    #        does conversion of a string value to an integer value.
```

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1427,9 +1430,53 @@ def UniqueOption(options, unique_options)
...
+    #      - **"typespec"** _Object_ additional type specification. See **"type"** for details.
+    #
+    # @option commandline [Hash<String, Array<String>>] "mappings" defines connection between
+    #   **"actions"** and its **"options"**. Key is action and value is list of options it
```
lslezak:
```suggestion
    #   **"actions"** and its **"options"**. The key is action and the value is a list of options it
```

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1427,9 +1430,53 @@ def UniqueOption(options, unique_options)
...
+    #   supports.
+    # @return [Object] false if there was an error or no changes to be written (for example "help").
```
lslezak:
```suggestion
    # @return [Object] false if there was an error or there are no changes to be written (for example "help").
```

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1427,9 +1430,53 @@ def UniqueOption(options, unique_options)
...
     #      true if the changes should be written, or a value returned by the
-    #      handler
+    #      handler. Actions that are read-only returns also true if succeed even if nothing to be written.
```
lslezak:
```suggestion
    #      handler. Actions that are read-only return also true on success even if there is nothing to write.
```

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1441,7 +1488,8 @@ def Run(commandline)
       return !Aborted() if !Init(commandline, WFM.Args)
 
       ret = true
-      read_only = true # do finish action only if module is not read only
+      # no action is readonly. But first module without readonly switch flag to false
```
lslezak:
```suggestion
      # no action is readonly, but the first module without "readonly" will switch the flag to `false`
```

library/commandline/src/modules/CommandLine.rb
```diff
@@ -1521,6 +1569,7 @@ def Run(commandline)
           # there is a handler, execute the action
           if !exec.nil?
             res = exec.call(options)
+            # unless action explicitelly mention that it is readonly run finish handler.
```
lslezak:
```suggestion
            # unless an action explicitly mentions that it is read-only it will run the finish handler
```

PR #1018, opened 2020-01-29
Speedup wsl

jreidinger:
during manual testing I see speed up from 16 seconds to less then 1 second.

trello: https://trello.com/c/tCNqpRaT/1594-8-research-ostumbleweed-p5-1157575-wsl-creating-user-in-firstboot-does-not-show-progess-bar


package/yast2.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Jan 29 13:22:50 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- Optimize run on WSL (bsc#1157575)
```
ancorgs: "Optimize" is a big optimistic. Let's say "Speed up". :wink:

library/systemd/src/lib/yast2/systemctl.rb
```diff
@@ -82,12 +84,16 @@ def target_units
     private
 
       def list_unit_files(type: nil)
+        return "" unless Yast::Systemd.Running
```
ancorgs: Maybe we should write something in y2logs like "Systemd is not running: reporting empty list of unit files"

jreidinger: makes sense for me. I just hope it is not called often

PR #1020, opened 2020-01-30
add ability for WSL desktop file whitelisting

jreidinger:
reason: WSL is limited environment, so many modules does not make sense to show. So whitelisting only ones that has specific key in desktop file.

### DIFF
--- a/library/desktop/test/desktop_test.rb
+++ b/library/desktop/test/desktop_test.rb
@@ -34,9 +34,59 @@ describe Yast::Desktop do
         "add-on"           => { "Name" => "YaST Add-On Products" },
         "lan"              => { "Name" => "YaST Network" },
         "services-manager" => { "Name" => "YaST Services Manager" },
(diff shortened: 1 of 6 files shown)


library/desktop/test/desktop_test.rb
```diff
@@ -34,9 +34,58 @@
         "add-on"           => { "Name" => "YaST Add-On Products" },
         "lan"              => { "Name" => "YaST Network" },
         "services-manager" => { "Name" => "YaST Services Manager" },
+        "sw-single"        => { "Name"=>"YaST Software Management" },
```
joseivanlopez: NP: missing spaces around `=>`

joseivanlopez:
Ups, there are conflicts. Please, fix them ;)

jreidinger:
@joseivanlopez fixed, I worry no one is familiar with this piece of code as it is ncurses only control center + desktop parses which is also not used often.

PR #1022, opened 2020-01-31
Add ProgressBar widget

joseivanlopez:
### Problem

The Expert Partitioner needs a new dialog to show the progress when applying changes. This new dialog requires a progress bar.

* https://bugzilla.suse.com/show_bug.cgi?id=1135366

### Solution

Add new classes `CWM::ProgresBar` and `CWM::DynamicProgressBar`.

### Testing

* Manually tested
* Added unit tests.

### Screenshots

<details>
<summary>Show/Hide</summary>

Screenshot from 2020-01-31 16-13-35

</details>

### DIFF
--- a/library/cwm/src/lib/cwm/dynamic_progress_bar.rb
+++ b/library/cwm/src/lib/cwm/dynamic_progress_bar.rb
@@ -21,7 +21,10 @@ require "abstract_method"
 require "cwm/progress_bar"
 
 module CWM
-  # Widget for a dynamic progress bar, where the label can be set for every step
+  # Widget for a dynamic progress bar, where the label can be set for every step.
+  #
+  # This progress bar is useful when steps are not known in advance or part of them are dynamically
+  # generated.
   #
   # @example
   #
@@ -50,8 +53,6 @@ module CWM
       super()
     end
 
-  private
-
     # @!method label
     #
     #   Label for the progress bar when no step is given, see {#forward}.
@@ -65,6 +66,8 @@ module CWM
     #
     #   @return [Integer]
     abstract_method :steps_count
+
+  private
 
     # @see ProgressBar#steps
     def steps
--- a/library/cwm/src/lib/cwm/dynamic_progress_bar.rb
+++ b/library/cwm/src/lib/cwm/dynamic_progress_bar.rb
@@ -1,0 +1,83 @@
+# Copyright (c) [2020] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "abstract_method"
+require "cwm/progress_bar"
+
+module CWM
+  # Widget for a dynamic progress bar, where the label can be set for every step
+  #
+  # @example
+  #
+  #   class MyProgressBar < CWM::DynamicProgessBar
+  #     def steps_count
+  #       3
+  #     end
+  #
+  #     def label
+  #       "Progress"
+  #     end
+  #   end
+  #
+  #   pg = MyProgressBar.new
+  #
+  #   pg.forward("step 1") #=> shows label "step 1"
+  #   pg.forward           #=> shows label "Progress"
+  #   pg.forward("step 3") #=> shows label "step 3"
+  class DynamicProgressBar < ProgressBar
+    # Moves the progress forward and sets the given step as label
+    #
+    # @see ProgressBar#forward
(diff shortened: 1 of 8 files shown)


library/cwm/src/lib/cwm/progress_bar.rb
```diff
@@ -0,0 +1,207 @@
...
+    # @param id [String] widget id; auto-generated if nothing is indicated.
+    # @param label [String] label for the progress bar. This is ignored in case that steps are known (see
+    #   steps parameter) and the steps are configured to be shown (see show_step parameter).
+    # @param total_steps [Integer] total number of steps that the progress bar can perform. This is
+    #   ignored in case that steps are know (see steps parameter).
+    # @param steps [Array<String>] list of steps. They will be used as label of the progress bar in case
+    #   steps are configured to be shown (see show_step parameter).
+    # @param show_step [Boolean] whether the current step should be used as label for the progreess bar.
+    #   This is ignored if the steps are unknown (see steps parameter).
+    def initialize(id: nil, label: nil, total_steps: 0, steps: [], show_step: true)
```
jreidinger: well, API style here is kind of different to CWM style. Here it is more like instantiate object and calling actions there. CWM way is more like specialized object ( usually child ) and then calling actions on it.

Difference is constructor. Usual CWM constructor is almost empty and just depends on it defined methods.

What I also do not like at this code is unnecessary complexity that mixture two styles. One is defined steps and one is undefined steps. So what it can look like?
There can be two classes one is ProgressBar which subclass has to define steps method and optional show_step method. And its subclass which can be called DynamicProgressBar, that require label method and steps_count and define steps as label * steps_count.
This will also solve all issues with id that need not to be autogenerated. So in the end it can be empty constructor and two classes. Where you pick behavior by class and not by combination of parameters in constructor.

library/cwm/test/progress_bar_test.rb
```diff
@@ -0,0 +1,211 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require_relative "test_helper"
```
jreidinger: can you also extend cwm/rspec with helper for progress bar widget tests?

library/cwm/src/lib/cwm/progress_bar.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+  private
+
+    # @!method steps
+    #
+    #   Steps for the progress bar
+    #
+    #   @return [Array<String>]
+    abstract_method :steps
```
jreidinger: do you check if they are preset in yardoc output? I am not sure as we usually do not generate doc for private methods.

library/cwm/src/lib/cwm/progress_bar.rb
```diff
@@ -0,0 +1,118 @@
...
+      steps.size
+    end
+
+    # Label to use for the current step
+    #
+    # @return [String]
+    def current_label
+      label = show_steps? ? current_step : nil
+
+      label || ""
```
jreidinger: so if I get it, if show_steps is set to false, it shows only "" always as label? Is there use case for it? How it will look?

library/cwm/src/lib/cwm/rspec.rb
```diff
@@ -225,3 +244,30 @@
...
   end
 end
+
+RSpec.shared_examples "CWM::ProgressBar" do
+  include_examples "CWM::CustomWidget"
+
+  describe "#steps" do
+    it "produces an Array of String" do
+      expect(subject.send(:steps)).to be_an Array
+      expect(subject.send(:steps)).to all(be_a(String))
```
jreidinger: :+1:

library/cwm/src/lib/cwm/progress_bar.rb
joseivanlopez: For the case when you want to show a progress bar without label. In fact, that is the case (for now) in the new dialog of the Expert Partitioner.

library/cwm/src/lib/cwm/dynamic_progress_bar.rb
```diff
@@ -0,0 +1,83 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "abstract_method"
+require "cwm/progress_bar"
+
+module CWM
+  # Widget for a dynamic progress bar, where the label can be set for every step
```
jreidinger: I would probably mention here where it can be useful. For me it is something like. `Useful when steps are not known in advance or part of step is dynamically generated.` Do I get it right?

joseivanlopez: Yes, that is.

library/cwm/src/lib/cwm/progress_bar.rb
joseivanlopez: See the screenshot in the description of this PR.

jreidinger: OK, makes sense for me. Maybe add here to yardoc `see {#show_steps?}` where it is also documented.

library/cwm/src/lib/cwm/progress_bar.rb
joseivanlopez: Nop, it is not generated. I will declare them as protected.

PR #1024, opened 2020-02-24
Reinitializing Augeas parser again before writing data to

schubi2:
/etc/sysctl.d/70-yast.conf (bsc#1151649).

### DIFF
--- a/library/general/src/lib/cfa/sysctl_config.rb
+++ b/library/general/src/lib/cfa/sysctl_config.rb
@@ -91,6 +91,11 @@ module CFA
 
     def load
       files.each(&:load)
+      # It seems that Augeas caches the data structure of last loaded
+      # file (/etc/sysctl.conf) which will not fit together with the
+      # data structure of /etc/sysctl.d/70-yast.conf which will be written
+      # in the save call. Therefore we are loading this data structure again:
+      yast_config_file&.load
     end
 
     # Saving all sysctl settings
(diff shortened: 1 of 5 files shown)


library/general/src/lib/cfa/sysctl_config.rb
```diff
@@ -91,6 +91,11 @@ def define_attr(attr)
 
     def load
       files.each(&:load)
+      # It seems that Augeas caches the data structure of last loaded
+      # file (/etc/sysctl.conf) which will not fit together with the
+      # data structure of /etc/sysctl.d/70-yast.conf which will be written
+      # in the save call. Therefore we are loading this data structure again:
+      yast_config_file&.load
```
jreidinger: this is caused by cache in parser. If you want to avoid it or remember file content for each file, use separate parser for each file. So proper fix is in fact change https://github.com/yast/yast-yast2/blob/372695e13437df4f94136ff4028da85837b2053f/library/general/src/lib/cfa/sysctl.rb#L55 to not be constant but instead create always new parser instance at https://github.com/yast/yast-yast2/blob/372695e13437df4f94136ff4028da85837b2053f/library/general/src/lib/cfa/sysctl.rb#L119

And of course having unit tests for this would be really nice.

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Feb 24 15:02:15 CET 2020 - schubi@suse.de
+
+- Reinitializing Augeas parser again before writing data to
```
lslezak: It seems that the fix is now rather "create an own Augeas instance" than "reinitializing"...

PR #1025, opened 2020-02-26
Updated docu for SysctlConfig class

schubi2:


library/general/src/lib/cfa/sysctl_config.rb
```diff
@@ -43,17 +43,7 @@ module CFA
...
-  # setting +forward_ipv4+ to true. However, under the hood, the kernel maps
-  # boolean values to "1" or "0". If you want to access to that raw value,
-  # you can prepend "raw_" to the method's name.
-  #
-  # @example Accessing the raw value of a setting
-  #   sysctl = SysctlConfig.new
-  #   sysctl.load
-  #   sysctl.raw_forward_ipv6 #=> "0"
-  #   sysctl.raw_forward_ipv6 = "1"
+
```
teclator:
```suggestion
```

package/yast2.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Feb 26 10:42:42 CET 2020 - schubi@suse.de
+
+- Updated docu for SysctlConfig class (bsc#1151649).
```
teclator: It is only about documentation, that is why I commented it could be done next time the class was touched or the package, thus, I think that we can omit the version bump

teclator: https://github.com/yast/yast-network/pull/1041#pullrequestreview-364702028

PR #1026, opened 2020-03-05
Do not crash when loading /usr/etc/services

dgdavid:
## Problem

`YaST::PortAliases` gets the list of aliases by directly reading `/etc/services`, crashing when running in a system using the new configuration files layout (i.e., the file is available at `/usr/etc/services` instead)

* https://bugzilla.suse.com/show_bug.cgi?id=1162853
* https://trello.com/c/eDSVnfOz/

## Solution

To adapt mentioned YaST module to make use of [`getent`](http://man7.org/linux/man-pages/man1/getent.1.html) output instead of reading the file directly, since it will read the right file: `/usr/etc/services` or `/etc/services`, taking precedence the first one when both are present.

## Test

* Added unit tests
* Also tested manually via YaST -> ftp-server -> Expert Settings -> [x] Open port in Firewall -> Finish.

## Notes

In spite of this PR includes a big amount of changes, there is room for improvements yet. However, IMHO it doesn't pay off to going beyond sent changes without a full refactorization in mind, which would include breaking api changes.

---

Readings related to the new configuration files layout:

* https://github.com/thkukuk/atomic-updates_and_etc/
* https://gist.github.com/joseivanlopez/bcf7185904922afae7d93f7518192864


library/network/src/modules/PortAliases.rb
```diff
@@ -1,137 +1,93 @@
-# ***************************************************************************
+# Copyright (c) [2013-2020] SUSE LLC
 #
-# Copyright (c) 2002 - 2012 Novell, Inc.
```
dgdavid: To be honest, I'm not sure about which year choose. 2002? 2013?

library/network/src/modules/PortAliases.rb
```diff
@@ -1,137 +1,93 @@
...
-#
-# ***************************************************************************
-# File:  modules/PortAliases.ycp
-# Package:  Ports Aliases.
-# Summary:  Definition of Port Aliases.
-# Authors:  Lukas Ocilka <locilka@suse.cz>
-#
-# $Id$
-#
-# Global Definition of Port Aliases for services taken from /etc/services.
```
dgdavid: It's weird because the documentation said that, but actually there are some services _pre-loaded_. I have preferred to be conservative and keep them as _KNOWN_SERVICES_, but seems that they are not needed at all.

:question:

library/network/src/modules/PortAliases.rb
```diff
@@ -1,137 +1,93 @@
...
+require "yast2/execute"
 require "shellwords"
 
 module Yast
+  # Module providing full listing of port aliases for services, supporting also multiple aliases for
+  # the same port (like "http", "www", "www-http" for port 80)
   class PortAliasesClass < Module
     include Yast::Logger
 
+    KNOWN_SERVICES = {
```
dgdavid: See the comment above

library/network/src/modules/PortAliases.rb
```diff
@@ -141,175 +97,143 @@ def AllowedPortNameOrNumber
...
+    # EtherNet/IP-1         2222
+    # EtherNet-IP-2         44818
+    # rfb                   5900 vnc-server
+    #
+    # @return [Array<String>]
+    def load_services_database
+      Yast::Execute.stdout.on_target!(
+        ["/usr/bin/getent", "services"],
+        # remove the protocol from the output
+        ["/usr/bin/sed", "-r", "s/(\\S)(\\/\\S+)(.*)?/\\1\\3/"],
```
dgdavid: I know this looks a bit complicated. The intention here, as stated, is to filter out the _duplicated_ lines, where duplicate means "lines for the same service name, port and aliases but different protocol". I think this is faster than processing them in Ruby, but I'm not against to change it.

What do you think?

library/network/src/modules/PortAliases.rb
```diff
@@ -141,175 +97,143 @@ def AllowedPortNameOrNumber
...
-      command = "/usr/bin/grep \"^[^#].*[ \\t]#{port_number}/\" /etc/services " \
-        "| /usr/bin/sed \"s/\\([^ \\t]*\\)[ \\t]*.*/\\1/\""
-      found = Convert.to_map(SCR.Execute(path(".target.bash_output"), command))
-      aliases = []
+    # Returns list of aliases (including the port number) for service.
+    #
+    # NOTE: given argument will be also included
+    #
+    # @param needle [String] the name, alias or port to look for a service
+    # @return [Array<String>]
```
imobachgs: What about describing the returned value?

library/network/src/modules/PortAliases.rb
```diff
@@ -141,175 +97,143 @@ def AllowedPortNameOrNumber
...
 
-      # lets use the first found entry, as often there are two entries for tcp and udp
-      port = matching_lines.first[/^\S+\s+(\d+)/, 1]
-      if port.nil?
-        log.error "wrong line in /etc/services: #{matching_lines.first}"
-        return nil
-      end
+    # Returns the port for requested service (if any)
+    #
+    # NOTE: when given argument looks like a digit, it will be returned after a proper conversion
```
imobachgs: `@note` ?

library/network/src/modules/PortAliases.rb
```diff
@@ -141,175 +97,143 @@ def AllowedPortNameOrNumber
...
+    # removing the protocol.
+    #
+    # Returned lines will look like
+    #
+    # EtherNet/IP-1         2222
+    # EtherNet-IP-2         44818
+    # rfb                   5900 vnc-server
+    #
+    # @return [Array<String>]
+    def load_services_database
```
imobachgs: Why not just `services_database` (or something similar)? The `load_` part suggests that the information is being "loaded/cached/stored" in some place.

library/network/src/modules/PortAliases.rb
imobachgs: It looks quite unreadable to me :wink:. I would give it a try using Ruby as I do not expect a noticeable performance impact.

package/yast2.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Mar  5 13:00:43 UTC 2020 - David Diaz <dgonzalez@suse.com>
+
+- Do not crash when reading the services database in a system
```
imobachgs: The entry looks too wordy to me. What about something more concise like the following example?

*Read the list of network service properly, no matter where it is stored (bsc#1162853).*

library/network/src/modules/PortAliases.rb
dgdavid: In fact, it's being _loaded_ via `getent`

package/yast2.changes
dgdavid: Much better, thanks

library/network/src/modules/PortAliases.rb
dgdavid: I can confirm that in my system all of them are also in the `/usr/etc/services` database.

dgdavid: And finally I removed them. Please, have a look to https://github.com/yast/yast-yast2/pull/1026/commits/dd2118148c4fd03bb91721ccf25b9cc2fb825550

dgdavid:
@imobachgs the PR has been updated according to your code review suggestions. Additionally, I also [removed a not useful cache](https://github.com/yast/yast-yast2/pull/1026/commits/b53161d877d8f3987ce4da1bdc77dee25f3d7d00).

Please, re-review it :)

Thanks

dgdavid:
After the manual test, I have have realized that [Cheetah](https://github.com/openSUSE/cheetah/) is dumping in the logs the `getent services` output (11750 lines in my system). 

Screenshot_20200305_163601



I think this is not desired. Overriding the `logger` option to `nil` when calling to [YaST::Execute.on_target!](https://github.com/yast/yast-yast2/blob/b0eebf01c5485448393933563082cb908de0d3fc/library/system/src/lib/yast2/execute.rb#L82) solves _the problem_. Is it the right approach? Are we polluting the log files with the output of commands every time we use `Yast::Execute`?

@jreidinger, friendly ping :point_up:

library/network/src/modules/PortAliases.rb
```diff
@@ -141,175 +65,114 @@ def AllowedPortNameOrNumber
...
-        found_alias_port = Ops.get(@SERVICE_NAME_TO_PORT, port) do
-          LoadAndReturnNameToPort(port)
-        end
-        if !found_alias_port.nil?
-          service_aliases = Builtins.add(
-            service_aliases,
-            Builtins.tostring(found_alias_port)
-          )
+        result[port] ||= Service.new(port, [])
+        result[port].aliases |= aliases
```
imobachgs: Why not `||=` (for consistency)?

dgdavid: Because its an union. That code add the new aliases to the previous one.

Maybe using the _long_ syntax could help to avoid the confusion.

```ruby
result[port].aliases = result[port].aliases.union(aliases)
```

Although the naming is a bit confusing too.

imobachgs: Ah, ok, I got confused. Then it is OK.

PR #1027, opened 2020-03-06
Restore the CWM::RichText scroll after updating its content

dgdavid:
## Problem

After changing/updating its content (aka value), a CWM::RichText is not able to restore the vertical scroll. This is specially annoying in those scenarios where that widget is being used to emulate complex interfaces, such as MultiSelection list.

* https://bugzilla.suse.com/show_bug.cgi?id=1049965

## Solution

Add a flag to know if it will be necessary to save/restore the scroll before/after updating the value.

## Test

* Tested manually
* Added unit tests.

## Screencast

<details>
<summary>Click to show/hide</summary>

---

ezgif com-optimize(5)

</details>

---

Related to 

* https://github.com/libyui/libyui/pull/149
* https://gist.github.com/shundhammer/1bc79af7d1a31892bd977678bc2af2c1#conservative-approach-save-scrollbar-values
* https://github.com/yast/yast-registration/pull/483#discussion_r385282607

### DIFF
--- a/library/cwm/src/lib/cwm/common_widgets.rb
+++ b/library/cwm/src/lib/cwm/common_widgets.rb
@@ -368,6 +368,13 @@ module CWM
 
     # Determines if the vertical scroll must be kept after updating the content
     #
+    # @note Useful only to keep the sense of continuity when redrawing basically with the same text
+    #
+    # Keeping the vertical scroll after changing the value is mostly intended to be used after a
+    # redraw because of a user action. However, using it after changing the content noticeably
+    # (e.g., displaying different product descriptions), will look like a randomly positioned
+    # vertical scroll.
+    #
     # @return [Boolean] true if the vertical scroll must be kept; false otherwise
     def keep_scroll?
       false
--- a/library/cwm/src/lib/cwm/common_widgets.rb
+++ b/library/cwm/src/lib/cwm/common_widgets.rb
@@ -365,6 +365,40 @@ module CWM
     self.widget_type = :richtext
 
     include ValueBasedWidget
+
+    # Determines if the scroll should be restored after updating the content
+    #
+    # @return [Boolean] true if the scroll should be restored; false otherwise
+    def keep_scroll?
+      false
+    end
+
+    # Updates the content
+    #
+    # Depending on #keep_scroll?, the scroll will be saved and restored.
+    #
+    # @param val [String] the new content for the widget
+    def value=(val)
+      current_vscroll = vscroll
+      super
+      self.vscroll = current_vscroll if keep_scroll?
+    end
+
+  private
+
+    # Saves the current vertical scroll
+    #
+    # @return [String] current vertical scroll value
+    def vscroll
+      Yast::UI.QueryWidget(Id(widget_id), :VScrollValue)
+    end
+
+    # Sets vertical scroll
+    #
+    # @param value [String] the new vertical scroll value
+    def vscroll=(value)
+      Yast::UI.ChangeWidget(Id(widget_id), :VScrollValue, value)
+    end
   end
 
   # Time field widget
--- a/library/cwm/src/lib/cwm/common_widgets.rb
+++ b/library/cwm/src/lib/cwm/common_widgets.rb
@@ -366,16 +366,16 @@ module CWM
 
     include ValueBasedWidget
 
-    # Determines if the scroll should be restored after updating the content
+    # Determines if the vertical scroll must be kept after updating the content
     #
-    # @return [Boolean] true if the scroll should be restored; false otherwise
+    # @return [Boolean] true if the vertical scroll must be kept; false otherwise
     def keep_scroll?
       false
     end
 
     # Updates the content
     #
-    # Depending on #keep_scroll?, the scroll will be saved and restored.
+    # Depending on #keep_scroll?, the vertical scroll will be saved and restored.
     #
     # @param val [String] the new content for the widget
     def value=(val)
--- a/library/cwm/test/common_widgets_test.rb
+++ b/library/cwm/test/common_widgets_test.rb
@@ -6,7 +6,6 @@ require "cwm/common_widgets"
 require "cwm/rspec"
 
 describe CWM::RadioButtons do
-
   class TestRadioButtons < CWM::RadioButtons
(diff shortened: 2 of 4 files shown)


library/cwm/src/lib/cwm/common_widgets.rb
```diff
@@ -365,6 +365,40 @@ class RichText < AbstractWidget
     self.widget_type = :richtext
 
     include ValueBasedWidget
+
+    # Determines if the scroll should be restored after updating the content
+    #
+    # @return [Boolean] true if the scroll should be restored; false otherwise
+    def keep_scroll?
+      false
```
lslezak: Um, maybe the default value should be `true`. I think the more common case is to keep the scroll position after changing the RichText content.

What do you think?

dgdavid: I agree, the most common case is (or should be, IMO) to keep the scroll position. However, I used `false` as a default value because I'm not sure if changing the behavior will be break things.

So, probably we should collect more opinions before taking a final decision, shouldn't we?
Probably we should collect a few more opinions, to tak

library/cwm/test/common_widgets_test.rb
```diff
@@ -59,3 +58,47 @@ def hspacing
...
+        expect(Yast::UI).to receive(:ChangeWidget).with(widget_id, :VScrollValue, anything)
+
+        subject.value = "Test"
+      end
+    end
+
+    context "when set to not restore the scroll" do
+      let(:keep_scroll) { false }
+
+      it "saves the scroll position" do
```
jreidinger: is this important behavior? Why testing it as it is lost? I think we should not test this behavior as we can easily remove it in future.

library/cwm/test/common_widgets_test.rb
```diff
@@ -59,3 +58,47 @@ def hspacing
...
+        expect(Yast::UI).to receive(:QueryWidget).with(widget_id, :VScrollValue)
+
+        subject.value = "Test"
+      end
+
+      it "restores the scroll" do
+        expect(Yast::UI).to receive(:ChangeWidget).with(widget_id, :VScrollValue, anything)
+
+        subject.value = "Test"
+      end
```
jreidinger: well, it saves, but it cannot be used outside. So if we use rspec as behavior documentation that I expect something like `keeps scrolling values` but it is also not precise as we remember only vertical scrolling and not horizontal right?

And what is missing here is mentioning original behavior that is kept, so it sets value of widget.

library/cwm/src/lib/cwm/common_widgets.rb
```diff
@@ -365,6 +365,40 @@ class RichText < AbstractWidget
     self.widget_type = :richtext
 
     include ValueBasedWidget
+
+    # Determines if the scroll should be restored after updating the content
```
jreidinger: it is just vertical scrolling now.

library/cwm/test/common_widgets_test.rb
dgdavid: > vertical scrolling and not horizontal right?

Right. At some point, I had _vertical scrolling_ there, but moving/changing things around...

> And what is missing here is mentioning original behavior that is kept, so it sets value of widget.

Also true. But I was waiting to collect enough feedback to know what we want to do: preserve the original behavior or change it as @lslezak proposed.

library/cwm/src/lib/cwm/common_widgets.rb
dgdavid: I'll update those documentation lines. Thanks

library/cwm/test/common_widgets_test.rb
dgdavid: > is this important behavior? Why testing it as it is lost?

I don't understand these questions. What's the problem adding unit testing here?

>  I think we should not test this behavior as we can easily remove it in future.

I get confused. If we remove this option in the future (!?), we simply can remove the associated tests. Can't we?

jreidinger: > > is this important behavior? Why testing it as it is lost?
> 
> I don't understand these questions. What's the problem adding unit testing here?

Well, rspec idea is to tests specification. And saving scroll option is not visible from outside and it is also not interested for user (of API) as he cannot use it. So it makes more sense for me to have specification like `it "does not modify vertical scrolling" do`

> 
> > I think we should not test this behavior as we can easily remove it in future.
> 
> I get confused. If we remove this option in the future (!?), we simply can remove the associated tests. Can't we?

question is does behavior of method change if we no longer store scrolling? My issue here is that test is very tight with exact implementation and so very fragile. Ideally in rspec if you change implementation you should change test content, but not specification written in tests. That is difference between unit testing and behavior testing.

library/cwm/test/common_widgets_test.rb
jreidinger: > 
> > And what is missing here is mentioning original behavior that is kept, so it sets value of widget.
> 
> Also true. But I was waiting to collect enough feedback to know what we want to do: preserve the original behavior or change it as @lslezak proposed.

It is probably misunderstanding, I mean here that we do not test that we change content of richtext which is the most important behavior of value= method. And we should do it regardless if scroll saving is set or not.

library/cwm/test/common_widgets_test.rb
dgdavid: > I don't understand these questions. What's the problem adding unit testing here?

Ok, I think I got it now, after seeing the comment in the full context. You mean testing that the vscroll has been saved when it will be not used and it is not public.

I think it does not hurt, but I'll remove it and try to improve the other examples.

dgdavid: Ups, I hadn't seen your comment until refresh the page.

Thanks!

library/cwm/src/lib/cwm/common_widgets.rb
dgdavid: As discussed in the #yast channel at the freenode.net IRC server, we are to keep returning `false` as default because updating to a different content will produce not expected behavior. This _feature_ is intended to be used with the same or almost the same content, to give a sense of continuity when the widget is _redrawn_.

PR #1028, opened 2020-03-06
Skip repository reloading (bsc#1165501)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1165501 
- This is related to https://github.com/yast/yast-pkg-bindings/pull/125, restoring the sources during installation has a side effect of deselecting the selected products or patterns
- 4.2.68

### Tests
- [x] Tested manually in Online installation
- [x] Tested manually in Offline installation

### DIFF
--- a/library/packages/src/modules/Product.rb
+++ b/library/packages/src/modules/Product.rb
@@ -198,7 +198,7 @@ module Yast
       if use_installed_products?
         PackageSystem.EnsureTargetInit
       else
-        PackageSystem.EnsureSourceInit
+        PackageSystem.EnsureSourceInit unless Stage.initial
       end
 
       Pkg.PkgSolve(true)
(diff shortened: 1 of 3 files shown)


library/packages/src/modules/Product.rb
```diff
@@ -198,7 +198,7 @@ def load_zypp
       if use_installed_products?
```
teclator: live_installation mode is not used anymore, isn't it?, thus, this method will check if  `!Stage.initial` and in the else there is again a !Stage.initial, so looks like we could do something like 

```ruby
PackageSystem.EnsureTargetInit if Stage.Initial
```

lslezak: Lets keep it simple for the RC1...

PR #1029, opened 2020-03-11
[CWM] Add new widget to manage multi status selection lists

dgdavid:
## :construction:  Better PR description is WIP

> Providing following features:
>
>  - Allow to select more than one item.
>  - Able to represent multiple statuses: no selected, selected, auto selected.
>  - Able to represent disabled items.
>  - Emit different events to distinguish the interaction through check box input or its label.
>  - Automatic text wrapping.
>  - Keep the vertical scroll.

---

Related to 

* https://trello.com/c/CmaNrDAw
* https://github.com/yast/yast-registration/pull/483#discussion_r385108398

### DIFF
--- a/library/cwm/src/lib/cwm/multi_status_selector.rb
+++ b/library/cwm/src/lib/cwm/multi_status_selector.rb
@@ -34,6 +34,10 @@ module CWM
   #
   # If you want to see it in action, have a look at yast2-registration or yast2-packager modules.
   #
+  # TODO: make possible to use it more than once in the same dialog, maybe by using the parent
+  # widget_id as a prefix for the event_id. See {CWM::MultiStatusSelector#handle} and
+  # {CWM::MultiStatusSelector::Item.event_id}.
+  #
   # @example Defining a MultiStatusSelector to manage products selection (with dependencies)
   #
   #     require "cwm/multi_status_selector"
--- a/library/cwm/src/lib/cwm/multi_status_selector.rb
+++ b/library/cwm/src/lib/cwm/multi_status_selector.rb
@@ -175,7 +175,7 @@ module CWM
 
     # Updates the content based on items list
     def refresh
-      content.value = items.map(&:to_s).join("<br>")
+      content.value = items.map(&:to_richtext).join("<br>")
     end
 
     # Convenience widget to keep the content updated
--- a/library/cwm/src/lib/cwm/multi_status_selector.rb
+++ b/library/cwm/src/lib/cwm/multi_status_selector.rb
@@ -1,0 +1,470 @@
+# Copyright (c) [2020] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "cwm"
+require "abstract_method"
+
+module CWM
+  # Custom widget to manage multi status selection items
+  #
+  # It uses a RichText to emulate the multi selection list meeting following requirements:
+  #
+  #  - Allow to select more than one item.
+  #  - Able to represent multiple statuses: no selected, selected, auto selected.
+  #  - Items can be enable or disabled.
+  #  - Emit different events to distinguish the interaction through check box input or its label.
+  #  - Automatic text wrapping.
+  #  - Keep the vertical scroll.
+  #
+  # If you want to see it in action, have a look at yast2-registration or yast2-packager modules.
+  #
+  # @example Defining a MultiStatusSelector to manage products selection (with dependencies)
+  #
+  #     require "cwm/multi_status_selector"
+  #
+  #     class MyMultiStatusSelector < CWM::MultiStatusSelector
+  #       attr_reader :items
+  #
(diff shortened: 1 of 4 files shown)


library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -0,0 +1,470 @@
...
+      # Determines whether running in installation mode
+      #
+      # We do not use Stage.initial because of firstboot, which runs in 'installation' mode
+      # but in 'firstboot' stage.
+      #
+      # @return [Boolean] Boolean if running in installation or update mode
+      def installation?
+        Yast::Mode.installation || Yast::Mode.update
+      end
+    end
```
dgdavid: Question: could we simply use a `class` attribute for those links and properly style them via .qss? Or is not possible because they are inside a RichText?

@lslezak, @shundhammer :point_up_2: :point_up_2:

library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -0,0 +1,470 @@
...
+
+    # @macro seeAbstractWidget
+    def contents
+      HBox(content)
+    end
+
+    # @macro seeAbstractWidget
+    def handle(event)
+      if event["ID"].to_s.match?(Item.event_id)
```
jreidinger: why match? and not include?

library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -0,0 +1,470 @@
...
+        id, fired_by = event["ID"].split(Item.event_id)
+
+        item = find_item(id)
+
+        return unless item
```
jreidinger: Is it common? Looks a bit strange to not find it. I expect at least log entry.

library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -0,0 +1,470 @@
...
+    # @param [Item] the item that fired the event
+    def label_event_handler(item)
+      log.debug("Unhandled label event fired by #{item.inspect}")
+    end
+
+    # Returns the item with given id
+    #
+    # @param id [String] the id of the searched item
+    # @return [Item, nil] the item matching with given id
+    def find_item(id)
```
jreidinger: as id can be any object with `#to_s` why not allow to pass here also object with to_s ?

library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -0,0 +1,470 @@
...
+        @status = AUTO_SELECTED
+      end
+
+      # Returns richtext representation for the item
+      #
+      # Basically, an string containing two <a> or <span> tags, depending on the #enabled? method.
+      # One for the check box input and another for the label.
+      #
+      # @return [String] the item richtext representation
+      def to_s
```
jreidinger: I am not sure if to_s is ideal here as it can be confused. What about to_richtext or to_html ?

library/cwm/test/multi_status_selector_test.rb
```diff
@@ -0,0 +1,339 @@
...
+
+        it "displays the proper ion" do
+          expect(subject.to_s).to include("checkbox-off.svg")
+        end
+      end
+
+      context "and the item has an unknown status" do
+        let(:status) { :unknown }
+
+        it "displays the proper ion" do
```
jreidinger: it is a bit confusing here. Why not mention that it shows unselected icon? and there is also typo "ion" ( and also above )

library/cwm/src/lib/cwm/multi_status_selector.rb
dgdavid: Not, it shouldn't be common. In fact, previously I had there a `# FIXME: this is not expected, but...` comment.

Do you think we should avoid it?

library/cwm/src/lib/cwm/multi_status_selector.rb
dgdavid: Great idea. I'll use `to_richtext` because as far as I know it is a subset of HTML.

library/cwm/test/multi_status_selector_test.rb
dgdavid: > also typo "ion"
:smile: :roll_eyes: 

I'll fix it using your suggestion. Thanks.

library/cwm/src/lib/cwm/multi_status_selector.rb
dgdavid: Just a leftover, I guess. I'll use `#include?` instead. Thanks

library/cwm/src/lib/cwm/multi_status_selector.rb
dgdavid: I changed it, but I'm not sure if

* is what you mean by allowing to pass here an object responding to `#to_s`
* the resulting code is good/simple enough

What do you think now?

library/cwm/src/lib/cwm/multi_status_selector.rb
jreidinger: probably yes or at least add log entry or exception like ArgumentError

library/cwm/src/lib/cwm/multi_status_selector.rb
jreidinger: it looks fine for me.

library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -34,6 +34,10 @@ module CWM
   #
   # If you want to see it in action, have a look at yast2-registration or yast2-packager modules.
   #
+  # TODO: make possible to use it more than once in the same dialog, maybe by using the parent
+  # widget_id as a prefix for the event_id. See {CWM::MultiStatusSelector#handle} and
+  # {CWM::MultiStatusSelector::Item.event_id}.
```
jreidinger: yep, widget id as prefix for item sounds reasonable for me.

dgdavid: Perfect!

Let's keep the `TODO` there for the next iteration :roll_eyes:

PR #1031, opened 2020-03-13
[CWM] Minor improvements for the MultiStatusSelector widget

dgdavid:
### Problem

There is not enough space between items when running in Qt

### Solution

To surround items in `<p>` when not running in text mode.

## Screenshots

<details>
<summary>Click to show/hide</summary>

---

#### Before

Screenshot_sl5sp2_2020-03-13_01:13:27

#### After

Screenshot_sl5sp2_2020-03-13_02:05:47

</details>

---

See also https://github.com/yast/yast-yast2/pull/1029

### DIFF
--- a/library/cwm/src/lib/cwm/multi_status_selector.rb
+++ b/library/cwm/src/lib/cwm/multi_status_selector.rb
@@ -175,7 +175,17 @@ module CWM
 
     # Updates the content based on items list
     def refresh
-      content.value = items.map(&:to_richtext).join("<br>")
+      new_value = items.map do |item|
+        item_content = item.to_richtext
+
+        if Yast::UI.TextMode
+          "#{item_content}<br>"
+        else
+          "<p>#{item_content}</p>"
+        end
+      end
+
+      content.value = new_value.join
     end
 
     # Convenience widget to keep the content updated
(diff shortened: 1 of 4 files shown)


library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -175,7 +175,17 @@ def find_item(needle)
 
     # Updates the content based on items list
     def refresh
-      content.value = items.map(&:to_richtext).join("<br>")
+      new_value = items.map do |item|
+        item_content = item.to_richtext
+
+        if Yast::UI.TextMode
+          "#{item_content}<br>"
```
jreidinger: this will create <br> at the end. Do you try it in ncurses if it does not force empty line at the end? ( compared to previous code )

dgdavid: Sure, I'll do

jreidinger: and test case for it?

dgdavid: It seems to produce an acceptable output, without a new line after the latest item.

See the screenshot below.

Screenshot_20200313_044228

dgdavid: > and test case for it?

Done

PR #1034, opened 2020-03-23
Add Popup.SuppressFeedback method

jreidinger:
needed for trello: https://trello.com/c/AyGNkbK3/1708-3-sle15-sp2-p1-1165705-not-able-to-setup-network-when-connection-with-registration-server-failed-a-blocking-popupfeedback

### DIFF
--- a/library/general/src/modules/Popup.rb
+++ b/library/general/src/modules/Popup.rb
@@ -1035,7 +1035,7 @@ module Yast
 
     # Hides feedback during block execution.
     # After block it is shown again. When exception is raised, then it is
-    # shown again.
+    # NOT shown again.
     # @see {Feedback}, {ShowFeedback} and {ClearFeedback} for details
     # @param block block to execute
     def SuppressFeedback(&block)
--- a/library/general/src/modules/Popup.rb
+++ b/library/general/src/modules/Popup.rb
@@ -1009,6 +1009,8 @@ module Yast
       )
 
       @feedback_open = true
+      @feedback_headline = headline
+      @feedback_message = message
 
       nil
     end
@@ -1017,6 +1019,8 @@ module Yast
     # The popup is automatically closed at the end
     # (even when an exception is raised)
     # @see {ShowFeedback} and {ClearFeedback} for details
+    # To suppress temporary feedback see {SuppressFeedback}
+    #
     # @param headline [String] popup headline (displayed in bold)
     # @param message [String] message with details, displayed below the headline
     # @param block block to execute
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Popup.rb
```diff
@@ -1029,6 +1033,21 @@ def Feedback(headline, message, &block)
       ClearFeedback()
     end
 
+    # Hides feedback during block execution.
+    # After block it is shown again. When exception is raised, then it is
+    # shown again.
```
lslezak: I do not see any `ensure`, how is that open again?

library/general/src/modules/Popup.rb
```diff
@@ -1029,6 +1033,21 @@ def Feedback(headline, message, &block)
...
+    # @see {Feedback}, {ShowFeedback} and {ClearFeedback} for details
+    # @param block block to execute
+    def SuppressFeedback(&block)
+      raise ArgumentError, "block must be supplied" unless block
+
+      return block.call unless @feedback_open # just call block if feedback is not open
+
+      ClearFeedback()
+      block.call
+      ShowFeedback(@feedback_headline, @feedback_message)
```
lslezak: There is a different return value, if a feedback is displayed it returns `block.call` value otherwise it returns `ShowFeedback()`  return value which is `nil`.

jreidinger: well, no return value is documented, so it is void and no value is guaranteed.

library/general/src/modules/Popup.rb
jreidinger: Am typo in documentation. I want to have there "it is NOT shown again". Reason is that during exception it probably want to show something different then feedback that issue happen.

lslezak: Ah, then OK. :smiley:

library/general/test/popup_test.rb
```diff
@@ -43,6 +43,35 @@
...
+  describe ".SuppressFeedback" do
+    before do
+      allow(ui).to receive(:BusyCursor)
+      allow(ui).to receive(:GetDisplayInfo).and_return({})
+    end
+
+    it "closes a popup dialog and opens it at the end when feedback is shown" do
+      # twice as first is done by Feedback and second by SuppressFeedback
+      expect(ui).to receive(:OpenDialog).twice
+      expect(ui).to receive(:CloseDialog).twice
```
lslezak: Maybe we should test the correct OpenDialog/CloseDialog/OpenDialog/CloseDialog sequence so we can ensure the correct final state.

PR #1035, opened 2020-03-24
Fixed RichText multi-selection alignment (bsc#1167523)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1167523
- I tried several approaches
  - Adding `vertical-align: middle` styling did not work well, the text was moved up, but too much :worried: 
  - Adding `padding-bottom: 3px` worked fine in 800x600 installation, but was broken in 1280x1024 (it uses bigger font so the 3px padding was not enough). Unfortunately Qt accepts only pixel units here, we would need some unit relative to the font size like `em` which is not accepted :worried: 
  - Finally using a table works in all cases! :+1: 
- Some details can be found the Qt documentation: https://doc.qt.io/qt-5/richtext-html-subset.html
- Additionally I noticed some other broken `style` attributes with missing `=` character :scream: 

### The Original Look

Full screen
richtext_alignment_original

1:1 crop
richtext_alignment_detail

### With the Fix

*Notice also a smaller space between the items, that means less scrolling...*

Full screen (800x600)

richtext_alignment_fixed_new

1:1 crop (800x600)

richtext_alignment_fixed_detail_new

Full screen (1280x1024)

richtext_alignment_fixed_hires_new

1:1 crop (1280x1024)

richtext_alignment_fixed_hires_detail_new

Textmode (looks the same, no regression :smile:)
richtext_alignment_fixed_textmode

### DIFF
--- a/library/cwm/src/lib/cwm/multi_status_selector.rb
+++ b/library/cwm/src/lib/cwm/multi_status_selector.rb
@@ -437,7 +437,7 @@ module CWM
         if enabled?
           "<a href=\"#{id}#{INPUT_EVENT_ID}\" style=\"#{text_style}\">#{icon}</a>"
         else
-          "<span style=\"#{text_style}\">#{icon}</a>"
+          "<span style=\"#{text_style}\">#{icon}</span>"
         end
       end
 
--- a/library/cwm/src/lib/cwm/multi_status_selector.rb
+++ b/library/cwm/src/lib/cwm/multi_status_selector.rb
@@ -401,9 +401,8 @@ module CWM
       #
       # @return [String] the item richtext representation
       def to_richtext
-        # add extra space in graphical mode
-        space = Yast::UI.TextMode ? "" : " "
-        "<tr><td>#{checkbox_input}#{space}</td><td> #{checkbox_label}</td></tr>"
+        # add some small space (&nbsp;) between the icon and the label
+        "<tr><td>#{checkbox_input}&nbsp;</td><td>#{checkbox_label}</td></tr>"
       end
 
       # Determines whether running in installation mode
@@ -448,7 +447,7 @@ module CWM
         if enabled?
           "<a href=\"#{id}#{LABEL_EVENT_ID}\" style=\"#{text_style}\">#{label}</a>"
         else
-          "<span style=\"#{text_style}\">#{label}</a>"
+          "<span style=\"#{text_style}\">#{label}</span>"
         end
       end
 
--- a/library/cwm/src/lib/cwm/multi_status_selector.rb
+++ b/library/cwm/src/lib/cwm/multi_status_selector.rb
@@ -125,15 +125,13 @@ module CWM
     def refresh
       new_value = items.map do |item|
         item_content = item.to_richtext
-
-        if Yast::UI.TextMode
-          "#{item_content}<br>"
-        else
-          "<p>#{item_content}</p>"
-        end
+        # HACK: tables rows are displayed side-by-side in text mode (ugh?),
+        # add an extra line break in that case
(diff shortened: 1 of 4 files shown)


library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -436,7 +437,7 @@ def checkbox_input
         if enabled?
           "<a href=\"#{id}#{INPUT_EVENT_ID}\" style=\"#{text_style}\">#{icon}</a>"
         else
-          "<span style\"#{text_style}\">#{icon}</a>"
+          "<span style=\"#{text_style}\">#{icon}</a>"
```
dgdavid: Nice catch!

But _we_ (mostly me) also forget to use the right tag to close it. The typical copy&paste problem, sorry.

```suggestion
          "<span style=\"#{text_style}\">#{icon}</span>"
```

library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -447,7 +448,7 @@ def checkbox_label
         if enabled?
           "<a href=\"#{id}#{LABEL_EVENT_ID}\" style=\"#{text_style}\">#{label}</a>"
         else
-          "<span style\"#{text_style}\">#{label}</a>"
+          "<span style=\"#{text_style}\">#{label}</a>"
```
dgdavid: Let's not only add the missing `=` but also close the `span` properly :)
```suggestion
          "<span style=\"#{text_style}\">#{label}</span>"
```

library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -125,15 +125,13 @@ def contents
     def refresh
       new_value = items.map do |item|
         item_content = item.to_richtext
-
-        if Yast::UI.TextMode
-          "#{item_content}<br>"
-        else
-          "<p>#{item_content}</p>"
-        end
+        # HACK: tables rows are displayed side-by-side in text mode (ugh?),
```
dgdavid: Rows side-by-side? Really? :see_no_evil:

library/cwm/src/lib/cwm/multi_status_selector.rb
```diff
@@ -397,12 +395,15 @@ def auto_select!
...
+      # Basically a string containing a table row with two <a> or <span> tags,
+      # depending on the #enabled? method.
       # One for the check box input and another for the label.
       #
       # @return [String] the item richtext representation
       def to_richtext
-        "#{checkbox_input} #{checkbox_label}"
+        # add extra space in graphical mode
+        space = Yast::UI.TextMode ? "" : " "
+        "<tr><td>#{checkbox_input}#{space}</td><td> #{checkbox_label}</td></tr>"
```
dgdavid: Another way to do it could be:

```suggestion
        gap = Yast::UI.TextMode ? " " : "  "
        "<tr><td>#{checkbox_input}#{gap}</td><td>#{checkbox_label}</td></tr>"
```

If there is a problem having two spaces in a row, we can use

```
gap = Yast::UI.TextMode ? "&nbsp;" : "&nbsp;&nbsp;"

```

library/cwm/src/lib/cwm/multi_status_selector.rb
lslezak: Yes, I was also surprised a lot...

library/cwm/src/lib/cwm/multi_status_selector.rb
lslezak: It turned out that Qt displays `&nbsp;` a bit wider than a normal space so it is enough to have just one `&nbsp;` in both Qt and ncurses.

PR #1036, opened 2020-03-25
Force a cleand and read of network interfaces configuration aftet writing (bsc#1166778)

teclator:
## Problem

During installation, the network config is cached when it is read. In case that the network interfaces configuration is written, then, the `SCR` configuration is flush, but it is not read again as it is cache by `Yast::Lan`

#### Initial config

DHCLIENT_SE_HOSTNAME_Any
Cat-DHCLIENT_SET_HOSTNAME-Any

#### DHCP Hostname to eth0

Eth0
Cat-DHCP_CLIENT_SET_HOSTNAME-Eth0

#### DHCP Hostname to eth1
Eth1
Cat-DHCLIENT_SET_HOSTNAME-Eth1

#### Back and forth between eth0 and eth1

Cat-DHCLIENT_SET_HOSTNAME-Eth1Eth0Eth1

### Expected result 

The Interfaces that are not selected for setting the Hostname by DHCP should remove the DHCLIENT_SET_HOSTNAME attribute.

- https://bugzilla.suse.com/show_bug.cgi?id=1166778

## Solution

As network cache the config when read (https://github.com/yast/yast-network/blob/master/src/include/network/lan/complex.rb#L174). The easiest solution seems to be a clean and read of network interfaces config after writing it.

## Note

It seems that there could be also a bug in the wicked hostname extension which no change the current hostname when setting it by a different interface (example changing between eth0 and eth1 with different hostnames).

When the /var/run/wicked/extensions/hostname/* content is removed it is updated correctly.

https://github.com/openSUSE/wicked/blob/master/extensions/hostname#L70


library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -954,6 +954,7 @@ def Write(devregex)
 
       # Finish him
       SCR.Write(path(".network"), nil)
+      CleanCacheRead()
```
jreidinger: please add comment why it is needed.

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Mar 25 12:12:00 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Force a cache clean and read of the current network interfaces
```
jreidinger: I would write it as "Reread interfaces configuration after writing it" which seems simple and still not user oriented. So maybe adding something like "to avoid wrong values when reopen dialog in installation" ?

library/network/src/modules/NetworkInterfaces.rb
teclator: Sure, I was still working on it, that is why it is still in draft :)

library/network/test/network_interfaces_test.rb
```diff
@@ -306,4 +316,40 @@ def mock_ppp
...
+      expect(::FileUtils.compare_file(ifcfg_copy, ifcfg_file)).to eq(true)
+    end
+
+    it "deletes removed interfaces" do
+      size = Yast::NetworkInterfaces.List("").size
+      Yast::NetworkInterfaces.Delete("copy")
+      Yast::NetworkInterfaces.Commit()
+      Yast::NetworkInterfaces.Write("")
+      expect(Yast::NetworkInterfaces.List("").size).to eq(size - 1)
+    end
```
jreidinger: test for that reread?

teclator: Ok, will add an explicit test, I checked in previous test that it was write and modifications were read but it is not mention as documentation

PR #1038, opened 2020-03-27
merge 'SLE-15-SP1' into 'master'

teclator:
It merges #1036

### DIFF
--- a/package/yast2.spec
+++ b/package/yast2.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2
-Version:        4.2.78
+Version:        4.2.79
 Release:        0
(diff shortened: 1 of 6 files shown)


package/yast2.spec
```diff
@@ -17,7 +17,8 @@
 
 
 Name:           yast2
-Version:        4.2.78
+Version:        4.2.79
+
```
jreidinger: is it intentional new line? usually it is not here as it is related to Release later

teclator: I probably left it by error when adapted the fixing conflict in the spec file. Removing it

PR #1039, opened 2020-03-31
Xml parser

jreidinger:
trello: https://trello.com/c/TDfgW1oy/1549-8-improve-used-xml-parser

Tasks:

- [x] test for generating XML
- [x] test for parsing XML
- [x] new parser that does not depend on .xml agent


lslezak:
What about the Nokogiri dependency? Should  whole `yast2.rpm` depend on the `rubygem-nokogiri` or should the clients which use this XML module depend on it?

It's quite big:

```console
# zypper in ruby2.5-rubygem-nokogiri

The following 2 NEW packages are going to be installed:
  ruby2.5-rubygem-mini_portile2 ruby2.5-rubygem-nokogiri

2 new packages to install.
Overall download size: 9.1 MiB. Already cached: 0 B. After the operation,
additional 9.6 MiB will be used.
```

Adding it to `yast2.rpm` would increase the needed space in installed system quite a lot, and we do not need it in installed system in most cases...

jreidinger:
@lslezak I need to check why it is so big, as it is just C wrapper around libxml2 and few ruby classes on top. Reason for nokogiri was in fact minimal dependencies it has

jreidinger:
@lslezak I check it and surprisingly the biggest part is gem cache itself, as it contain probably C sources of non trivial size, that is not so big compiled:

```
0	/etc/alternatives/nokogiri
0	/etc/alternatives/nokogiri-1.10.9
0	/etc/alternatives/nokogiri.ruby2.6
0	/usr/bin/nokogiri
0	/usr/bin/nokogiri-1.10.9
0	/usr/bin/nokogiri.ruby2.6
1	/usr/bin/nokogiri.ruby2.6-1.10.9
9	/usr/lib64/ruby/gems/2.6.0/cache/nokogiri-1.10.9.gem
1	/usr/lib64/ruby/gems/2.6.0/extensions/x86_64-linux/2.6.0/nokogiri-1.10.9/nokogiri
1	/usr/lib64/ruby/gems/2.6.0/extensions/x86_64-linux/2.6.0/nokogiri-1.10.9
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/bin
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/ext/nokogiri
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/ext
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/css
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/decorators
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/html/sax
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/html
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/xml/node
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/xml/pp
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/xml/sax
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/xml/xpath
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/xml
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri/xslt
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/nokogiri
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/xsd/xmlparser
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib/xsd
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/lib
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/patches/libxml2
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9/patches
1	/usr/lib64/ruby/gems/2.6.0/gems/nokogiri-1.10.9
1	/usr/lib64/ruby/gems/2.6.0/specifications/nokogiri-1.10.9.gemspec
0	/usr/share/doc/packages/ruby2.6-rubygem-nokogiri
10	total
```

lslezak:
The *.gem file is so big because it contains full `libxml2` and `libxslt` source tarballs... (~9MB)

library/xml/src/modules/XML.rb
```diff
@@ -1,6 +1,6 @@
 # ***************************************************************************
 #
-# Copyright (c) 2002 - 2012 Novell, Inc.
+# Copyright (c) 2002 - 2020 Novell, Inc.
```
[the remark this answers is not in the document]
jreidinger: not sure as novell was original holder...I think it does not matter anymore, but who knows.

library/xml/src/modules/XML.rb
```diff
@@ -97,96 +132,226 @@ def xmlCreateDoc(doc, doc_settings)
...
     end
 
     # Write YCP data into formated XML string
     # @param [Symbol] doc_type Document type identifier
     # @param [Hash] contents  a map with YCP data
-    # @return [String] String with XML data
+    # @return [String, nil] String with XML data or nil if error happen
+    # @raise [XMLInvalidObject] when non supported contents is passed
     def YCPToXMLString(doc_type, contents)
       contents = deep_copy(contents)
```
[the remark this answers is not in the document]
jreidinger: well, we do not modify data, so not needed.

library/xml/src/modules/XML.rb
```diff
@@ -97,96 +132,226 @@ def xmlCreateDoc(doc, doc_settings)
...
+      log.info "Reading #{xml_file}"
+      XMLToYCPString(SCR.Read(path(".target.string"), xml_file))
     end
 
-    # Read XML string into YCP
-    # @param xml_string string to read
-    # @return Map with YCP data
+    # Reads XML string
+    # @param [String] xml_string to read
+    # @return [Hash] parsed content or nil if error happen and in such case
```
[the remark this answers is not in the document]
jreidinger: actually I forgot that nil, it is no longer returned, as it is replaced by exceptions.

library/xml/src/modules/XML.rb
```diff
@@ -97,96 +132,226 @@ def xmlCreateDoc(doc, doc_settings)
...
+    rescue Nokogiri::XML::SyntaxError => e
+      raise XMLParseError, e.message
+    end
+
+    # Validates given schema
+    #
+    # @param xml [String] path or content of XML
+    # @param schema [String] path or content of relax ng schema
+    # @return [String] string with errors or empty string
+    def XMLValidation(xml, schema)
```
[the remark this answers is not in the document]
jreidinger: I name it this way only for consistency with other methods. Fine for me to use validate.

library/xml/src/modules/XML.rb
```diff
@@ -97,96 +132,226 @@ def xmlCreateDoc(doc, doc_settings)
...
-        Builtins.y2warning("can't convert empty XML string")
-        {}
+        schema_content = SCR.Read(path(".target.string"), schema)
+        schema_path = File.dirname(schema)
+        # change directory so relative include works
+        Dir.chdir(schema_path) { validator = Nokogiri::XML::RelaxNG(schema_content) }
       end
+
+      doc = Nokogiri::XML(xml)
+      validator.validate(doc).map(&:message).join("\n")
```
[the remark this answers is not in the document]
jreidinger: Fine for me to return array.

library/xml/src/modules/XML.rb
```diff
@@ -97,96 +132,226 @@ def xmlCreateDoc(doc, doc_settings)
...
+        value = text.to_i
+      when "boolean"
+        value = BOOLEAN_MAP[text]
+        raise XMLInvalidContent, "xml node '#{node.name}' is invalid. Only true and false is allowed for boolean." if value.nil?
+      when "list"
+        value = []
+        children.each do |kid|
+          # always pass new hash to prevent overwrite for list with same elements
+          r = {}
+          parse_node(kid, r)
```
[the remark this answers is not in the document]
jreidinger: this won't work as I need later to use that r to get values

library/xml/src/modules/XML.rb
```diff
@@ -97,96 +132,226 @@ def xmlCreateDoc(doc, doc_settings)
...
+        raise XMLInvalidContent, "xml node '#{node.name}' is empty. Forbidden for integer." if text.empty?
+        raise XMLInvalidContent, "xml node '#{node.name}' is invalid integer." if text !~ /-?\d+/
+
+        value = text.to_i
+      when "boolean"
+        value = BOOLEAN_MAP[text]
+        raise XMLInvalidContent, "xml node '#{node.name}' is invalid. Only true and false is allowed for boolean." if value.nil?
+      when "list"
+        value = []
+        children.each do |kid|
```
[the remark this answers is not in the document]
jreidinger: maybe even map can be used as we have more knowledge about it.

library/xml/src/modules/XML.rb
```diff
@@ -97,96 +132,226 @@ def xmlCreateDoc(doc, doc_settings)
...
+          special_names = metadata["listEntries"] || {}
+          element_name = special_names[key] || "listentry"
+          value.each do |list_value|
+            add_element(doc, metadata, element, element_name => list_value)
+          end
+        when ::Hash
+          element[type_attr] = "map"
+          add_element(doc, metadata, element, value)
+        when nil
+          raise XMLNilObject, contents
```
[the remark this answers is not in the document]
jreidinger: well, this was decided on xml parser meeting, that we want exception in that cases.
As I said my main premise is if YCPToXML does not raise exception, then its result has to be same as XMLToYCP and it is not true for nil, if we skip it. So I propose there to have nil type, but conclusion was to use exception as we should not serialize nil.

package/yast2.spec
```diff
@@ -185,6 +185,13 @@ if [ -f "/etc/sysctl.d/30-yast.conf" ]; then
     fi
 fi
 
+# fix xlm problems in older control.xml to prevent xml parser failures
```
mvidner, line 188: Given that [we have agreed](https://github.com/yast/yast-installation-control/pull/96#issuecomment-623306963) to make `config:type="string"` the default, this change would be unnecessary if we also allowed `config:type="disksize"` to mean `string` for backward compatibility.

Then we would not need to edit /etc files in RPM post scripts, which is INTRUSIVE and error prone. I consider it a bigger problem than continuing to allow `disksize` in the YaST-XML layer.

jreidinger:
@mvidner should I add also exporting/parsing type shortcut `t` here?

library/xml/test/xml_test.rb
```diff
@@ -187,6 +187,50 @@
...
+
+      it "recognizes the 'config:type' attribute even with a different prefix" do
+        input = "<test xmlns=\"http://www.suse.com/1.0/yast2ns\" xmlns:c=\"http://www.suse.com/1.0/configns\">\n" \
+          "  <foo c:type=\"symbol\">sym</foo>\n" \
+          "</test>\n"
+        expected = { "foo" => :sym }
+        expect(subject.XMLToYCPString(input)).to eq expected
+      end
+
+      it "in case of conflict, the shortest attribute wins (t over type over config:type)" do
```
jreidinger: is it valid XML according to schema? I think it should forbid multiple attributes.

library/xml/test/xml_test.rb
```diff
@@ -354,32 +354,32 @@
...
-      "<test xmlns=\"http://www.suse.com/1.0/yast2ns\" xmlns:config=\"http://www.suse.com/1.0/configns\">\n" \
-      "  <test>\n" \
-      "    <lest config:type=\"boolean\">false</lest>\n" \
-      "    <int config:type=\"integer\">5</int>\n" \
-      "    test value \n" \
-      "  </test>\n" \
-      "</test>\n"
-
-    expect { subject.XMLToYCPString(input) }.to raise_error(Yast::XMLInvalidContent)
+      expect { subject.XMLToYCPString(input) }.to raise_error(Yast::XMLInvalidContent)
```
jreidinger, line 380: :+1: too much nesting when I am placing it

library/xml/src/modules/XML.rb
```diff
@@ -253,7 +253,7 @@ def parse_node(node, result)
       # we need just direct text under node. Can be splitted with another elements
       # but remove whitespace only text
       name = node.name
-      type = node["type"] || detect_type(text, children, node)
+      type = node["t"] || node["type"] || detect_type(text, children, node)
```
jreidinger, line 256: btw this means we can also use "config:t=", but I would not promote it.

library/xml/src/modules/XML.rb
```diff
@@ -300,8 +300,7 @@ def add_element(doc, metadata, parent, contents)
         raise XMLInvalidKey, key unless key.is_a?(::String)
 
         value = contents[key]
-        type_attr = metadata["typeNamespace"] ? "config:" : ""
```
jreidinger: maybe we should document that typeNamespace is now ignored and is deprecated as it does not affect anything?

library/xml/test/xml_test.rb
mvidner: OK, will fix to `raise`

library/xml/src/modules/XML.rb
mvidner: Will do

package/yast2.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Thu May 14 07:32:16 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- Re-implement XML parser (bsc#1170886):
+-- use as backend nokogiri instead of hand written libxml2 wrapper
```
mvidner, line 5: So are we able to drop the old one?

jreidinger: not right now, as there are one module in autoYaST that calls directly xml agent, but plan is to replace it with XML module.

mvidner: OK, it is https://github.com/yast/yast-autoinstallation/blob/f61e00734ee8c6caa7810a0ffd48afb422a31281/src/modules/AutoinstClass.rb#L55

Then there is https://github.com/yast/yast-update/blob/77b9d0f206002e3e34eccab8beace13f1c5b59e2/src/modules/RootPart.rb#L1214 using `.anyxml`, https://github.com/yast/yast-yast2/blob/master/library/general/src/servers_non_y2/ag_anyxml

jreidinger: good catches. I think we should simply drop that ag_anyxml and use nokogiri directly as it is used for pam, so not "our" xml

library/xml/src/modules/XML.rb
```diff
@@ -293,12 +282,13 @@ def parse_node(node, result)
     # @param [Nokogiri::XML::Document] doc
     # @param [Hash] metadata for current doc
     # @param [Nokogiri::XML::Node] parent
-    # @param [Hash] content to write
+    # @param [Hash] contents content to write
+    # @param [Array] parent_array (for better error reporting)
     # @return [void]
-    def add_element(doc, metadata, parent, contents)
+    def add_element(doc, metadata, parent, contents, parent_array = nil)
```
jreidinger, line 286: np, I would use named parameter instead of optional. This will help if we need more parameters in future.

library/xml/src/modules/XML.rb
```diff
@@ -97,96 +119,236 @@ def xmlCreateDoc(doc, doc_settings)
...
+            add_element(doc, metadata, element, { element_name => list_value }, value)
+          end
+        when ::Hash
+          element[type_attr] = "map"
+          add_element(doc, metadata, element, value)
+        else # including nil
+          # without parent_array, a ["foo", nil, "bar"] would always
+          # unhelpfully report {"listentry"=> nil} as the parent object
+          o = parent_array || contents
+          raise XMLSerializationError.new("Cannot represent #{value.inspect}, part of #{o.inspect}", o)
```
jreidinger, line 320: I worry that even that part can be useless if it is something like "[nil]". What we probably should do is always log whole structure we serialize and raise exception. As I think it is always our fault to provide wrong ycp value to serializer, so it always will need logs and our fix.

mvidner: I think it is reasonably balanced now, when we haven't checked all the usage sites. If we do that, we can improve the library further.

PR #1041, opened 2020-04-03
Change the way YaST determines if systemd is running or not

teclator:
## Problem

YaST network does not show the global tab in the running system because it checks whether systemd is running or not and it is not properly detected since a change in the sysfs cgroup systemd hierarchy

- https://trello.com/c/9eYR4f6p/4042-tw-p2-1168307-lan-no-global-tab-sys-fs-cgroup-systemd-no-such-file-or-directory

## Solution

Change the way YaST determines if systemd is running.

## Test

Tested manually in a running system running the **yast2 network** module and in WSL.

### DIFF
--- a/library/systemd/src/modules/Systemd.rb
+++ b/library/systemd/src/modules/Systemd.rb
@@ -38,7 +38,6 @@ module Yast
       @systemd_path = "/usr/lib/systemd/systemd"
       @default_target_symlink = "/etc/systemd/system/default.target"
       @systemd_targets_dir = "/usr/lib/systemd/system"
-      @systemd_mountdir = "/sys/fs/cgroup/systemd"
 
       textdomain "base"
     end
@@ -52,7 +51,7 @@ module Yast
     # Check whether systemd init is currently running
     # @return boolean true if systemd init is running
     def Running
-      FileUtils.IsDirectory(@systemd_mountdir) == true
+      SCR.Read(path(".target.string"), "/proc/1/comm")&.chomp == "systemd"
     end
 
     # Set default runlevel for systemd (assumes systemd is installed)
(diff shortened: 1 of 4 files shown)


library/systemd/src/modules/Systemd.rb
```diff
@@ -52,7 +52,7 @@ def Installed
     # Check whether systemd init is currently running
     # @return boolean true if systemd init is running
     def Running
-      FileUtils.IsDirectory(@systemd_mountdir) == true
+      SCR.Read(path(".target.string"), "/proc/1/comm")&.chomp == "systemd"
```
teclator, line 55: Another alternative:
```suggestion
      File.directory?(File.join(@running_dir, "/system"))
```

library/systemd/src/modules/Systemd.rb
```diff
@@ -38,7 +38,7 @@ def main
       @systemd_path = "/usr/lib/systemd/systemd"
       @default_target_symlink = "/etc/systemd/system/default.target"
       @systemd_targets_dir = "/usr/lib/systemd/system"
-      @systemd_mountdir = "/sys/fs/cgroup/systemd"
+      @running_dir = "/run/systemd"
```
jreidinger, line 41: is it needed without alternative?

library/systemd/src/modules/Systemd.rb
jreidinger: I am not sure if we should call it in target as I think systemd cannot run in chroot and mount bind proc, so it is in the end also local proc

library/systemd/src/modules/Systemd.rb
teclator: No, it is not, we can remove it if we are fine with the current solution

library/systemd/src/modules/Systemd.rb
teclator: That is why shouldn't matter, but yep, it is better to use local

PR #1044, opened 2020-04-13
CWM improvements

imobachgs:
Two small improvements after the first review of https://github.com/yast/yast-autoinstallation/pull/592.

* When setting the value of a `ComboBox` item, the new value will be added to the list of items if it was not already included and the combo is editable.
* A MenuButton listens to events from its children.

### DIFF
--- a/library/cwm/src/lib/cwm/common_widgets.rb
+++ b/library/cwm/src/lib/cwm/common_widgets.rb
@@ -184,6 +184,14 @@ module CWM
     include ValueBasedWidget
     include ItemsSelection
     abstract_method :label
+
+    alias_method :orig_value=, :value=
+    def value=(val)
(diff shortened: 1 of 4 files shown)


library/cwm/src/lib/cwm/common_widgets.rb
```diff
@@ -184,6 +184,18 @@ class ComboBox < AbstractWidget
     include ValueBasedWidget
     include ItemsSelection
     abstract_method :label
+
+    alias_method :orig_value=, :value=
+    def value=(val)
+      change_items([[val, val]] + items) if editable? && !items.map(&:first).include?(val)
+      self.orig_value = val
+    end
```
jreidinger, line 192: change looks good, just please document it.

jreidinger (approved):

PR #1045, opened 2020-04-13
Remove deleted secondary IP addresses

teclator:
## Problem

Removed secondary IP addresses through the yast2-network dialog are marked to be deleted but are not removed from the ifcfg files since (https://github.com/yast/yast-yast2/pull/789)

- https://trello.com/c/xHnv0zCT/4047-osdistribution-p5-1146020-setting-additional-ip-addresses-in-eth-adapter-configurations-delete-of-additional-ip-address-not-poss
- https://bugzilla.suse.com/show_bug.cgi?id=1146020

## Solution

Delete not only removed interfaces but also removed secondary IP addresses.

### DIFF
--- a/library/network/src/modules/NetworkInterfaces.rb
+++ b/library/network/src/modules/NetworkInterfaces.rb
@@ -1609,7 +1609,7 @@ module Yast
     #   removed
     # @param alias_num [String] index num of the alias that needs to be removed
     def delete_alias(devices, iface, alias_num)
-      dev_map = devices.values.find { |d| d.keys.include? (iface) } || {}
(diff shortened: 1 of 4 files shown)


library/network/src/modules/NetworkInterfaces.rb
```diff
@@ -1591,6 +1590,39 @@ def get_devices(devregex)
       devices
     end
 
+    # Convenience method to delete an interface config file from the system
+    #
+    # @param iface [String] interface name of the config file to be deleted
+    def delete_device(iface)
+      # delete config file
```
imobachgs, line 1597: I guess this comment is not needed anymore.

PR #1047, opened 2020-04-15
Bnc#1081441

srinidhibs:
Display appropriate Bugzilla URL depending on the release name of the product on which the error occurred.

In case of an error, the error dialog will request the user to file a report on https://bugzilla.opensuse.org when the release name in `/etc/os-release` contains "openSUSE". For any other release name, the
Bugzilla URL displayed would be https://bugzilla.suse.com.

## Screenshots

**Before**
Error Screenshot on any SUSE release

**After (on openSUSE system)**
Error Screenshot on openSUSE

I tested this by commenting out **NAME=** attribute from `/etc/os-release` file. In this case, the `OSRelease.ReleaseName` method will return the default value of "SUSE Linux" and in that case, https://bugzilla.suse.com URL is displayed to the user.

Fixes [bsc#1081441](https://bugzilla.novell.com/show_bug.cgi?id=1081441)

### DIFF
--- a/library/control/src/modules/InstError.rb
+++ b/library/control/src/modules/InstError.rb
@@ -109,7 +109,7 @@ module Yast
     # @param [String] details (displayed as a plain text, can contain multiple lines)
     def ShowErrorPopUp(heading, error_text, details)
       bugzilla_url = "http://bugzilla.suse.com/"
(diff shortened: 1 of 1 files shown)


jreidinger (approved):
thanks for patch

PR #1048, opened 2020-04-24
Running chrooted fix

teclator:
## Problem

In case of a live installation, systemd is detected as running failing in case we try to show the properties of a systemd unit.

- https://bugzilla.suse.com/show_bug.cgi?id=1168849

## Solution

- Check whether we are running inside a chroot in order to return the systemd unit properties.

### DIFF
--- a/library/systemd/src/lib/yast2/systemd/unit.rb
+++ b/library/systemd/src/lib/yast2/systemd/unit.rb
@@ -106,7 +106,9 @@ module Yast2
       # @return [Yast2::Systemd::UnitProperties]
       def show(property_text = nil)
         # Using different handler during first stage (installation, update, ...)
-        if Yast::Stage.initial && !Yast::Systemd.Running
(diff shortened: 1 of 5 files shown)


library/systemd/src/lib/yast2/systemd/unit.rb
```diff
@@ -106,7 +106,7 @@ def refresh!
       # @return [Yast2::Systemd::UnitProperties]
       def show(property_text = nil)
         # Using different handler during first stage (installation, update, ...)
-        if Yast::Stage.initial && !Yast::Systemd.Running
+        if Yast::WFM.scr_chrooted? || (Yast::Stage.initial && !Yast::Systemd.Running)
```
jreidinger, line 109: add comment that systemd cannot run in chroot

teclator:
:heavy_check_mark: Created IBS submit [request #215320](https://build.suse.de/request/show/216590)

PR #1049, opened 2020-04-24
Merge SLE-15-SP2 into 'master'

teclator:
It merges #1048

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Apr 24 14:53:03 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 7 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Apr 24 14:53:03 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Avoid using systemctl calls when already started with the
+  installation and thus, running inside the chroot (bsc#1168849)
+- 4.2.83
```
jreidinger, line 6: so we still in sync so we do not need minor bump, right?

teclator: exactly

PR #1055, opened 2020-05-27
autoinst_issues/list.add : Changed first parameter from tag to classes

schubi2:
- https://trello.com/c/vAZPOYpg/1838-8-better-error-handling
- https://bugzilla.suse.com/show_bug.cgi?id=1171335 "AutoYaST: improve error reporting mechanism "

- autoinst_issues/list.add : Changed first parameter from tag to  classes
- Added general AY issue classes: InvalidValue, MissingValue

### DIFF
--- a/library/general/test/autoinst_issues/issues_presenter_test.rb
+++ b/library/general/test/autoinst_issues/issues_presenter_test.rb
@@ -24,7 +24,7 @@ require "installation/autoinst_issues/issue"
 require "installation/autoinst_issues/list"
 require "installation/autoinst_issues/issues_presenter"
 
-module Y2Autoinstallation
+module Test
   module AutoinstIssues
     class MissingSection < ::Installation::AutoinstIssues::Issue
       def initialize(*args)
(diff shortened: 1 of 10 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon May 18 10:21:09 UTC 2020 - schubi@suse.de
+- autoinst_issues/list.add : Changed first parameter from tag to
```
imobachgs, line 3: np: Please, keep the empty line here :)

library/general/test/autoinst_issues/issues_presenter_test.rb
```diff
@@ -24,7 +24,7 @@
 require "installation/autoinst_issues/list"
 require "installation/autoinst_issues/issues_presenter"
 
-module Installation
+module Y2Autoinstallation
```
imobachgs, line 27: I would not use a module name that already exists (to avoid potential problems). You do not even need to use a module at all. Just declaring a DummyMissingSection class should be enough.

PR #1059, opened 2020-06-05
Fix system service actions

joseivanlopez:
## Problem

When a systemd service is initially stopped, the `SystemService` object will always consider the service as stopped. But sometimes the service could be started later via its socket. In that case, if we set the service to be restarted, the systemd service would never be stopped because `SystemService` believes that the systemd service continues stopped.

Moreover, in case that the service is configured to be started on demand, `SystemService` tries to start the socket instead of the service. But it would fail to start the socket because the service is already running.

* https://bugzilla.suse.com/show_bug.cgi?id=1162514
* https://trello.com/c/gEHPeax1/1882-sles15-sp1-p5-1162514-iscsi-session-is-not-connected-after-successful-connection-via-yast2-iscsi-client

## Solution

There are two possible solutions:

a) To check the current systemd status on the system to determine whether to stop/start the service.

b) Always try to stop/start the service independently on its current state.

Option b) was implemented. When a systemd service is already stopped, another call to `systemctl stop service` does not affect at all. And the same happens when trying to start an already started service.

## Testing

* Manually tested
* Adapted unit tests

### DIFF
--- a/library/systemd/src/lib/yast2/system_service.rb
+++ b/library/systemd/src/lib/yast2/system_service.rb
@@ -1,6 +1,4 @@
-# encoding: utf-8
-
-# Copyright (c) [2018] SUSE LLC
+# Copyright (c) [2018-2020] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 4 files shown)


joseivanlopez:
MU: https://build.suse.de/request/show/219809

PR #1060, opened 2020-06-08
Fix system service actions (forward port to SP2)

joseivanlopez:
See https://github.com/yast/yast-yast2/pull/1059.


joseivanlopez:
MU: https://build.suse.de/request/show/219816

PR #1062, opened 2020-06-09
Fix Xen detection

dgdavid:
Using the information available at /proc/xen instead of relying in the
`uname -r` output.


library/general/test/arch_test.rb
```diff
@@ -7,16 +7,99 @@
 require "yast"
 
 describe Yast::Arch do
+  subject { described_class }
+
+  before do
+    # need to reset all initialization of the module for individual
+    # test cases which mock different hardware
+    # otherwise values in Arch.rb remain cached
+    module_path = File.expand_path("../src/modules/Arch.rb", __dir__)
```
dgdavid, line 16: Note: To avoid the [warning produced by this technique](https://travis-ci.org/github/yast/yast-yast2/builds/696486416#L2227), 

> /usr/src/app/library/general/src/modules/Arch.rb:520: warning: already initialized constant Yast::Arch

I changed this by moving the [local variables initialization](https://github.com/yast/yast-yast2/blob/852ccd7f514635fb0b9c4c261be58e0cac58bfc7/library/general/src/modules/Arch.rb#L34-L60) to their own method and adding a `#reset!` one to _re-initialize_ them (a.k.a, clean the cache).

However, since it was only for the test purpose I rolled it back.

PR #1063, opened 2020-06-17
Reduce exported profile

imobachgs:
This PR adds support to determine the default start_mode for a system service.

```ruby
require "yast2/system_service"
cups = Yast2::SystemService.find("cups")
cups.default_start_mode #=> :on_boot
cups.start_mode #=> :on_boot
cups.default_start_mode? #=> true

# Let's change cups start mode
cups.start_mode = :on_demand
cups.default_start_mode? #=> false
```

Trello: https://trello.com/c/1RXM7d8u/1851-8-reduce-profile-i-do-not-export-not-relevant-information
Bug: https://bugzilla.opensuse.org/show_bug.cgi?id=1172749


library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -205,14 +205,16 @@ def substate
...
+      @current_start_mode = start_mode_from(service, socket, :enabled?)
+    end
+
+    # Determines whether the start mode is set to the default values
+    #
+    # @return [Boolean]
+    def default_start_mode
+      return @default_start_mode unless @default_start_mode.nil?
+
+      @default_start = start_mode_from(service, socket, :preset_enabled?)
```
dgdavid, line 217: Do you mind `@default_start_mode`?

library/systemd/src/lib/yast2/system_service.rb
```diff
@@ -205,14 +205,16 @@ def substate
...
-          :on_boot
-        elsif socket&.enabled?
-          :on_demand
-        else
-          :manual
-        end
+      @current_start_mode = start_mode_from(service, socket, :enabled?)
+    end
+
+    # Determines whether the start mode is set to the default values
```
dgdavid, line 211: Actually the method sets the default mode, doesn't it?

imobachgs: No, it does not set but reads the value. But the description is wrong. Thanks for noticing it.

PR #1064, opened 2020-06-18
Updated Yast::XML.validate arguments

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1170886
- Distinguish between a `String` argument (containing a XML document/schema) and `Pathname` (path to a file)
- Avoid problems with file names containing a new line character (really unlikely, but still
possible...)
- The `Yast::XML.validate` method has been added recently and my [quick search](https://github.com/search?q=org%3Ayast+%22XML.validate%22+org%3Ayast&type=Code) did not found any usage in YaST. Which means we can safely change the type and meaning of the arguments.

### DIFF
--- a/library/xml/src/modules/XML.rb
+++ b/library/xml/src/modules/XML.rb
@@ -23,6 +23,7 @@
 require "yast"
 
 require "nokogiri"
+require "pathname"
 
 module Yast
   # Exception used when trying to serialize a Ruby object that we cannot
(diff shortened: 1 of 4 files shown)


mvidner:
Thanks! Now that I think of it I recall a different pattern used in some libraries:
A `String` is a filename, and an `IO` is a file. So if you have in-memory data without a real file, you can use `StringIO`.
The advantage is that you can stream huge files without reading them all at once.

Just an idea, your design is fine.

lslezak:
I did some more research and unfortunately we cannot use `IO` easily here. The problem is that [the validator is created](https://github.com/yast/yast-yast2/blob/20f7696e2836cf93647d925c8ce612a8c2771666/library/xml/src/modules/XML.rb#L209)  inside a `Dir.chdir` block to ensure the included files are read relatively to the main schema file. So we need to know the directory where the file is stored.

But the generic `IO` class does not provide any method for getting a filename from an opened IO. (And in some cases it even does not make sense, like for `StringIO`.)

Alternatively we could use the `File` class (`IO` subclass) which provides the [#path](https://ruby-doc.org/core-2.5.0/File.html#method-i-path) instance method. That should be fine.

lslezak:
The question is what is better, String/Pathname or String/File?

lslezak:
Or alternatively we could remove that `Dir.chdir` functionality and move the responsibility to the caller. But I do not like that, it sounds too error prone, you can easily miss that...

mvidner:
Ah, the schema includes need a path. That's enough of a reason to stick with the Pathname|String design.

PR #1066, opened 2020-06-25
new Mode: autoyast_clone_system

schubi2:
Problem
=========
We do not really have a flag to recognize if the AY export functions are called while cloning the system
(via yast clone_system) or are called in the AY configuration UI. The Mode.config is set in both cases.

Solution
========
This now Mode:  autoyast_clone_system which checks the argument of the yast call.

### DIFF
--- a/library/general/src/modules/Mode.rb
+++ b/library/general/src/modules/Mode.rb
@@ -293,6 +293,12 @@ module Yast
       mode == "autoinst_config"
     end
 
+    # AutoYaST is running in the clone_system mode
+    # which is called by "yast clone_system"
+    def autoyast_clone_system
+      ARGV[0] == "clone_system"
+    end
+
     # test mode wrappers
 
     # Synonym of {#testsuite}.
--- a/library/general/test/mode_test.rb
+++ b/library/general/test/mode_test.rb
@@ -383,4 +383,15 @@ describe Yast::Mode do
       end
     end
   end
+
+  describe "#autoyast_clone_system" do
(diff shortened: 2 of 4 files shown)


library/general/src/modules/Mode.rb
```diff
@@ -293,6 +293,12 @@ def config
       mode == "autoinst_config"
     end
 
+    # AutoYaST is running in the clone_system mode
+    # which is called by "yast clone_system"
+    def autoyast_clone_system
+      ARGV[0] == "clone_system"
```
jreidinger, line 299: what about cloning in AY UI? there is also clone button and also menu option to clone system. So it is not called just in this one case. Maybe it will be better to set some flag in country when auto with change is called? So we know user set it in UI? https://www.rubydoc.info/github/yast/yast-yast2/Installation/AutoClient#change-instance_method

schubi2: Hm, I am expecting these problems in other modules too (e.g. I have seen it in the ssh export too ) So I would prefer a global place for that decision in Mode module.
If it is needed we also could set the Mode.clone_mode in the AY UI too, but up to now I do not see a need for it.
So I think we are more flexible the the "Mode solution".
Other opinion ?

jreidinger: what is actually goal? you need to use mode in every module or do you want to use mode in AY only when cloning? if later, then you can remember in AY for which modules UI is ran and base it on it?

schubi2:
We have come to the agreement that we do not need it anymore.

PR #1067, opened 2020-06-26
Reduce profile on-demand

imobachgs:
The AutoClient#export method can receive a hash as an argument. The idea is to support getting arguments like the `target`. See https://github.com/yast/yast-autoinstallation/pull/631 for further information.

### DIFF
--- a/library/general/src/lib/installation/auto_client.rb
+++ b/library/general/src/lib/installation/auto_client.rb
@@ -61,8 +61,10 @@ module Installation
       when "Import"
         import(param)
       when "Export"
+        target = param["target"] if param.is_a?(Hash)
+        target ||= "default"
         m = method(:export)
-        m.arity.zero? ? export : export(param)
+        m.arity.zero? ? export : export(target: target.to_sym)
       when "Summary"
         summary
       when "Reset"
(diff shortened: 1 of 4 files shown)


library/general/src/lib/installation/auto_client.rb
```diff
@@ -99,8 +100,9 @@ def import(_profile)
     #
     # The profile is a Hash or an Array according to the configuration item
     # `X-SuSE-YaST-AutoInstDataType`
+    # @param _params [Hash<String,Object>] Additional parameters
```
jreidinger, line 103: I would document such parameters with `@option` yardoc tag

imobachgs: Thanks for your review. This comment takes me to another question: would it be better to define this method as:

```ruby
export(target: target)
```

Using a `_params` allows us to extend it easier in the future (without having to touch this code anymore). But an explicit declaration could make the code more clear (and easier to read). After all, I do not expect to start adding arguments to `export` every month. ;-) And Ruby already offers quite some control about how to detect which arguments are supported.

Any though about this?

jreidinger: keywords based looks better for me. And having symbol as param also looks fine for me. It will also provide one point where default is defined if we convert all auto clients to objects.

PR #1068, opened 2020-06-29
Avoid failure with release notes

joseivanlopez:
## Problem

Logging crashes when trying to download release notes and the proxy was inoperative.

* https://bugzilla.suse.com/show_bug.cgi?id=1173447

## Solution

Avoid failure when the proxy is inoperative.

## Testing

Added new unit tests.


joseivanlopez:
MU: https://build.suse.de/request/show/221251

PR #1071, opened 2020-07-01
Better management of libzypp repovars

ancorgs:
## Problem

Libzypp URLs can include the so-called [Repository Variables](https://doc.opensuse.org/projects/libzypp/HEAD/zypp-repovars.html) (repovars for the friends). They can have several formats like `$variable`, `${variable}`, `${variable-word}` or `${variable+word}`. They can appear in many parts of the URL, including the host, the path and even the port! (well, see some clarifications about the port below)

Of course, those vars are expanded before the URL is used for any real connection.

Before this pull request, the `Repository` class used two attributes to store the expanded and raw version of those URLs as [Ruby's URI](https://ruby-doc.org/stdlib-2.7.1/libdoc/uri/rdoc/URI.html) objects. Unfortunately, having curly brackets as part of the host, the path or the port of an URL does not conform to [RFC3986](https://tools.ietf.org/html/rfc3986). As a consequence Ruby thrown exceptions that went directly to our user's face.

See https://bugzilla.opensuse.org/show_bug.cgi?id=1172867  (and also the corresponding Trello card at https://trello.com/c/8DgGs7J1/)

## Solution

Instead of using `URI` objects directly, introduce a new `ZyppUrl` class that offers all the functionality of the `URI` one (with the same API to keep compatibility) and, in addition, handles repovars correctly in most cases. The only situation that is not handled properly is when those vars are used as part of the port section of the URL (see `FIXME`), but I manually tested that such thing doesn't work in general. Despite what libzypp's documentation says, `zypper` fails to add a repository with such URL and it fails in all operations if the repository is added manually.

Moreover, with that new class there is no need to store both (raw and expanded) URLs, the `Repository` object stores only the raw one, since the expanded URL is something the new class can calculate from the raw version (asking libzypp, of course).

Last but not least, when a non-understandable URL is parsed the new class creates an empty URL as result, instead of throwing an exception. That should save us from crashing in front of user's face next time somebody invents a new crazy feature for libzypp URLs or if repovars are used as part of the port specification (assuming that will actually work in the future and will be used by someone).

## Testing

- Added unit tests to prove this can handle correctly URLs like `http://example.com/$var/oss`, `http://example.com/${var}/oss`,  `http://example.com/${var-word}/oss`, `http://example.com/${var+word}/oss`.
- Added unit test to prove that using repovars in the port section (like `http://example.com:$releaseport/path`) do not cause a crash (the URL is considered to be empty).
- Tested manually upgrading from a Leap 15.0 with the following repository: `https://download.opensuse.org/repositories/LibreOffice:/Factory/openSUSE_Leap_${releasever}`. See the following screenshots.

Without the patch:

repovars-crash

With the patch, the repository is correctly visualized in the list and the process can continue (at least, to the installation summary screen):

repovars-repolist

### DIFF
--- a/library/packages/src/lib/y2packager/zypp_url.rb
+++ b/library/packages/src/lib/y2packager/zypp_url.rb
@@ -19,8 +19,11 @@ module Y2Packager
   #
   # Libzypp URLs do not conform to rfc3986 because they can include the so-called
   # Repository Variables. Those vars can have several formats like $variable,
-  # ${variable}, ${variable-word} or ${variable+word}. They can appear in any
-  # component of the URL (host, path, port...) except in scheme, user or password.
+  # ${variable}, ${variable-word} or ${variable+word}. According to libzypp's
+  # documentation, the variables can appear in any component of the URL (host, path,
+  # port...) except in scheme, user or password. But, at the time of writing, zypper
+  # throws an "invalid port component" error when trying to use variables as part of
+  # the port, even with the most recent available version of libzypp.
   #
   # See https://doc.opensuse.org/projects/libzypp/HEAD/zypp-repovars.html
   #
@@ -85,6 +88,9 @@ module Y2Packager
     #    suggests and may affect the reliability of the parsing.
     # 2) Extract the port section before parsing the URL if it contains repovars, storing
     #    its value in a separate instance variable and restoring it on demand.
+    #
+    # Anyways, using variables in the port of an URL doesn't seem to really work in libzypp,
+    # although it's documented to be valid.
     #
     # @return [String]
     def port
--- a/library/packages/src/lib/y2packager/zypp_url.rb
+++ b/library/packages/src/lib/y2packager/zypp_url.rb
@@ -30,6 +30,7 @@ module Y2Packager
   #
   class ZyppUrl < SimpleDelegator
     Yast.import "Pkg"
+    include Yast::Logger
 
     # Repository schemes considered local (see #local?)
     # https://github.com/openSUSE/libzypp/blob/a7a038aeda1ad6d9e441e7d3755612aa83320dce/zypp/Url.cc#L458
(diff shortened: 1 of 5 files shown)


library/packages/src/lib/y2packager/zypp_url.rb
```diff
@@ -0,0 +1,157 @@
...
+    #
+    # @return [String]
+    def hostname
+      repovars_unescape(uri.hostname)
+    end
+
+    # See URI::Generic#hostname
+    #
+    # TODO: escaping does not work here because the port wouldn't accept the
+    # escaped characters either. We likely need to modify the regexp used by
```
dgdavid, line 69: Although is true that they say

> Flexibility to have an alternate URI::Parser (or just different patterns and regexp's).  — https://ruby-doc.org/stdlib-2.7.1/libdoc/uri/rdoc/URI.html

It isn't clear to me that we can provide to the parser the regexp to check the port. If I'm not wrong, only `ESCAPED`, `UNRESERVED`, `RESERVED`, `DOMLABEL`, `TOPLABEL`, `HOSTNAME` regexps [can be _redefined_](https://github.com/ruby/uri/blob/41c79e77628b2c65acc51e6a652641db490eaf8a/lib/uri/rfc2396_parser.rb#L334-L339), but not the one used to [check the `PORT`](https://github.com/ruby/uri/blob/41c79e77628b2c65acc51e6a652641db490eaf8a/lib/uri/rfc2396_parser.rb#L394).

ancorgs:
This pull request targets the master branch, I asked in the bug whether we should merge it into `SLE-15-SP2` as well. But the review process can start already (I will add the changelog and the version bump after maint-coord has decided about the branch).

library/packages/src/lib/y2packager/zypp_url.rb
```diff
@@ -0,0 +1,180 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2017-2020 SUSE LLC, All Rights Reserved.
```
dgdavid, line 2: NP: actually, this file didn't exist until now :)

library/packages/src/lib/y2packager/zypp_url.rb
```diff
@@ -0,0 +1,180 @@
...
+
+module Y2Packager
+  # This class represents a libzypp URL
+  #
+  # Libzypp URLs do not conform to rfc3986 because they can include the so-called
+  # Repository Variables. Those vars can have several formats like $variable,
+  # ${variable}, ${variable-word} or ${variable+word}. According to libzypp's
+  # documentation, the variables can appear in any component of the URL (host, path,
+  # port...) except in scheme, user or password. But, at the time of writing, zypper
+  # throws an "invalid port component" error when trying to use variables as part of
```
dgdavid, line 25: Quite interesting!

ancorgs:
No reply from maint-coord, so I decided to merge this to master only. We can always backported afterwards... although I doubt this is much useful for SLE (where I don't think anybody use repovars) or Leap (since self-update is not used by default).

PR #1073, opened 2020-07-07
 Command line interface: Do not start an UI while evaluating current …

schubi2:
…language settings (bsc#1173133).


schubi2:
Wrong tree

PR #1077, opened 2020-07-09
add space to SPACE_CHARS (bsc#1173907)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1173907

`SPACE_CHARS` https://github.com/yast/yast-yast2/blob/master/library/types/src/modules/String.rb#L42 doesn't include a space.

This makes it impossible to work with ESSIDs containing a space. See
https://github.com/yast/yast-network/blob/master/src/lib/y2network/widgets/wireless_essid.rb#L95-L97

## Solution

Add it.

## Evil side effects

AFAICS `SPACE_CHARS` is only used in `Yast::String.CPrint` and that's only used in the ESSID dialog. So I wouldn't expect any evil fallout by this change.


shundhammer:
No version bump and change log?

PR #1080, opened 2020-07-17
Wizard layout

joseivanlopez:
## Problem

YaST Control File has been extended and now there are more settings to configure the installation layout, see https://github.com/yast/yast-installation-control/pull/99. Before that, the installation layout could only be partially configured by means of the *installation_ui* setting.

Now there are more possibilities to configure the layout. Moreover, the layout should be configurable for both: the installation and the firstboot clients. There should be an easy way to configure the layout taking into account all the new possible variants.

## Solution

A new class (*UI::Wizards::Layout*) is provided to ease the layout configuration. For example, having a control file with these settings:

~~~
<globals>
  <installation_layout>
    <mode>title-on-top</mode>
    <banner>true</banner
  </installation_layout>
</globals>
~~~

the layout can be configured and run with:

~~~
layout = UI::Wizards::Layout.from_product_features

layout.open_wizard do
  # ...
end
~~~

VirtualBox_opensuse TW_23_07_2020_07_56_40

Also note that *ProductFeatures* has been adapted to use settings from the control file when the firstboot stage is running. This makes possible to configure the firstboot layout from its *firstboot.xml* control file.

## Testing

* Manually tested.
* Added unit tests.


library/control/src/modules/ProductFeatures.rb
```diff
@@ -188,10 +188,14 @@ def Restore
     end
 
     # Initialize the features structure if needed
+    #
+    # Note that the values are normally read from a control file or set to default values. But, in normal
```
jreidinger, line 192: I would use `@note` so it is properly rendered

library/general/src/lib/ui/installation/layout.rb
```diff
@@ -0,0 +1,166 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "ui/installation/layout_config"
+
+module UI
+  module Installation
+    # Class to define the wizard layout for the installation
```
jreidinger, line 25: is it really only for installation? Or it can be used also for firstboot or even for common wizards?

library/general/src/lib/ui/installation/layout.rb
```diff
@@ -0,0 +1,166 @@
...
+        # Creates a new layout with a left sidebar to place the installation steps
+        #
+        # @return [Layout]
+        def with_steps
+          config = LayoutConfig.new.tap { |c| c.mode = LayoutConfig::Mode::STEPS }
+
+          new(config)
+        end
+
+        # Creates a new layout with without a sidebar but with the title of the dialogs on the left
```
jreidinger, line 59: I do not get that `with without`

library/general/src/lib/ui/installation/layout.rb
```diff
@@ -0,0 +1,166 @@
...
+      #
+      # @return [Boolean]
+      def with_steps?
+        config.mode == LayoutConfig::Mode::STEPS
+      end
+
+      # Whether the layout has the title of the dialogs on the left
+      #
+      # @return [Boolean]
+      def with_title_on_left?
```
jreidinger, line 116: I would probably use shorted `title_on_left?`

library/general/src/lib/ui/installation/layout.rb
```diff
@@ -0,0 +1,166 @@
...
+    private
+
+      # @return LayoutConfig
+      attr_reader :config
+
+      # Constructor
+      #
+      # @param [Layout, nil] Layout configuration
+      def initialize(config = nil)
+        @config = config || LayoutConfig.new
```
jreidinger, line 162: well, I would not use nil at all here. Simply I would do  ( and it is also better rendered in yardoc )
```
      def initialize(config = LayoutConfig.new)
          @config = config
```

library/general/src/lib/ui/installation/layout.rb
```diff
@@ -0,0 +1,166 @@
...
+      def open_wizard(&block)
+        Yast::UI.SetProductLogo(banner?)
+
+        case config.mode
+        when LayoutConfig::Mode::STEPS
+          Yast::Wizard.OpenNextBackStepsDialog
+        when LayoutConfig::Mode::TITLE_ON_LEFT
+          Yast::Wizard.OpenLeftTitleNextBackDialog
+        else
+          Yast::Wizard.OpenNextBackDialog
```
jreidinger, line 139: To be honest I do not like much all these comparison to constant. It is a too much for me and here is also hidden default, which is not mentioned anywhere else. So what I propose is to create also class for mode that has those constants and it will return object with methods like `steps?`, `title_left?`, `title_top?` or wizard_method which can be used e.g. here as 
`Yast::Wizard.public_send(config.mode.wizard_method))`

jreidinger: ah, lower I see default, so here that else does not make sense at all for me and is just confusing. I would at least use when for title on top and else which raise for invalid value.

library/general/src/lib/ui/installation/layout_config.rb
```diff
@@ -0,0 +1,133 @@
...
+        @installation_layout = {}
+
+        @installation_layout[:mode] =
+          case installation_layout["mode"]
+          when "steps"
+            Mode::STEPS
+          when "title-on-left"
+            Mode::TITLE_ON_LEFT
+          when "title-on-top"
+            Mode::TITLE_ON_TOP
```
jreidinger, line 124: again I am missng here validation else. So we can quitckly get if mode is not valid or if mode can be empty, then it should use default or empty mode

library/general/src/lib/ui/installation/layout_config.rb
```diff
@@ -0,0 +1,133 @@
...
+          end
+
+        @installation_layout[:banner] = !!installation_layout["banner"]
```
jreidinger, line 127: so default for banner is true?

library/general/test/ui/installation/layout_config_test.rb
```diff
@@ -0,0 +1,156 @@
...
+          "banner" => true
+        }
+      end
+
+      let(:installation_ui) { nil }
+
+      it "sets the mode according to the mode option" do
+        config = described_class.from_product_features
+
+        expect(config.mode).to eq(UI::Installation::LayoutConfig::Mode::STEPS)
```
jreidinger, line 112: and with mode as object you can write something like `expect(config.mode).to be_steps`

library/general/src/lib/ui/installation/layout.rb
```diff
@@ -0,0 +1,166 @@
...
+    # Moreover, for each layout, a top banner can be added.
+    #
+    # @example
+    #   layout1 = UI::Installation::Layout.with_steps
+    #   layout2 = UI::Installation::Layout.with_title_on_left
+    #   layout3 = UI::Installation::Layout.with_title_on_top
+    #
+    #   layout3.show_banner
+    #
+    #   layout1.open_wizard do
```
jreidinger, line 42: To be honest I hope that this is not the main point to this layout. Why not add method also to Yast::Wizard? like OpenWithLayout? which just call this? or course lazy loading. As wizard is first class which you inspect if you want to open it.

library/general/src/lib/ui/installation/layout.rb
joseivanlopez: Well, it could be used at every place where a wizard can be used, but it is mainly designed for installation mode (basically installer and firstboot). I got inspired by *UI::InstallationDialog*. Some layout elements like the top banner or the sidebar with steps or placing the title on the left are only used by YasT in the installer and firstboot clients. For that reason this class is focused on configuring an "installation" layout. Moreover, the provided stylesheets usually only take care of such elements during the installation/firstboot (installation.qss). IMHO, opening it to a more generic layout might make confusion.

library/general/src/lib/ui/installation/layout.rb
joseivanlopez: Well, the main points of this layout class are:

* Allow all possible configurations for the installation/firstboot in a object oriented way.
* Automatically configure the layout according to the control file.
* Abstract to the user about some low level details like which Wizard to open depending on the configuration.
* Abstract to the user about low level details like how to show/hide a banner.

Yes, maybe we can add *Wizard#OpenWithLayout(layout)*.

library/general/src/lib/ui/installation/layout.rb
jreidinger: I am just a bit worried if we by naming not limit its future usage when there is no reason. What about namespace `Wizards`? So it will be `UI::Wizards::Layout` ? So it also clearly indicate that it is for wizard like dialogs.

joseivanlopez: Yes, thinking twice about, you are right. Although the idea was born for installation/firstboot, I don't see any reason to limit it. And yes, `UI::Wizards::Layout` looks a good place. Thanks!

library/general/src/lib/ui/wizards/layout.rb
```diff
@@ -0,0 +1,286 @@
...
+        TITLE_ON_TOP = :title_on_top
+
+        private_constant :STEPS, :TREE, :TITLE_ON_LEFT, :TITLE_ON_TOP
+
+        class << self
+          # Mode for a layout with a left sidebar
+          #
+          # @return [Mode]
+          def steps
+            new(STEPS)
```
jreidinger, line 63: well why methods and always same object. Why it cannot be immutable and something like 
```
STEPS = Mode.new(:steps)
TREE = Mode.new(:tree)
```
etc. ? And if you would like to have methods then `def tree; TREE; done`

joseivanlopez: Finally I have done something similar to your proposal, but saving them to variable instances of the class.

library/wizard/src/modules/Wizard.rb
```diff
@@ -606,6 +606,22 @@ def OpenLeftTitleNextBackDialog
       nil
     end
 
+    def OpenWithLayout(layout)
+      UI.SetProductLogo(layout.banner?)
+
+      if layout.mode.steps?
+        public_send(:OpenNextBackStepsDialog)
```
jreidinger, line 613: as you call it here in Wizard, public_send is not needed and you can simply use `OpenNextBackStepsDialog`

joseivanlopez: It complains because missing constant when calling `OpenNextBackStepsDialog` directly. Using `self.OpenNextBackStepsDialog` instead.

PR #1082, opened 2020-07-22
Add a basic CFA class to manage nsswitch.conf

dgdavid:
## Problem

Factory started to serve the default `nsswitch.conf` file under `/usr/etc/` instead of `/etc`. However, YaST will try to read `/etc/nsswitch.conf` even if it does not existe **yet**.

* https://bugzilla.suse.com/show_bug.cgi?id=1173119
* https://trello.com/c/uDtev8AO/1949-5-add-support-for-usr-etc-nsswitchconf

## Solution

Since `nsswitch.conf` is a [complex configuration file which does not allow overrides of single entries](https://en.opensuse.org/openSUSE:Packaging_UsrEtc#Variant_3), `/etc/nsswitch.conf` must contain all desired configuration when it exists. So, YaST will read the configuration from `/usr/etc/nsswitch.conf` when `/etc/nsswitch.conf` is not available but will **always write to `/etc/nsswitch.conf`**. That way, it ensures that **the first time the file is written** it will contain not only the modified values but also the rest of the default configuration.

## Tests

* Added a unit test

### DIFF
--- a/library/general/src/lib/cfa/nsswitch.rb
+++ b/library/general/src/lib/cfa/nsswitch.rb
@@ -107,7 +107,8 @@ module CFA
     #
     # @param db_name [String] the database name, e.g., "passwd" or "hosts"
     #
-    # @return [Array<String>] database service specifications
+    # @return [Array<String>, nil] database service specifications or nil if db_name is not found
+    # otherwise
     def services_for(db_name)
       databases[db_name]&.services
     end
@@ -138,15 +139,11 @@ module CFA
     #
     # @see CFA::BaseModel#save
     def save
-      return false unless modified?
-
-      new_content = @parser.serialize(data)
+      return unless modified?
 
       @parser.file_name = write_path if @parser.respond_to?(:file_name=)
-      @file_handler.write(write_path, new_content)
-      @current_content = new_content
-
-      true
+      @current_content = @parser.serialize(data)
+      @file_handler.write(write_path, @current_content)
     end
 
     # Whether the content has changed
--- a/library/general/src/lib/cfa/nsswitch.rb
+++ b/library/general/src/lib/cfa/nsswitch.rb
@@ -1,0 +1,273 @@
+# Copyright (c) [2020] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "cfa/base_model"
+require "yast2/target_file"
+
+Yast.import "FileUtils"
+
+module CFA
+  # Model to handle the Name Service Switch configuration file
+  #
+  # @note Since nsswitch.conf does not allow to overrides single entries, the content will be
+  #   read from /usr/etc/nsswitch.conf only when /etc/nsswitch.conf file does not exist yet.
+  #   In any case, changes will be always written to /etc/nsswitch.conf.
+  #
+  # @example Reading service specifications of a database
+  #   file = Nsswitch.new
+  #   file.load
+  #   file.services_for("hosts") #=> ["db", "files"]
+  #
+  # @example Writing service specifications for a database
+  #   file = Nsswitch.new
+  #   file.load
+  #   file.update_entry("hosts", ["db", "files"])
+  #   file.save
+  #
+  # @example Deleting a database entry
+  #   file = Nsswitch.new
+  #   file.load
+  #   file.delete_entry("hosts")
+  #   file.save
+  #
+  # @example Loading shortcut
+  #   file = Nsswitch.load
+  #   file.entries #=> ["passwd", "group", "shadow", "hosts", "networks", "aliases"]
+  class Nsswitch < BaseModel
+    include Yast::Logger
+
+    # Path to local configuration file
+    LOCAL_PATH = "/etc/nsswitch.conf".freeze
+    private_constant :LOCAL_PATH
+
+    # Path to vendor configuration file
+    VENDOR_PATH = "/usr/etc/nsswitch.conf".freeze
+    private_constant :VENDOR_PATH
+
+    class << self
+      # Instantiates and loads the file
+      #
+      # This method is basically a shortcut to instantiate and load the content in just one call.
+      #
+      # @param file_handler [#read,#write] something able to read/write a string (like File)
+      # @return [Nsswitch] File with the already loaded content
+      def load(file_handler: Yast::TargetFile)
+        new(file_handler: file_handler).tap(&:load)
+      end
+    end
+
+    # Constructor
+    #
+    # @param file_handler [#read,#write] something able to read/write a string (like File)
+    #
+    # @see CFA::BaseModel#initialize
+    def initialize(file_handler: Yast::TargetFile)
+      super(AugeasParser.new("nsswitch.lns"), load_path, file_handler: file_handler)
+    end
+
+    # Loads the file content
+    #
+    # If the file does not exist, consider it as empty.
+    #
+    # @see CFA::BaseModel#load
+    def load
+      super
+      fix_collection_names(data)
+      @current_content = @parser.serialize(data)
+    rescue Errno::ENOENT # PATH does not exist yet
+      self.data = @parser.empty
+      @loaded = true
+    end
+
+    # Returns a list of defined databases
+    #
+    # @return [Array<String>] List of database names
+    def entries
+      databases.keys
+    end
+
+    # Service specifications for given database name
+    #
+    # @param db_name [String] the database name, e.g., "passwd" or "hosts"
+    #
+    # @return [Array<String>] database service specifications
+    def services_for(db_name)
+      databases[db_name]&.services
+    end
+
+    # Update (or create if it does not exist yet) the entry for given database name
+    #
+    #
+    # @note Current services, if any, are completely replaced
+    #
+    # @see #entry_for
+    # @see DatabaseEntry#services=
+    #
+    # @param db_name [String] the database name, e.g., "passwd" or "hosts"
+    # @param services [Array<String>] service specifications for the database
+    def update_entry(db_name, services)
+      database = entry_for(db_name)
+      database.services = services
+    end
+
+    # Delete entry for given database name
+    #
+    # @param db_name [String] the database db_name, e.g., "passwd" or "hosts"
+    def delete_entry(db_name)
+      data.delete(database_matcher(db_name))
+    end
+
+    # Writes the current content to /etc/nsswitch.conf
+    #
+    # @see CFA::BaseModel#save
+    def save
+      return false unless modified?
+
+      new_content = @parser.serialize(data)
+
+      @parser.file_name = write_path if @parser.respond_to?(:file_name=)
+      @file_handler.write(write_path, new_content)
+      @current_content = new_content
+
+      true
+    end
+
+    # Whether the content has changed
+    #
+    # @return [Boolean] true if content has been modified; false otherwise.
+    def modified?
+      @current_content != @parser.serialize(data)
+    end
+
+  private
+
+    # The path for loading the content
+    #
+    # @return [String] path to local configuration file if it exists; path to vendor one otherwise
+    def load_path
+      Yast::FileUtils.Exists(LOCAL_PATH) ? LOCAL_PATH : VENDOR_PATH
+    end
+
+    # The path for writing the content
+    #
+    # @return [String] path to local configuration file
+    def write_path
+      LOCAL_PATH
+    end
+
+    # Return the CFA::Matcher for given database name
+    #
+    # @param db_name [String] the database name, e.g., "passwd" or "hosts"
+    #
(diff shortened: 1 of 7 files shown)


library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+    #
+    # @see #entry_for
+    # @see DatabaseEntry#services=
+    #
+    # @param db_name [String] the database name, e.g., "passwd" or "hosts"
+    # @param services [Array<String>] service specifications for the database
+    def update_entry(db_name, services)
+      database = entry_for(db_name)
+      database.services = services
+    end
```
dgdavid, line 128: ### :warning:  note that actions (a.k.a. reactions) are not supported. :warning: 

This could be problematic limitation since `#update_entry` has been designed to fully replace the service specifications for a database **without** taking care about actions. Then, previous configuration contains 

  ```
   hosts:          dns [!UNAVAIL=return] files
  ```

  and a `CFA::Nsswitch#update_entry("hosts", ["db", "dns", "files"])` is performed, the result will be a file containing 

  ```
   hosts:          db dns files
  ```

This is not a problem with the current [yast-pam/Nsswitch](https://github.com/yast/yast-pam/blob/96a24a906231730c90bca79d3754da9cc4e70b1c/src/modules/Nsswitch.rb) module implementation because it _just replace a fragment of a line in a text file_ with the given value, usually [the result of manipulate the previous read value](https://github.com/search?q=%22Nsswitch.WriteDb%22+user%3Ayast&type=Code).

Adding `#append_service_to` and `#prepend_service_to` **does not** help because the problem still being the same, actions are not supported and the code will fully replace the service specifications with the new ones.

Do you think it worth it to add support for such complex structure at this time? (I think so, but I'd like to know others POV before proceeding)

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+      return false unless modified?
+
+      new_content = @parser.serialize(data)
+
+      @parser.file_name = write_path if @parser.respond_to?(:file_name=)
+      @file_handler.write(write_path, new_content)
+      @current_content = new_content
+
+      true
+    end
```
dgdavid, line 150: An optional `path` param in both CFA::BaseModel methods, [#load](https://github.com/config-files-api/config_files_api/blob/3e71bed70411b6f7dc66a7c6b16d1ec1872dc8c7/lib/cfa/base_model.rb#L55-L59) and [#save](https://github.com/config-files-api/config_files_api/blob/3e71bed70411b6f7dc66a7c6b16d1ec1872dc8c7/lib/cfa/base_model.rb#L42-L45), could help to simplify this models in which is needed to force/fix different target for those actions.

What do you think, @jreidinger?

library/general/src/lib/cfa/nsswitch.rb
jreidinger: well, idea of CFA is to not modify what is not needed or what we do not understand. So CFA tries quite heavy to keep existing stuff ( including comments, etc. ). So I am not sure about support, but we should at least not break it.

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+    # Loads the file content
+    #
+    # If the file does not exist, consider it as empty.
+    #
+    # @see CFA::BaseModel#load
+    def load
+      super
+      fix_collection_names(data)
+      @current_content = @parser.serialize(data)
+    rescue Errno::ENOENT # PATH does not exist yet
```
jreidinger, line 94: well, this is a bit abuse of knowing file_handler, but currently I do not know better way to check by asking file handler if file exists. It is on my todo list.

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+    # @return [Array<String>] List of database names
+    def entries
+      databases.keys
+    end
+
+    # Service specifications for given database name
+    #
+    # @param db_name [String] the database name, e.g., "passwd" or "hosts"
+    #
+    # @return [Array<String>] database service specifications
```
jreidinger, line 110: please mention that it can also return nil if db_name is not found.

library/general/src/lib/cfa/nsswitch.rb
jreidinger: Well, thinking again about it, I see some similarities with other code we have. I think it is ok to remove actions as long as there is warning in UI that it will completely replace entries.

library/general/src/lib/cfa/nsswitch.rb
jreidinger: well, it is a bit tricky as CFA goal is carefully edit config file. So writing to different file is a bit tricky part, but all in all, I agree that it should be simplified. maybe that optional param that has precedence over `@file_path` is the way.

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+      @current_content = new_content
+
+      true
+    end
+
+    # Whether the content has changed
+    #
+    # @return [Boolean] true if content has been modified; false otherwise.
+    def modified?
+      @current_content != @parser.serialize(data)
```
jreidinger, line 156: what is goal of this method? It is quite costy, as it do complete serialization. So to shortcut in save is a bit strange.

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+    def save
+      return false unless modified?
+
+      new_content = @parser.serialize(data)
+
+      @parser.file_name = write_path if @parser.respond_to?(:file_name=)
+      @file_handler.write(write_path, new_content)
+      @current_content = new_content
+
+      true
```
jreidinger, line 149: It is different then BaseModel save, where it returns void and here it is boolean if it is modified. What is reason?

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+    end
+
+  private
+
+    # The path for loading the content
+    #
+    # @return [String] path to local configuration file if it exists; path to vendor one otherwise
+    def load_path
+      Yast::FileUtils.Exists(LOCAL_PATH) ? LOCAL_PATH : VENDOR_PATH
+    end
```
jreidinger, line 166: well, I worry this is kind of tricky, because if you use different load_path and write path. Result is that write path will be always complete file. Which I am not sure if it is goal of that vendor config files, where it expect to have in local path only modifications and not complete config file, or am I wrong?

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+      database = CFA::AugeasTreeValue.new(CFA::AugeasTree.new, db_name)
+      data.add("database[]", database)
+      DatabaseEntry.new(database)
+    end
+
+    # Available databases
+    #
+    # @return [Hash{String => DatabaseEntry}]
+    def databases
+      raw_databases.each_with_object({}) do |database, collection|
```
jreidinger, line 201: Do you measure performance in some real scenarios? As each call to databases recreates whole hash, so a quite performance suboptimal.

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+    private_constant :COLLECTION_KEYS
+
+    # Fix collection names adding the missing "[]"
+    #
+    # When a collection has only one element, Augeas does not add "[]" suffix, reason why is
+    # necessary to fix at least those collections that are going to be hitted for either, read or
+    # write.
+    #
+    # @param tree [AugeasTree,AugeasTreeValue]
+    def fix_collection_names(tree)
```
jreidinger, line 225: well, maybe we should add this helper to cfa base itself?

library/general/src/lib/cfa/nsswitch.rb
```diff
@@ -0,0 +1,273 @@
...
+      tree.data.each do |entry|
+        entry[:key] << "[]" if COLLECTION_KEYS.include?(entry[:key])
+        fix_collection_names(entry[:value].tree) if entry[:value].is_a?(AugeasTreeValue)
+        fix_collection_names(entry[:value]) if entry[:value].is_a?(AugeasTree)
+      end
+    end
+
+    # Internal class for managing database entries
```
jreidinger, line 233: :+1: that it operates on data itself and ot not try to make some fancy thing. Here we failed hard in past with ntp-client which contain a lot of buggy behavior.

library/general/test/cfa/nsswitch_test.rb
```diff
@@ -0,0 +1,190 @@
...
+      it "returns false" do
+        expect(subject.save).to eq(false)
+      end
+
+      it "does nothing" do
+        expect(file_handler).to_not receive(:write)
+        subject.save
+      end
+    end
+  end
```
jreidinger, line 106: as we quite depends on lense that is not under our control, but we need to be sure we use it correct for both save/load I prefer to do real file read/write and verify that it is what we expect. file handler MemoryFile is your friend. See https://github.com/config-files-api/config_files_api/blob/master/spec/augeas_writer_spec.rb

package/yast2.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Jul 22 16:08:16 UTC 2020 - David Diaz <dgonzalez@suse.com>
+
+- Add a CFA based class to manage nsswitch.conf (bsc#1173119).
```
jreidinger, line 4: well, I also expect that it will be used. So I would like to see also its usage.

library/general/src/lib/cfa/nsswitch.rb
dgdavid: > that optional param that has precedence over `@file_path` is the way.

That's what exactly I meant :)

library/general/src/lib/cfa/nsswitch.rb
dgdavid: I don't like it too much either, but this method is there to avoid writing to the file when actually there are not changes pending.

library/general/src/lib/cfa/nsswitch.rb
dgdavid: No, there is no strong reason. It was just because the use of `return false` made me think returning true when successfully. I'll change it to return `nil` when file not modified yet and keep the `@file_handler.write` line as the last line of the method. Although having a true/false as return value in a `#save` method looks reasonable to me.

library/general/src/lib/cfa/nsswitch.rb
dgdavid: > well, I worry this is kind of tricky, 

Yes, it is :(

> because if you use different load_path and write path. Result is that write path will be always complete file. 

Which is what we are looking for here.

> Which I am not sure if it is goal of that vendor config files, where it expect to have in local path only modifications and not complete config file, or am I wrong?

I think that's what we all expect, but is not true for this case (and probably others, but I don't know them yet). Maybe I should improve the PR description to explain it better, but basically there is a _variant_ of configuration files that are too complex and does not allow to override a single value in their local counterpart. So, once you want to change something but preserve the other values defined in the vendor file, you must make a full copy from /usr/etc/example.conf to /etc/example.conf file. If not, the default values will be used for entries not defined in the local config file.

This was tested by myself (see https://bugzilla.suse.com/show_bug.cgi?id=1173119#c2) and also documented in https://en.opensuse.org/openSUSE:Packaging_UsrEtc

I hope it is more clear now.

library/general/src/lib/cfa/nsswitch.rb
dgdavid: I was worried about this too, but I didn't notice a huge penalty when working with it in an `irb` console using my own /usr/etc/nsswitch.conf file.

For sure, it would be nice to optimize it, do you have any suggestion/idea to do so?

library/general/test/cfa/nsswitch_test.rb
dgdavid: I'll take a look, thanks!

package/yast2.changes
dgdavid: Yes, the idea is

* to adapt [yast-pam/Nsswitch.rb](https://github.com/yast/yast-pam/blob/96a24a906231730c90bca79d3754da9cc4e70b1c/src/modules/Nsswitch.rb) module to use it instead the SCR agent, or
* to directly use it replacing the [Nsswitch.ReadDb](https://github.com/search?q=%22Nsswitch.ReadDb%22+user%3Ayast&type=Code) and [Nsswitch.WriteDb](https://github.com/search?q=%22Nsswitch.WriteDb%22+user%3Ayast&type=Code) calls. 

What do you think would be better?

library/general/src/lib/cfa/nsswitch.rb
jreidinger: yes, clear now. So please document it also in code. So basically vendor config acts as defaults and user one as modified.....I am not sure if this whole concept is good idea as it is hard for user to see if there are new/changed options than with old .rpmsave files, but it is different topic.

library/general/src/lib/cfa/nsswitch.rb
jreidinger: well, you can cache it and when doing operations that modify database then clear cache

dgdavid: To be honest, I discarded this idea because I thought it does not worth it. Let's finish polishing the other aspects first and postpone this for the end, if you don't mind.

package/yast2.changes
dgdavid: As discussed in IRC, the first option is _the right one_ since 

> \<jreidinger> dgdavid: and note that we cannot replace all usage of NSSSwitch module as it is used also in samba, which is in perl.

dgdavid:
As suggested and agreed, finally the CFA model has been added as part of yast2-pam. See https://github.com/yast/yast-pam/pull/20.

PR #1083, opened 2020-07-23
Firewalld: Added method to check which zones have been modified from default settings

teclator:
## Problem

When the user ask about a reduced profile we export all the available zones even if not modified from the default values.

    https://trello.com/c/bLXXJXbk/1904-5-reduce-firewall-section-size

## Solution

Provide a method to check which zones have been modified from the default settings

Required by https://github.com/yast/yast-firewall/pull/134


imobachgs:
Oh, it looks like there is a conflict.

PR #1084, opened 2020-07-24
Update is_wsl to match osrelease spellings under both WSL1 and WSL2

jfkw:
Microsoft kernels are named "4.0-19041-Microsoft" on WSL1 and
"4.19.104-microsoft-standard" on WSL2. Match on lowercase "-microsoft".
(boo#1174183)

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Jul 24 08:06:27 UTC 2020 - Jeff Kowalczyk <jkowalczyk@suse.com>
+
+- update is_wsl function to match wsl1 and wsl2 osrelease spellings
+  (boo#1174183)
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


teclator:
It looks good to me, I just wondering if could make sense to differentiate between both versions but as it seems to not be a need yet we could just add this fix.

package/yast2.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Jul 24 08:06:27 UTC 2020 - Jeff Kowalczyk <jkowalczyk@suse.com>
+
+- update is_wsl function to match wsl1 and wsl2 osrelease spellings
+  (boo#1174183)
```
teclator, line 5:
```suggestion
  (boo#1174183)
- 4.3.17
```

teclator:
@jfkw thnx for the fix!

PR #1087, opened 2020-07-28
changes

jreidinger:
trello: https://trello.com/c/qlBmJwiq/1934-5-handle-new-exceptions-when-serializing-deserializing-xml


imobachgs:
It looks like RuboCop is not happy.

PR #1090, opened 2020-08-25
[AutoYaST] Profile element path

imobachgs:
## The problem

This PR introduces the concept of *profile element path*. Until now, we referred to an element's path in two different places:

* When dealing with ask-lists. In that situation, the path was something like `users,0,username`.
* When reporting profile errors. The equivalent to the previous example is `users[1] > username`.

Why not try to unify them?

## The solution

This PR introduces the `AutoinstProfile::ElementPath` class. This class abstracts the path concepts. The same representation will be used in ask-lists and error reporting.

xpath-autoinst-issues

## Related PRs

* https://github.com/yast/yast-autoinstallation/pull/672
* https://github.com/yast/yast-storage-ng/pull/1121


library/general/src/lib/installation/autoinst_profile/element_path.rb
```diff
@@ -0,0 +1,110 @@
...
+            element = (part =~ /\A\d+\Z/) ? part.to_i : part
+            all.push(element)
+          end
+          new(*parts)
+        end
+      end
+
+      # Constructor
+      #
+      # @param parts [Array<String>] Profile path parts
```
dgdavid, line 54: NP: Array<String, Integer>

library/general/src/lib/installation/autoinst_profile/element_path.rb
```diff
@@ -0,0 +1,110 @@
...
+      def initialize(*parts)
+        @parts = parts
+      end
+
+      # Returns the path parts
+      #
+      # @return [Array<String>] An array containing the path parts
```
dgdavid, line 61: NP: Array<String, Integer>

library/general/test/installation/autoinst_profile/element_path_test.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  describe "#join" do
+    subject(:path) { described_class.new("users", 1) }
+
+    it "returns a new profile path including all the parts" do
+      expect(subject.join("username")).to eq(described_class.new("users", 1, "username"))
+    end
+
+    context "when strings are paths are given" do
```
dgdavid, line 40: NP: strings and paths

dgdavid (approved):
LGTM, just a few minor comments

dgdavid (approved):

PR #1096, opened 2020-09-08
add symmetric crypting to GPG and Password Dialog

jreidinger:
trello: https://trello.com/c/TQPehs2U/2015-8-refactor-the-autoyast-profile-loading-process

goal: adding symetric gpg to gpg module, so it can be used as backend when autoyast profile is encrypted. Also add shared PasswordDialog.


imobachgs (approved):
LGTM

library/gpg/test/gpg_test.rb
```diff
@@ -0,0 +1,103 @@
...
+# find current contact information at www.suse.com.
+
+require_relative "test_helper"
+require "tempfile"
+
+Yast.import "GPG"
+
+describe Yast::GPG do
+  let(:file_content) do
+    "-----BEGIN PGP MESSAGE-----
```
lslezak, line 28: Um, why not having it as a plain data file? That would avoid that save/remove `around` block...

Like we have https://github.com/yast/yast-yast2/tree/master/library/packages/test/y2packager/zypp/test_repo/repodata or https://github.com/yast/yast-yast2/tree/master/library/packages/test/data/rpm

lslezak (commented):
Data file in a test?

library/gpg/test/gpg_test.rb
jreidinger: well, I consider it small enough, but moving is ok for me

library/general/src/lib/ui/password_dialog.rb
```diff
@@ -0,0 +1,86 @@
...
+      @label = label
+
+      super()
+    end
+
+    def dialog_content
+      res = VBox(
+        Left(Heading(@label)),
+        HBox(
+          MinWidth(18, Label(_("Password:"))),
```
lslezak, line 51: We already have [Label.Password](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Label.rb#L321) and [Label.ConfirmPassword](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Label.rb#L327), used like [here](https://github.com/yast/yast-users/blob/master/src/include/users/dialogs.rb#L517-L518).

lslezak (commented):
Label.Password and Label.ConfirmPassword

library/gpg/test/gpg_test.rb
```diff
@@ -0,0 +1,94 @@
...
+  describe "#encrypted_symmetric" do
+    let(:file_content) { "test\n" }
+
+    it "will create output file that is encrypted" do
+      Tempfile.open do |f|
+        path = f.path
+        f.close
+        f.unlink
+        described_class.encrypt_symmetric(@path, path, "test")
+        expect(described_class.encrypted_symmetric?(path)).to eq true
```
lslezak, line 90: The created temporary file is kept on the disk, it should be removed at the end.

library/gpg/test/gpg_test.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  describe "#encrypted_symmetric" do
+    let(:file_content) { "test\n" }
+
+    it "will create output file that is encrypted" do
+      Tempfile.open do |f|
+        path = f.path
+        f.close
+        f.unlink
```
lslezak, line 88: There is a small race condition between unlinking the temporary file and writing it. In theory some other process could create a file with the same name in the meantime...

Does GPG has force option to overwrite an existing file? Or output the encrypted file on stdout and redirect it with `>` to the target file...

lslezak (commented):
Small test issues...

lslezak (approved):

PR #1097, opened 2020-09-09
Update progress status when an error pop up is displayed (bsc#1175926)

lslezak:
## The Problem

- The download progress in the main window still shows the details about the previously downloaded file when an error pop up is displayed for a different file. That's quite confusing.
- See https://bugzilla.suse.com/show_bug.cgi?id=1175926

refresh_progress_broken

## The Fix

- Just reset the progress status before displaying the popup
- I wanted to display the URL of the failed file in the progress but unfortunately the URL parameter of the callback contains the repository URL (like `http://example.com`), not the failed file URL (like `http://example.com/repodata/repomd.xml`). That would be confusing so I decided to just clear the old label.

refresh_progress_fixed

## Additional Logging

- For easier debugging I have added a debug log to all callbacks so we can see the order in which they were called. Activated only when `Y2DEBUG_CALLBACKS` is set to `1`.

### DIFF
--- a/library/packages/src/modules/PackageCallbacks.rb
+++ b/library/packages/src/modules/PackageCallbacks.rb
@@ -141,7 +141,7 @@ module Yast
     # at start of file providal
     #
     def StartProvide(name, archivesize, remote)
-      log.debug("Starting callback #{self.class}::#{__method__}")
+      log.info("Starting callback #{self.class}::#{__method__}") if ENV["Y2DEBUG_CALLBACKS"] == "1"
       Builtins.y2milestone("StartProvide: name: %1, remote: %2", name, remote)
       if remote
         sz = String.FormatSizeWithPrecision(archivesize, 2, false)
@@ -172,7 +172,7 @@ module Yast
     # during file providal
     #
     def ProgressProvide(percent)
-      log.debug("Starting callback #{self.class}::#{__method__}")
+      log.info("Starting callback #{self.class}::#{__method__}") if ENV["Y2DEBUG_CALLBACKS"] == "1"
       Builtins.y2milestone("ProgressProvide: %1", percent)
       if @_provide_popup
         UI.ChangeWidget(Id(:progress), :Value, percent)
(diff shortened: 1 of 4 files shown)


library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -2402,6 +2467,7 @@ def ProgressProgress(id, val_raw, val_percent)
     # @param [Array<String>] stages Descriptions of the stages
     # @param [String] help Help text describing the process
     def ProcessStart(task, stages, help)
+      log.info("Starting callback #{self.class}::#{__method__}") if ENV["Y2DEBUG_CALLBACKS"] == "1"
```
jreidinger, line 2470: well, it looks like a lot of code and lot of duplication. What about adding hook when method is called?
see https://stackoverflow.com/questions/3236194/defining-method-called-how-do-i-make-a-hook-method-which-gets-called-every-t

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -44,6 +44,50 @@ class PackageCallbacksClass < Module
...
+
+        return
+      end
+
+      # add a new *_hook method as a wrapper for the original method,
+      # log the name of the called method
+      hook = <<-HOOK
+      def #{name}_hook(*params)
+        log.info("Starting callback #{self}::#{name}")
+        #{name}_without_hook(*params)
```
jreidinger, line 78: btw when you are at it, you can also log result of callback like
```
log.info("Starting callback #{self}::#{name}")
res = #{name}_without_hook(*params)
log.info("Callback #{self}::#{name} finished with #{res}")
res
```

library/packages/src/modules/PackageCallbacks.rb
```diff
@@ -44,6 +44,52 @@ class PackageCallbacksClass < Module
...
+        return
+      end
+
+      # add a new *_hook method as a wrapper for the original method,
+      # log the name of the called method
+      hook = <<-HOOK
+      def #{name}_hook(*params)
+        log.info("Starting callback #{self}::#{name}")
+        @result = #{name}_without_hook(*params)
+        log.info("Callback #{self}::#{name} returned: #{@result.inspect}")
```
jreidinger, line 79: ok, it needs to be `log.info("Callback #{self}::#{name} returned: #\{res\}")

jreidinger (approved):

PR #1098, opened 2020-09-16
Allow to hide dialog heading

joseivanlopez:
## Problem

A Wizard dialog always reserves space for its heading, even when the title is not set or it is set as empty. In the Expert Partitioner we want to remove the *Expert Partitioner* title and to place the menu bar at the very top. But, due to such reserved space for the heading, the menu bar starts just after some blank gap, which looks quite strange.

* Part of PBI: https://trello.com/c/dT7X5aC1/2045-menu-layout-rendering-improvements

_Keeping the heading text:_

expert_partitioner_installation_before

_[Setting the header as empty](https://github.com/yast/yast-storage-ng/commit/ffa251b77884ae3e237e8f97133baf11bdb34ad6):_

expert_partitioner_installation_no_title

## Solution

The heading of the dialog is now managed by a `ReplacePoint` widget. The content of the heading is only added when the title is not empty. Take into account that the *Release Notes* button belongs to the header, so this button also goes away when the title is not set.

NOTE: This changes only affects to the ncurses version. For the graphics dialog, see https://github.com/libyui/libyui-qt/pull/133.

Screenshot_tw_2020-09-16_14:36:33

### DIFF
--- a/library/wizard/src/modules/Wizard.rb
+++ b/library/wizard/src/modules/Wizard.rb
@@ -282,12 +282,8 @@ module Yast
             HSpacing(1),
             VBox(
               VSpacing(0.2),
-              HBox(
-                # translators: dialog title to appear before any content is initialized
-                Heading(Id(:title), Opt(:hstretch), _("Initializing ...")),
-                HStretch(),
-                ReplacePoint(Id(:relnotes_rp), Empty())
-              ),
+              # translators: dialog title to appear before any content is initialized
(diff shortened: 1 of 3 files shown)


dgdavid (approved):
LGTM

dgdavid:
@joseivanlopez,

I have checked it during the installation. Everything looks good, see the screenshots in the description.

So, if nobody has something against, you can go ahead and merge it.

PR #1099, opened 2020-09-18
Add support for nesting tables

joseivanlopez:
Part of the dirty prototype at `partitioner-ui-03`. This requires new libyui features that are not ready yet, see https://github.com/libyui/libyui/pull/171.

This PR is not intended to go to master yet.

### DIFF
--- a/library/cwm/src/lib/cwm/table.rb
+++ b/library/cwm/src/lib/cwm/table.rb
@@ -119,9 +119,17 @@ module CWM
   private
 
     def format_items(items)
-      items.map do |item|
-        Item(Id(item.first), *item[1..-1])
-      end
+      items.map { |i| format_item(i) }
+    end
+
+    def format_item(item)
+      content = item.dup
+
+      content[0] = Id(content[0])
+
+      content[-2] = format_items(content[-2]) if content[-2].is_a?(Array)
+
+      Item(*content)
     end
   end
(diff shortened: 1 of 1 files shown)


dgdavid (approved):
Not sure if I understood what are you doing here, but let's go ahead with the partitioner UI prototype.

joseivanlopez:
> Not sure if I understood what are you doing here, but let's go ahead with the partitioner UI prototype.

Basically this allows to have nested items.

ancorgs:
If I got it correctly, this basically works by allowing `#items` to contain nested arrays like this (imagine a table with two columns):

```
[
  [:id1, "plain", "row"]
  [:id2, "parent", "root",  [[:c1, "children1", "leaf"], [:c2, "children2", "leaf"]], :open]
]
```

So, basically, if the penultimate element of a row is an array, it assumes:
  - That array is a collection of nested rows
  - The last element is the open/closed flag

That was good enough for our quick and dirty prototype... but I would try something better in the definitive version. Something closer to what we have with `Tree` and `TreeItem`.

Let's say every entry of `#items` can contain an array (like the one with `:id1` in the example above) or an object of type `TableItem`. And then we define that `TableItem` class to be something more reasonable than just a nested array.

I will start implementing that, if nobody disagrees.

PR #1103, opened 2020-09-24
Do not crash when trying to parse "nil" XML file (bsc#1176593)

lslezak:
- See https://bugzilla.suse.com/show_bug.cgi?id=1176593
- YaST crashed because it wanted to parse `nil` file

## Test

- Tested manually with the latest SP3 build

### DIFF
--- a/library/control/src/modules/WorkflowManager.rb
+++ b/library/control/src/modules/WorkflowManager.rb
@@ -1230,6 +1230,11 @@ module Yast
       ret
     end
 
+    # Read and remember the registration requirement status from an installation
+    # control XML file.
+    # The stored values can be read by the WorkflowsRequiringRegistration() method.
+    # @param filename [String] path to the XML file
+    # @return [Boolean] true if the file has been read properly, false in case of an error
     def IncorporateControlFileOptions(filename)
       return false if filename.nil?
 
--- a/library/control/src/modules/WorkflowManager.rb
+++ b/library/control/src/modules/WorkflowManager.rb
@@ -1231,6 +1231,8 @@ module Yast
     end
 
     def IncorporateControlFileOptions(filename)
+      return false if filename.nil?
+
       begin
         update_file = XML.XMLToYCPFile(filename)
       rescue RuntimeError => e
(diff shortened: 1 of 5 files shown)


library/xml/src/modules/XML.rb
```diff
@@ -165,7 +165,7 @@ def YCPToXMLString(doc_type, contents)
     # @return [Hash] parsed content
     # @raise [XMLDeserializationError] when non supported XML is passed
     def XMLToYCPFile(xml_file)
-      raise XMLDeserializationError, "Cannot find XML file" if SCR.Read(path(".target.size"), xml_file) <= 0
+      raise XMLDeserializationError, "Cannot find XML file" if xml_file.nil? || SCR.Read(path(".target.size"), xml_file) <= 0
```
jreidinger, line 168: Here I would prefer ArgumentError as for me passing nil is not valid argument and has nothing to do with XMLDeserialization

jreidinger: Also it would easier distringuish between not found and nil file...and BTW when you touch it, can you add xml_file path to this exception message?

library/control/src/modules/WorkflowManager.rb
```diff
@@ -1231,6 +1231,8 @@ def WorkflowRequiresRegistration(src_id)
     end
 
     def IncorporateControlFileOptions(filename)
+      return false if filename.nil?
```
mvidner, line 1234: 1. how do we know that this is the guilty caller?
2. since we're touching the method, let's document its API and side effects

library/xml/src/modules/XML.rb
mvidner: RIght. When dealing with exceptions I find it useful to categorize them like the HTTP status codes: 4xx (app got bad data) or 5xx (app messed up by itself)
Therefore a separate Argument Error (500) seems useful.
And adding the file name is good too.

library/control/src/modules/WorkflowManager.rb
lslezak: 1. Backtrace of the raised exception
2. I'll try that...

mvidner: >     1. Backtrace of the raised exception

Duh. You're right. Now I looked too and found that the `nil` comes from `GetCachedWorkflowFilename` which is even documented.

>     2. I'll try that...

thanks!

mvidner (approved):

PR #1104, opened 2020-09-28
Fix selector without opt

imobachgs:
Fixes #1044 in case that the `opt` method is not defined.

### DIFF
--- a/library/cwm/src/lib/cwm/common_widgets.rb
+++ b/library/cwm/src/lib/cwm/common_widgets.rb
@@ -201,7 +201,7 @@ module CWM
   private
 
     def editable?
-      opt.include?(:editable)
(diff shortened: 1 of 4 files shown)


library/cwm/src/lib/cwm/common_widgets.rb
```diff
@@ -201,7 +201,7 @@ def value=(val)
   private
 
     def editable?
-      opt.include?(:editable)
+      respond_to?(:opt) && opt.include?(:editable)
```
joseivanlopez, line 204: Is not editable by default?

imobachgs: Yes, it is not editable by default.

joseivanlopez (approved):
LGTM

PR #1105, opened 2020-10-01
Remove rubocop susefirewall references

teclator:
The exceptions are not needed anymore as we already removed the files .. see #806


teclator:
There are some modules that still import "SuSEFirewall2", thus, we have decided to revert the drop by now.

PR #1108, opened 2020-10-07
bsc#1177405: AutoYaST crashes with an empty <raid_options> section

imobachgs:
Fixes [bsc#1177405](https://bugzilla.suse.com/show_bug.cgi?id=1177405). It does not try to import the information from a hash when it is not actually a hash.

```
<partition>
  <raid_options/>
</partition>
```

Is exported to: 

```ruby
{ raid_options: "" }
```

because there is not enough information to determine whether it is should be a string or a hash (the `raid_options` is supposed to be a hash).


ancorgs (approved):
LGTM

PR #1110, opened 2020-10-08
bsc#1177405: SectionWithAttributes#new_from_hashes accepts an Enumerable

imobachgs:
Extend #1108 solution to do support an enumerable.


teclator (approved):
LGTM

PR #1111, opened 2020-10-14
Support for nested items in CWM::Table

ancorgs:
This replaces #1099.

Adds a new `TreeItem` class that can be used to specify nested items.

See https://trello.com/c/g0CpGq6T/1915-3-partitioner-nested-tables

This is how `examples/object_api_table.rb` looks like now, after adding a second table to it with nested elements:

object_api_table

### DIFF
--- a/library/cwm/src/lib/cwm/table.rb
+++ b/library/cwm/src/lib/cwm/table.rb
@@ -2,6 +2,29 @@ require "abstract_method"
 require "cwm/custom_widget"
 
 module CWM
+  # An entry of a {Table}
+  class TableItem
+    # @return [String, Symbol]
+    attr_reader :id
+
+    # @return [Array]
+    attr_reader :values
+
+    # @see Table#items
+    # @return [Array<TableItem, Array>]
+    attr_reader :children
+
+    # @return [Boolean] whether children are initially expanded
+    attr_reader :open
+
+    def initialize(id, values, open: true, children: [])
+      @id = id
+      @values = values
+      @open = open
+      @children = children
+    end
+  end
+
   # Represents Table widget
   class Table < CustomWidget
     # @method header
@@ -19,10 +42,14 @@ module CWM
     #   end
     abstract_method :header
 
-    # gets initial two dimensional array of Table content
-    # one element in the first dimension contain as first element id and then
-    # rest is data in table, which can be e.g. terms. Then it have to be enclosed in
-    # `cell` term.
+    # Array with the table content
+    #
+    # Each element can be a {TableItem} or an Array
+    #
+    # If it's an array, the row contains no nested elements. The first element represents
+    # the id of the row and the rest is used as data for its cells. Those can be e.g. terms.
+    # Then it has to be enclosed in `cell` term.
+    #
     # @see for more complex example see examples directory
     # @note default value is empty array. It is useful when computation expensive content
     #   need to be set. In such case, it is better to keep empty items to quickly show table and
@@ -33,9 +60,11 @@ module CWM
     #   def items
     #     [
     #       [:first_user, "Joe", "Doe"],
-    #       [:best_user, "Chuck", "Norris"]
+    #       TableItem.new(:best_user, ["Chuck", "Norris"])
     #     ]
     #   end
+    #
+    # @return [Array<TableItem, Array>]
     def items
       []
     end
@@ -43,7 +72,7 @@ module CWM
     # change list on fly with argument. Useful when content of widget is changed.
     # @note items and change_items is consistent with ItemsSelection mixin, just format of
     #   items is different due to nature of Table content.
-    # @param items_list [Array<Array<Object>>] same format as {#items}
+    # @param items_list [Array] same format as {#items}
     def change_items(items_list)
       Yast::UI.ChangeWidget(Id(widget_id), :Items, format_items(items_list))
     end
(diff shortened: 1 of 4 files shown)


library/cwm/src/lib/cwm/table.rb
```diff
@@ -119,9 +148,16 @@ def multiselection?
...
-      end
+      items.map { |i| format_item(i) }
+    end
+
+    # @see #format_items
+    def format_item(item)
+      return Item(Id(item.first), *item[1..-1]) if item.is_a?(Array)
+
+      children = format_items(item.children)
+      open_tag = item.open ? :open : :close
```
shundhammer, line 159: that symbol needs to be :close**d** , not :close !

library/cwm/src/lib/cwm/table.rb
```diff
@@ -19,10 +42,14 @@ class Table < CustomWidget
...
-    # one element in the first dimension contain as first element id and then
-    # rest is data in table, which can be e.g. terms. Then it have to be enclosed in
-    # `cell` term.
+    # Array with the table content
+    #
+    # Each element can be a {TableItem} or an Array
+    #
+    # If it's an array, the row contains no nested elements. The first element represents
+    # the id of the row and the rest is used as data for its cells. Those can be e.g. terms.
+    # Then it have to be enclosed in `cell` term.
```
shundhammer, line 51: it **has* to be...

library/cwm/src/lib/cwm/table.rb
```diff
@@ -33,17 +60,19 @@ class Table < CustomWidget
     #   def items
     #     [
     #       [:first_user, "Joe", "Doe"],
-    #       [:best_user, "Chuck", "Norris"]
+    #       TableItem.new(:best_user, ["Chuck", "Norris"])
```
jreidinger, line 63: I am fine even if you declare plain array deprecated and use just that new table item. Just can you add helper for creating item? like `item(:best_user, ["Chuck", "Norris"])` ? or table_item if there are collision with UIShortcuts.

library/cwm/src/lib/cwm/table.rb
```diff
@@ -33,17 +60,19 @@ class Table < CustomWidget
...
     #     ]
     #   end
```
jreidinger, line 65: please add also example with nested items.

library/cwm/src/lib/cwm/table.rb
shundhammer: Actually, you don't even need to wrap it into that `Cell()` if it's jut a plain string. We can save that intermediate step if it's just an array of strings.

See also
  https://github.com/yast/yast-ycp-ui-bindings/blob/master/examples/Table-nested-items.rb#L48

Only if a cell needs an icon or a sort key it is necessary to use `Cell()` around the content.

ancorgs: Strictly speaking, `item` would not collide with the UIShortcuts because the shortcut is `Item` (with capital "I"). But maybe they are still too similar

ancorgs: Well, I see exactly the same happens with `cell` and `Cell`, so I guess it's fine.

library/cwm/src/lib/cwm/table.rb
ancorgs: Wrong symbol fixed.

On the other hand, the sort-keys were already supported  before my changes because each value can be a `Cell` element.

library/cwm/src/lib/cwm/table.rb
ancorgs: Fixed.

library/cwm/src/lib/cwm/table.rb
ancorgs: Added the `item` helper.

library/cwm/src/lib/cwm/table.rb
ancorgs: Done. Both here and in the `examples` directory.

library/cwm/src/lib/cwm/table.rb
```diff
@@ -109,6 +146,16 @@ def cell(*args)
       Yast::Term.new(:cell, *args)
     end
 
+    # helper to create a {TableItem}
+    # @param args content of item
```
jreidinger, line 150: maybe add link for arg to TableItem#initialize?

shundhammer (approved):
LGTM

library/cwm/src/lib/cwm/table.rb
```diff
@@ -119,9 +166,16 @@ def multiselection?
...
-      end
+      items.map { |i| format_item(i) }
+    end
+
+    # @see #format_items
+    def format_item(item)
+      return Item(Id(item.first), *item[1..-1]) if item.is_a?(Array)
+
+      children = format_items(item.children)
+      open_tag = item.open ? :open : :closed
```
jreidinger, line 177: open/closed is a tricky one, but looks correct. I need to google it :) https://ell.stackexchange.com/questions/15373/close-vs-closed-and-open-vs-opened

jreidinger (approved):
just NP

library/cwm/src/lib/cwm/table.rb
ancorgs: Done

library/cwm/src/lib/cwm/table.rb
shundhammer: It's not tricky at all; it's very simple English: "The door is open" vs. "The door is closed".

It might be a bit confusing that they are using a past participle ("closed" from "to close") as an adjective in one case and a plain adjective ("open", not "opened") in another. But languages are not always logical or consistent.

joseivanlopez (commented):
I guess we need a new version and the changelog entry. yast-storage-ng should update its dependency to this new version of yast2.

PR #1112, opened 2020-10-20
Update spec file and changelog

ancorgs:
## Problem

The changes introduced in #1111 missed the version bump and the changelog entry

## Solution

Add what was missing.


shundhammer (approved):
LGTM

PR #1113, opened 2020-11-05
CWM ComboBox: Query current items when extending the list with a new value.

teclator:
## Problem

When the user scan the available WiFI networks, the ComboBox is not populated with the results but only offers the orginal selection.

- https://bugzilla.suse.com/show_bug.cgi?id=1177137

## Solution

Query the list of items offered by the ComboBox in order to extend it with a new value.

### DIFF
--- a/library/cwm/src/lib/cwm/common_widgets.rb
+++ b/library/cwm/src/lib/cwm/common_widgets.rb
@@ -57,6 +57,13 @@ module CWM
       []
     end
 
+    # List the current items offered in the widget
+    #
+    # @return [Array<Array(String,String)>]
+    def current_items
+      Yast::UI.QueryWidget(Id(widget_id), :Items)
+    end
+
     # @return [WidgetHash]
     def cwm_definition
       super.merge(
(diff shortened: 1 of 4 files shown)


jreidinger (approved):

jreidinger (approved):

library/cwm/src/lib/cwm/common_widgets.rb
```diff
@@ -194,7 +201,8 @@ class ComboBox < AbstractWidget
     #
     # @param val [Object] Value to assign to the widget
     def value=(val)
-      change_items([[val, val]] + items) if editable? && !items.map(&:first).include?(val)
+      change_items([[val, val]] + current_items) if editable? && !current_items.map(&:first).include?(val)
```
joseivanlopez, line 204: This change makes some unit tests to fail. The method `#current_items` returns nil when running unit tests, so this line will fail when it tries to map over nil.

For example, it fails when testing this https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/widgets/fstab_options.rb#L566.

PR #1114, opened 2020-11-06
Fix #current_items

joseivanlopez:
## Problem

Last changes in CWM (see https://github.com/yast/yast-yast2/pull/1113) can make unit tests to fail. The method `#current_items` returns nil in unit tests, so the test will fail when `ComboBox#value=` tries to map over nil.

See https://github.com/yast/yast-yast2/pull/1113#pullrequestreview-525536006.

## Solution

Ensure that `#current_items` always returns a list.


ancorgs (approved):
LGTM

PR #1116, opened 2020-11-24
Revert current items

teclator:
## Problem

A [bug](https://bugzilla.suse.com/show_bug.cgi?id=1178832) was reported recently as consequence of #1044 . The problem is that when the ComboBox is editable and the value is not present in the list of items, the list of items is modified using `change_items` with the original items, a list that could have been modified like it is in the [yast-network essid widget](https://github.com/yast/yast-network/blob/941f20ab040e60fccbaf52bf7cf5261c629adcd3/src/lib/y2network/widgets/wireless_essid.rb#L83).

So, it was proposed to check the current widget items instead of the items list but as @shundhammer  pointed in https://bugzilla.suse.com/show_bug.cgi?id=1178832#c21, we created and inconsistency as we were not converting the list of `Items` to an `Array<string, string>`.

## Solution 

Revert the addition of the current_items method and fix the bug reported modifying [yast2-network](https://github.com/yast/yast-network/pull/1123) directly.

### DIFF
--- a/library/cwm/src/lib/cwm/common_widgets.rb
+++ b/library/cwm/src/lib/cwm/common_widgets.rb
@@ -61,7 +61,7 @@ module CWM
     #
     # @return [Array<Array(String,String)>]
     def current_items
-      Yast::UI.QueryWidget(Id(widget_id), :Items) || []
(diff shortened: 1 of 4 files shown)


shundhammer (approved):
LGTM :+1:

PR #1118, opened 2020-11-26
prepare usrmerge: install legacy symlinks via spec file only (boo#102…

lnussel:
https://bugzilla.suse.com/show_bug.cgi?id=1029961

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,10 +13,3 @@ Yast::Tasks.configuration do |conf|
   conf.install_locations["library/desktop/groups/*.desktop"] = File.join(Packaging::Configuration::DESTDIR, "/usr/share/applications/YaST2/groups/")
   conf.install_locations["library/general/hooks/README.md"] = File.join(Packaging::Configuration::DESTDIR, "/var/lib/YaST2/hooks/")
 end
-
-# define additional creation of legacy symlinks during installation
-task :install do
-  sh "/usr/bin/mkdir -p #{Packaging::Configuration::DESTDIR}/sbin"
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -13,10 +13,3 @@ Yast::Tasks.configuration do |conf|
   conf.install_locations["library/desktop/groups/*.desktop"] = File.join(Packaging::Configuration::DESTDIR, "/usr/share/applications/YaST2/groups/")
   conf.install_locations["library/general/hooks/README.md"] = File.join(Packaging::Configuration::DESTDIR, "/var/lib/YaST2/hooks/")
 end
-
-# define additional creation of legacy symlinks during installation
-task :install do
-  sh "/usr/bin/mkdir -p #{Packaging::Configuration::DESTDIR}/sbin"
```
jreidinger, line 19: can we detect that usrmerge variable also during runtime? As this install task is also used locally when testing ( not sure how much developers use this ).

lnussel: well, you could check whether /sbin is a symlink and skip the block then.

mvidner:
So usrmerge says that `/usr/sbin/foo` is canonical and `/sbin/foo` is a compatibility symlink, right?

I'm fine with that.

Rakefile
lslezak: I do not know where it might be used, but we should not rely on old symlinks anyway...

lslezak (approved):

jreidinger (approved):

PR #1122, opened 2020-12-08
add autowrapping to Yast2::Popup class

jreidinger:
how example looks like:

autowrap5


package/yast2.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Dec 11 15:56:52 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- Use Auto Wrapping of long lines for Yast2::Popup (bsc#1179893)
```
joseivanlopez, line 4: As commented on IRC, this PR does not fix bsc#1179893. Maybe we need to use another bug.

library/general/src/lib/yast2/popup.rb
```diff
@@ -23,6 +23,10 @@ class << self
       RICHTEXT_WIDTH = 60
       RICHTEXT_HEIGHT = 10
 
+      # Minimum width for auto wrapped labels. It is computed that max line or this
```
joseivanlopez, line 26: NP: I am not sure if this comment is clear. If I got it correctly, what about this?

~~~
# Minimum width for auto wrapped labels. This value is used when
# the label has longer lines.
~~~

package/yast2.changes
joseivanlopez: As Josef commented on IRC, now bsc#1179893 should be fixed because *Report* was adapted.

joseivanlopez (approved):
LGTM

PR #1124, opened 2020-12-15
Removed SCR agent .etc.inittab which is obsolete (bsc#1175494).

schubi2:
Because SysVinit is no longer supported.


dgdavid (approved):
Thanks @schubi2, it's nice to see obsolete code removed :) :ok_hand:
 
Just for reference, mention that the agent it's obsolete because [SysVinit](https://en.wikipedia.org/wiki/Init#SysV-style) is no longer supported.

PR #1125, opened 2020-12-22
bsc#1180142: do not use 'installation-helper' to create snapshots

imobachgs:
YaST uses an especial `installation-helper` binary to configure snapper and create snapshots during installation and offline upgrade. However, it does not work with LVM (at least not in all cases). This PR changes the `FsSnapshot` class to use the `snapper` binary (using the `--no-dbus` option) to create snapshots. The configuration phase is still done through `installation-helper`.

Additionally, it introduces a new kind of exception so errors when creating snapshots can be properly handled.

* Related to: https://github.com/yast/yast-update/pull/162 and https://github.com/yast/yast-installation/pull/901.

### DIFF
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -253,6 +253,9 @@ module Yast2
         raise SnapperNotConfigured unless configured?
 
         cmd = LIST_SNAPSHOTS_CMD.dup
+        # Add additional options only when running in normal model. Otherwise,
+        # the snapper version in the chroot might not understand them (e.g.,
+        # when upgrading from SLE 12 to SLE 15).
         cmd << " --disable-used-space" if Yast::Mode.normal
 
         out = Yast::SCR.Execute(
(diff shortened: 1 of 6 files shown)


library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -252,9 +252,12 @@ def create_post(description, previous_number, cleanup: nil, important: false)
       def all
         raise SnapperNotConfigured unless configured?
 
+        cmd = LIST_SNAPSHOTS_CMD.dup
+        cmd << " --disable-used-space" if Yast::Mode.normal
```
jreidinger, line 256: please document rationale here.

jreidinger (approved):

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -366,6 +373,23 @@ def update_etc_sysconfig_yast2
       def setup_snapper_quota
         Yast::Execute.on_target("/usr/bin/snapper", "--no-dbus", "setup-quota")
       end
+
+      # Parses the snapshots list
+      #
+      # @param content [String] Snapshots list content
+      # @return Array<Hash> An array of hashes containing the data for each snapshot
+      def parse_snapshots_list(content)
```
jreidinger, line 381: this method is kind of tricky as any parsing method, I think that one or two example lines can make it more clear what happens in method.

jreidinger (approved):

library/system/src/lib/yast2/fs_snapshot.rb
joseivanlopez: BTW, snapper offers a machine readable outuput, see http://snapper.io/2019/11/06/machine-readable-output.html.

PR #1131, opened 2021-01-20
Uniquely identify the Resolvable object (bsc#1178688, bsc#1176276)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1178688
- Uniquely identify the Resolvable object to avoid random failures

### Details

The `Resolvable` class uses the `kind`, `name`, `version`, `arch` and `source` attributes for mapping the Ruby objects to the libzypp objects. (It avoids using memory pointers or other "unsafe" options.) It assumed that a repository cannot contain the same package in exactly same version twice (or more than once).

But it turned out that this is not true for Product objects, the updates repository can contain several `sles-release` RPM packages defining the very same `SLES-15.2-0` product:

```
S | Name            | Type    | Version             | Arch   | Repository
--+-----------------+---------+---------------------+--------+------------
  | SLES            | product | 15.2-0              | x86_64 | sle-product
  | SLES            | product | 15.2-0              | x86_64 | sle-product
  | sles-release    | package | 15.2-52.1           | x86_64 | sle-product
  | sles-release    | package | 15.2-49.1           | x86_64 | sle-product
```

That means we need to add another attribute which is unique. And that's the path (location) in the repo on the medium (example: `./x86_64/yast2-4.1.77-lp151.2.19.1.rpm`). Zypp `Product` does not have a package path, but we can use it's reference package (usually the `*-release.rpm`).

## Dependencies

- This change needs a support in pkg-bindings, this PR is related to this one: https://github.com/yast/yast-pkg-bindings/pull/152

## Testing

- Added an unit test with testing data (manually tweaked, see the README.md there)
- Tested locally in an `irb` session
- Tested in SP2 by two L3 team members (https://bugzilla.suse.com/show_bug.cgi?id=1178688#c64, https://bugzilla.suse.com/show_bug.cgi?id=1178688#c67)

## Additional Changes

- I have removed the GPG signature from the `test_repo` testing repository. The used GPG key expired, libzypp could possibly reject it. Now the repository is unsigned, just like the new testing repository with the SLES products.


jreidinger (approved):

lslezak:
Submitted in https://build.suse.de/request/show/234300

PR #1136, opened 2021-02-03
Set the defaults for selinux variables

dgdavid:


dgdavid:
Since the absence of a boolean variable means _false_ and no setting `selinux_mode` [fallbacks to _disable_](https://github.com/yast/yast-security/blob/8dcbb8c8e0da07f6224db0a1bf4157e3dffe0d51/src/lib/y2security/selinux_config.rb#L157-L160), I think this PR makes no sense. 

Closing it.

PR #1137, opened 2021-02-04
Added supported migration "openSUSE Leap 15.3" -> SLES

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1181773
- In Leap 15.3 the product has been renamed from `openSUSE` to `Leap` :scream: 
- I'll keep also the old openSUSE migration (just in case :smiley:)
- Manually tested by Luboš (https://bugzilla.suse.com/show_bug.cgi?id=1181773#c2)

### DIFF
--- a/library/packages/src/lib/y2packager/product_upgrade.rb
+++ b/library/packages/src/lib/y2packager/product_upgrade.rb
@@ -41,7 +41,10 @@ module Y2Packager
       # this one is used when openSUSE is not available, e.g. booting SLE medium
       # (moreover the openSUSE medium should contain only one product so that
       # product should be used unconditionally)
-      ["openSUSE"]                                                        => "SLES"
+      ["openSUSE"]                                                        => "SLES",
+      # (installed) openSUSE Leap => (available) SLES,
+      # same as above, in 15.3+ the product has been renamed from "openSUSE" to "Leap"
+      ["Leap"]                                                            => "SLES"
     }.freeze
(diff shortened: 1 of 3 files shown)


library/packages/src/lib/y2packager/product_upgrade.rb
```diff
@@ -41,7 +41,10 @@ class ProductUpgrade
       # this one is used when openSUSE is not available, e.g. booting SLE medium
       # (moreover the openSUSE medium should contain only one product so that
       # product should be used unconditionally)
-      ["openSUSE"]                                                        => "SLES"
+      ["openSUSE"]                                                        => "SLES",
+      # (installed) openSUSE Leap => (available) SLES,
+      # same as above, in 15.3+ the product has been renamed from "openSUSE" to "Leap"
+      ["Leap"]                                                            => "SLES"
```
jreidinger, line 47: don't we need also openSUSE => Leap?

lslezak: Good question... But as we got no bug reports I hope we do not need that. The upgrade is covered quite well in openQA...

lslezak: And this is mainly for the SCC migration, you cannot do Leap -> Leap upgrade via SCC...

jreidinger (approved):

PR #1139, opened 2021-02-10
Merge SLE-15-SP1

imobachgs:
Merge #1125 and #1138 into the `SLE-15-SP2` branch.

### DIFF
--- a/library/system/src/lib/yast2/fs_snapshot.rb
+++ b/library/system/src/lib/yast2/fs_snapshot.rb
@@ -75,8 +75,8 @@ module Yast2
       "/usr/bin/snapper --no-dbus --root=%{root} --csvout list-configs " \
       "--columns config,subvolume | /usr/bin/grep \"^root,\" >/dev/null".freeze
 
-    CREATE_SNAPSHOT_CMD = "/usr/lib/snapper/installation-helper --step 5 --root-prefix=%{root} " \
-    "--snapshot-type %{snapshot_type} --description %{description}".freeze
+    CREATE_SNAPSHOT_CMD = "/usr/bin/snapper --no-dbus --root=%{root} create "\
+      "--type %{snapshot_type} --description %{description}".freeze
 
     LIST_SNAPSHOTS_CMD =
       "/usr/bin/snapper --no-dbus --root=%{root} --utc --csvout list --disable-used-space " \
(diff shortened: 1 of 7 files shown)


library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -75,8 +75,8 @@ class FsSnapshot
       "/usr/bin/snapper --no-dbus --root=%{root} --csvout list-configs " \
       "--columns config,subvolume | /usr/bin/grep \"^root,\" >/dev/null".freeze
 
-    CREATE_SNAPSHOT_CMD = "/usr/lib/snapper/installation-helper --step 5 --root-prefix=%{root} " \
-    "--snapshot-type %{snapshot_type} --description %{description}".freeze
+    CREATE_SNAPSHOT_CMD = "/usr/bin/snapper --no-dbus --root=%{root} create "\
+      "--type %{snapshot_type} --description %{description}".freeze
```
jreidinger, line 79: does snapper support such change? ( I mean from step 5 to create ) ?

library/system/src/lib/yast2/fs_snapshot.rb
```diff
@@ -331,7 +331,7 @@ def create(snapshot_type, description, previous: nil, cleanup: nil, important: f
         out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), cmd)
 
         if out["exit"] == 0
-          find(out["stdout"].to_i) # The CREATE_SNAPSHOT_CMD returns the number of the new snapshot.
+          all.last
```
jreidinger, line 334: where is that all defined? I do not see it here

library/system/src/lib/yast2/fs_snapshot.rb
imobachgs: Yes, actually it is the preferred way.

library/system/src/lib/yast2/fs_snapshot.rb
imobachgs: It is another class method (just like `find`): https://github.com/yast/yast-yast2/blob/b4d77b1629166dee60515414d7f023638d16c813/library/system/src/lib/yast2/fs_snapshot.rb#L260.

imobachgs: The problem is that the create snapshot command does not return the ID anymore like the installation-helper used to do.

jreidinger: OK, I just a bit worry that if someone adds caching to that method it start failing, but it is NP.

jreidinger (approved):

PR #1141, opened 2021-02-12
Fix bug introduced while adding auto wrapping

ancorgs:
#1122 introduced a bug in the Report module. See [here](https://github.com/yast/yast-yast2/pull/1122/files#diff-71d0b42ff18d870f2a798fdcfa236622d29ff278fb75e9a1a1d643e72d929d1cR445).

This should fix that bug.


dgdavid (approved):
LGTM

jreidinger (approved):

PR #1143, opened 2021-03-08
[ProductControl] Ask if user would like to continue when client is missing

dgdavid:
### Problem

> Yast-firstboot executes the clients enabled in its control file. But it shows an exception error when a client is not available (e.g., because the package is not installed). — _**extracted from https://trello.com/c/xA1YyIdv/ (internal link) description**_

### Solution

**Do not crash**, by checking if the client exists and asking the user what to do (continue or not) when it does not.

* If user decides to continue, the workflow will continue to _:next_ or _:back_ client/step, depending on the previous result
* If not, the installation/firstboot workflow will be completely _:abort_

### Tests

Only tested manually, since adding unit tests for this complex part of the code if out of the scope of this PR.

### Screenshots

| Before | After |
|-|-|
| Screenshot_sle15sp3_2021-03-08_09:41:51 | Screenshot_sle15sp3_2021-03-08_09:49:19 |

### Additional notes

After taking a look to the code, there are no doubts that [ProductControl.RunFrom](https://github.com/yast/yast-yast2/blob/4c985bb22eda5427a37b254609e9b9f822a43a1c/library/control/src/modules/ProductControl.rb#L1227-L1505) deserves a refactorization, getting rid of those _Builtins_ and extracting the logic to a specialized class. However, I conclude that postponing it for SP4 is a better idea.

---

Related links

* Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1180954
* GitHub issues: https://github.com/yast/yast-firstboot/issues/116

### DIFF
--- a/library/control/src/modules/ProductControl.rb
+++ b/library/control/src/modules/ProductControl.rb
@@ -1353,14 +1353,14 @@ module Yast
           Hooks.run("after_#{step_name}")
         else
           # Client not found. Ask the user if want to continue (related to bsc#1180954)
-          Builtins.y2error("Client '%1' not found", client_name)
+          log.error("Client '#{client_name}' not found")
 
           text = format(
             # TRANSLATORS: Message warning the user that a client is missing where %{client} is
             # replaced by the client name (e.g. "registration", "user")
             _(
               "Something went wrong and the expected '%{client}' dialog was not found.\n\n" \
-              "Would you like to continue anyway?"
+              "Would you like to skip dialog and continue anyway?"
             ),
             client: client_name
           )
@@ -1368,9 +1368,7 @@ module Yast
           continue = Yast2::Popup.show(text, buttons: :yes_no) == :yes
 
           if continue
-            Builtins.y2warning(
-              "Continuing the workflow after skipping the '%1' missing client", client_name
-            )
+            log.warning("Continuing after skipping the '#{client_name}' missing client")
             # If user decided to continue, uses the former_result (:next or :back)
             result = former_result
           else
--- a/library/control/src/modules/ProductControl.rb
+++ b/library/control/src/modules/ProductControl.rb
@@ -1353,6 +1353,8 @@ module Yast
           Hooks.run("after_#{step_name}")
         else
           # Client not found. Ask the user if want to continue (related to bsc#1180954)
+          Builtins.y2error("Client '%1' not found", client_name)
+
           text = format(
             # TRANSLATORS: Message warning the user that a client is missing where %{client} is
             # replaced by the client name (e.g. "registration", "user")
@@ -1365,8 +1367,16 @@ module Yast
 
           continue = Yast2::Popup.show(text, buttons: :yes_no) == :yes
 
-          # If user decided to continue, uses the former_result (:next or :back); :abort otherwise.
-          result = continue ? former_result : :abort
+          if continue
+            Builtins.y2warning(
+              "Continuing the workflow after skipping the '%1' missing client", client_name
+            )
+            # If user decided to continue, uses the former_result (:next or :back)
+            result = former_result
+          else
+            # :abort because user decided to not continue
+            result = :abort
+          end
         end
 
         Builtins.y2milestone("Calling %1 returned %2", argterm, result)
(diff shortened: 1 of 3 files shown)


library/control/src/modules/ProductControl.rb
```diff
@@ -1336,16 +1337,47 @@ def RunFrom(from, allow_back)
...
+            # TRANSLATORS: Message warning the user that a client is missing where %{client} is
+            # replaced by the client name (e.g. "registration", "user")
+            _(
+              "Something went wrong and the expected '%{client}' dialog was not found.\n\n" \
+              "Would you like to continue anyway?"
+            ),
+            client: client_name
+          )
+
+          continue = Yast2::Popup.show(text, buttons: :yes_no) == :yes
```
dgdavid, line 1368: Maybe we should something like  

```ruby
Yast2::Popup.show(text, buttons: { yes: Label.ContinueButton, no:  Label.AbortButton) == :yes
```

here to display the _Abort_ and _Continue_ buttons instead.

Thoughts?

library/control/src/modules/ProductControl.rb
```diff
@@ -1336,16 +1337,47 @@ def RunFrom(from, allow_back)
...
+        else
+          # Client not found. Ask the user if want to continue (related to bsc#1180954)
+          Builtins.y2error("Client '%1' not found", client_name)
+
+          text = format(
+            # TRANSLATORS: Message warning the user that a client is missing where %{client} is
+            # replaced by the client name (e.g. "registration", "user")
+            _(
+              "Something went wrong and the expected '%{client}' dialog was not found.\n\n" \
+              "Would you like to continue anyway?"
```
jreidinger, line 1363: I would say `Would you like to skip dialog and continue anyway?`

library/control/src/modules/ProductControl.rb
jreidinger: yes, continue/abort looks better for me. Question is if we should not use Report here as code is used also in other places and not just in firstboot.

library/control/src/modules/ProductControl.rb
```diff
@@ -1336,16 +1337,47 @@ def RunFrom(from, allow_back)
...
+          # This code will be triggered before the red pop window appears on the user's screen
+          Hooks.run("installation_failure") if result == false
 
-        Hooks.run("after_#{step_name}")
+          result = result.to_sym
+
+          Hooks.run("after_#{step_name}")
+        else
+          # Client not found. Ask the user if want to continue (related to bsc#1180954)
+          Builtins.y2error("Client '%1' not found", client_name)
```
shundhammer, line 1356: NP: This might be a good time to start using `log.error` (and `include YaST::Logger`, of course) instead of adding more old-style `Builtins.*`. But this is just a suggestion; feel free to ignore.

shundhammer (approved):
LGTM in general; just the one NP note.

library/control/src/modules/ProductControl.rb
dgdavid: You're completely right, I just wanted to no touch too much that code. But, as well as I did with the popup, using Yast2::Popup, I can start using here the YaST::Logger, as you suggest.

library/control/src/modules/ProductControl.rb
shundhammer: @jreidinger : I am not sure if that makes the user's decision any easier; it just complicates the question that we are asking. Remember that the user has to answer "Yes" or "No" here. Adding the aspect of skipping the dialog would make users want to get more options to choose from; but we have more options. Let's keep this simple.

And besides, it would of course be "skip **the** dialog". ;-)

library/control/src/modules/ProductControl.rb
dgdavid: So, maybe to apply this control/check when running in Stage.firstboot?

jreidinger: this check is useful also for other modes, so I would not do it. Popup is OK as in other cases it usually means bug in our control.xml which should be well tested by openqa.

jreidinger (approved):

PR #1149, opened 2021-03-29
Determine whether a widget is present/displayed

imobachgs:
This PR adds a method to determine whether a UI is actually present in the UI. We will use this code to fix [bsc#1184115](https://bugzilla.suse.com/show_bug.cgi?id=1184115), but it could be useful to fix other bugs as well.

[bsc#118411](https://bugzilla.suse.com/show_bug.cgi?id=1184115)

### DIFF
--- a/library/cwm/src/lib/cwm/abstract_widget.rb
+++ b/library/cwm/src/lib/cwm/abstract_widget.rb
@@ -191,7 +191,7 @@ module CWM
     # Determines whether the widget is currently displayed in the UI
     #
     # @return [Boolean]
-    def widget_present?
+    def displayed?
       Yast::UI.WidgetExists(Id(widget_id))
     end
 
--- a/library/cwm/src/lib/cwm/abstract_widget.rb
+++ b/library/cwm/src/lib/cwm/abstract_widget.rb
@@ -188,6 +188,13 @@ module CWM
       Yast::UI.SetFocus(Id(widget_id))
     end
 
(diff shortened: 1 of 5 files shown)


library/cwm/src/lib/cwm/abstract_widget.rb
```diff
@@ -188,6 +188,13 @@ def focus
       Yast::UI.SetFocus(Id(widget_id))
     end
 
+    # Determines whether the widget is currently displayed in the UI
+    #
+    # @return [Boolean]
+    def widget_present?
```
jreidinger, line 194: I am not sure if widget is not duplicte. Maybe simple `present?` will be enough?

imobachgs: Yes, I agree. But I fear that `present?` is already used (see https://github.com/yast/yast-network/blob/master/src/lib/y2network/widgets/wireless_scan_button.rb#L61). However, I could simply use "displayed?" or "drawn?"... or change the `WirelessScanButton`.

jreidinger: I like that `displayed?` That sounds good for me.

jreidinger: and when we are at it, you can extend documenation of value/value= by mentioning that it makes sense to call it only when displayed? is true

jreidinger (approved):

PR #1151, opened 2021-04-08
add initial user class based on getent

jreidinger:
trello: https://github.com/yast/yast-yast2/pull/new/users_api


joseivanlopez:
API looks ok to me.

library/general/src/lib/yast2/group.rb
```diff
@@ -0,0 +1,49 @@
+require "yast2/execute"
+
+module Yast2
```
ancorgs, line 3: Haven't we agreed on placing all YaST classes under Y2Something namespaces?

library/general/src/lib/yast2/group.rb
```diff
@@ -0,0 +1,49 @@
...
+  class Group
+    class << self
```
ancorgs, line 5: I'm not sure if I like to have so many class methods. In the past I used to use a similar approach, but now I feel is more correct to have a separate class for reading and querying, instead of a (ever growing) bunch of class methods.

library/general/src/lib/yast2/group.rb
jreidinger: Let me check it, but I think we agreed that 
Yast is for yast infrastructure and legacy code
Yast2 for code in Yast2.rpm
Y2Prefix for yast module code.

library/general/src/lib/yast2/group.rb
jreidinger: So having something like GroupReader? and GroupStorage? where the first has read and later has reset/all and then in yast2-users have also Writer?

library/general/src/lib/yast2/group.rb
imobachgs: Well, we tend to use `Y2Prefix` for code which is specific to an area (no matter in which module is placed). It is some kind of "logic" separation.

For instance, we have `Y2Network`, `Y2Autoinstallation`, `Y2Packager` , and so on. And you can find the `Y2Packager` module in yast2.rpm.

library/general/src/lib/yast2/group.rb
imobachgs: In network, we decided to separate reading/writing logic and, in the end, it looks like it was the right decision because it is easy to read/write the configuration from/to different sources. Of course, you can still keep class methods that hide those details.

Potentially, you could have a reader for `getent` and a separate one when `systemd-users` gets published :wink:.

ancorgs: Yes, that was my point. The classes `User` and `Group` living in yast-yast2 should be basically structs, with almost zero logic. Nothing about reading/writing or even about connecting objects.

For example, this code in the class `User` is already a problem to me (it couples the shared class `User` with the particular reader that is based in getent (which may not be the only reader):

```ruby
def users
  # lazy load to avoid circular deps
  require "yast2/user"
  User.all.select { |u| u.gid == gid || users_name.include?(u.name) }
end
```

library/general/src/lib/yast2/group.rb
```diff
@@ -0,0 +1,49 @@
...
+    def initialize(name, gid: nil, users_name: [])
+      @name = name
+      @gid = gid
+      @users_name = users_name
+    end
+
+    def users
+      # lazy load to avoid circular deps
+      require "yast2/user"
+      User.all.select { |u| u.gid == gid || users_name.include?(u.name) }
```
imobachgs, line 45: Bear in mind that you can only have a single representation of users/groups. For instance, it is not possible to have a separate representation of "current" and "target" states. I know, perhaps it is not useful, but I kind of like that approach that we followed in storage-ng and network-ng.

imobachgs: Additionally, what about having a `UsersCollection` class? It could define powerful query methods. Perhaps it is not that important at that point, but I just wonder whether it might be useful.

library/general/src/lib/y2user/configuration.rb
```diff
@@ -0,0 +1,52 @@
+module Y2User
```
joseivanlopez, line 1: I am not sure about using the same namespace in yast2 and yast-users-ng. I was discussing it with @ancorgs yesterday. I tend to see yast2 as a generic library with shared functionality. But using the same namespace as other modules looks like a collection of parts of modules that were moved for convenience.

Currently we have very varied namespaces in yast2: `Yast::Execute`, `Yast2::Popup`, `CWM::Table`. IMHO, we should decide about the namespace rules of code provided by yast2. For example, I like what we did for systemd classes, that is, to use `Yast2` as outer namespace and then creating inner namespaces if needed: `Yast2::CompoundAction`, `Yast2::Systemd::Service`, `Yast2::Systemd::Unit`, etc.

For the users classes, maybe we could have something like: `Yast2::Users::Configuration`, `Yast2::Users::User`, `Yast2::Users::Group`, etc. And yast2-users-ng can use `Yast2::Users` classes directly or provide new classes based on the yast2 ones (e.g., `Y2Users::User`) if needed.

What do you think?

jreidinger: to be honest I do not have strong opinion here, but what really puzzle me is that inconsistency. So we should not change it when we decide.

jreidinger: btw see https://github.com/yast/yast-yast2/pull/1151#discussion_r610461401

joseivanlopez: Yes, if fact that comment was what ring the bell. Having different modules populating the same namespace is something I don't like. I see it as kind of monkey patching. Anyway, if we have already decided to go with that approach, fine with me. But if there is no a clear criteria, then I would suggest to open the discussion the sooner the better.

library/general/src/lib/y2user/configuration.rb
```diff
@@ -0,0 +1,62 @@
...
+  # Holds references to elements of user configuration like users, groupps or passwords.
```
ancorgs, line 2: Typo: groupps

library/general/src/lib/y2user/reader/getent.rb
```diff
@@ -0,0 +1,117 @@
+require "yast2/execute"
+require "date"
+
+require "y2user/group"
+require "y2user/user"
+require "y2user/password"
+
+module Y2User
+  module Reader
```
ancorgs, line 9: Shouldn't this be `Y2User::Readers` (plural) in the same way we have `Y2Network::ConfigWriters`, `Y2Network::Dialogs`, `Y2Storage::Callbacks`, Y2Storage::Clients` and many more?

jreidinger: Maybe yes, I always think about it and never decide in just one way. So lets unify it.

library/general/src/lib/y2user/readers/users_simple.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module Y2User
+  module Readers
+    # Reads users configuration using old Yast Module UsersSimple.
+    class UsersSimple
+      def read_to(configuration)
+        users = Yast::UsersSimple.GetUsers
+        # TODO: only created users, not imported ones for now
+        users.each do |user|
+          configuration.users << User.new(configuration, user["uid"], gecos: [user["cn"]])
```
jreidinger, line 18: when writing this code I get idea that because configuration is passed, then user can auto register itself there. So it do configuration.users << self in constructor. What do you think about this idea?

ancorgs: I think it's acceptable

dgdavid: Just a question: should we check that the user is not already in the given configuration?

jreidinger: @dgdavid well, my idea is to have some validation method to check it as it is more tricky that it looks like as e.g. uid should not collide, but it is one tricky usage where it can be with e.g. user toor that has uid 0 and can act as root. So question is if user name is unique enough? What if there is ldap and local user with same name? What happen if it happen?

dgdavid: > So question is if user name is unique enough? What if there is ldap and local user with same name? 

So, perhaps we should use uid and username as _user primary key_. But not sure.

joseivanlopez: For me it is reasonable. And yes, we need some validations at some point.

BTW, when I was looking at this line, I have started asking myself if we could avoid the double explicit reference between User and Configuration (and the same for Password and Group). I mean, adding an user to a configuration should be enough to know the configuration that the user belongs to. For example, `User#configuration` could be a lazy method that finds the user object in any of the existing configurations. What do you think?

joseivanlopez: Ok, I wrote my comment after reading the code snipped but without reading your comment. Ok, doing it in the User constructor would also avoid to explicitly indicate the double reference.

joseivanlopez: BTW, what happens if we do:

~~~
user = User.new(config1, ...)
config2.users << user
~~~

Is that possible?

jreidinger: no, that should be prevented. Correct method is 
```
user = User.new(config1, ...)
user.clone_to(config2)
```

About idea with lazy evaluation I do not like it much as that references should be fast and can happen really often.
But I think some way to prevent such wrong usage would be nice. But maybe that validation is enough?

PR #1152, opened 2021-04-12
bsc#1184131: use /etc/login.defs for shadow configuration

imobachgs:
Back in 2019, we added support to read/write the shadow configuration from `/usr/etc` and `/etc` (see [bsc#1155735](https://bugzilla.suse.com/show_bug.cgi?id=1155735#c2)) for further details. This change was wrongly included in SLE 15 SP2 (and Leap 15.2): the shadow package included in that version only reads the settings from `/etc/login.defs`.

This PR partially reverts those changes, so YaST will only take `/etc/login.defs` into account.

* Replace `CFA::ShadowConfig` with `CFA::LoginDefs` in the `ShadowConfig` moduile (used by `yast2-users`).
* Drop `CFA::ShadowConfig`. It is only used in yast2-security. See https://github.com/yast/yast-security/pull/108.

Fixes [bsc#1184131](https://bugzilla.suse.com/show_bug.cgi?id=1184131).


library/general/src/modules/ShadowConfig.rb
```diff
@@ -18,13 +18,14 @@
 # find current contact information at www.suse.com.
 
 require "yast"
-require "cfa/shadow_config"
+require "cfa/login_defs"
 
 module Yast
   # This class allows to access the API to handle login.defs attributes from Perl
   #
+  # In SLE 15 SP2, it relies on CFA::LoginDefs instead of CFA::ShadowConfig like later versions.
```
lslezak, line 26: The `CFA::ShadowConfig` is still  used in SP1  and GA, ritght? So the correct wording should be
```suggestion
  # In SLE 15 SP2, it relies on CFA::LoginDefs instead of CFA::ShadowConfig like in the earlier releases (SP1, GA)
```
I'd make it more clean.

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Apr 12 15:12:41 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- The ShadowConfig module only considers the /etc/login.defs
+  file (bsc#1184131).
```
lslezak, line 5: It describes only half of the problem, I'd add more details:
```suggestion
  file (do not use unsupported /etc/login.defs.d/) (bsc#1184131).
```

library/general/src/modules/ShadowConfig.rb
lslezak: Ah, it seems I got it wrong, this looks correct, ignore this comment.

lslezak (approved):

PR #1156, opened 2021-04-15
Add a basic set of classes to register and present errors

imobachgs:
This PR is a follow-up of the discussion in https://github.com/yast/yast-network/pull/1204. It adds a set of simple classes to register and report errors. This API looks pretty much the same, but there are some internal differences.

**It is still a WIP, but feedback is welcome!**

The only noticeable change when it comes to the API is the concept of "location", which represents where the error was detected (a configuration file, an element of the AutoYaST profile, etc.).  At this point, it is just a string, but the plan is to have a better representation so you can easily group errors by location when presenting them to the user. However, the location is not mandatory.

<details>
<summary>Error Reporting in the Network Module</summary>

error-reporting
</details>

<details>
<summary>Error Reporting in the Network Module (multiple errors)</summary>

error-reporting
</details>

<details>
<summary>Code Example: Adding Issues</summary>

```ruby
require "yast"
require "yast2/issues"

issues = Yast2::Issues::List.new
# adding a generic issue
issues << Yast2::Issues::Issue.new("something went wrong", severity: :fatal) # a generic error
# adding a more specific issue
issues << Yast2::Issues::InvalidValue(
  "magic", fallback: "auto", location: "file:/etc/sysconfig/network/ifcfg-eth0:BOOTPROTO"
)

puts Yast2::Issues::Presenter.new(errors).to_plain
```
</details>

### This PR

- [x] Decide about a proper name. *Error* seems related to *exceptions* (*problem*, *issue*, ...)
- [x] Split the `errors.rb` into different files. Move them to `yast2`?
- [x] Add a mechanism to identify better the source of the error (the `ifcfg-eth0` file in this case).
- [x] Should `MissingValue` and `InvalidValue` be merged into a single class?
- [x] Provide a widget to display errors (?)
- [x] Add tests and documentation.
- [ ] Is OK to generate the message when building the object?
- [x] Move to `Y2Issues` (?)

### Near future

- [ ] Adapt `AutoinstIssues` to use this new mechanism.


library/general/src/lib/yast2/issues/invalid_value.rb
```diff
@@ -0,0 +1,42 @@
...
+require "yast"
+require "yast2/issues/issue"
+
+module Yast2
+  module Issues
+    # Represents a situation where an invalid value was given
+    class InvalidValue < Issue
+      # @param location [URI, String] Error location ("file:/etc/sysconfig/ifcfg-eth0:BOOTPROTO")
+      # @param value [#to_s] Invalid value
+      # @param fallback [#to_s] Value to use instead of the invalid one
```
jreidinger, line 31: severity not documented

library/general/src/lib/yast2/issues/issue.rb
```diff
@@ -0,0 +1,60 @@
...
+      include Yast::I18n
+
+      # @return [String,nil] Where the error is located.
+      attr_reader :location
+      # @return [String] Error message
+      attr_reader :message
+      # @return [Symbol] Error severity (:warn, :fatal)
+      attr_reader :severity
+
+      # @param [URI,String,nil] Where the error is located. Use a URI or
```
jreidinger, line 50: invalid yardoc, which param is described?

jreidinger:
in general I do not like much that so much nesting even for basic use case like in your example. What about Y2Issues namespace instead of Yast2::Issues ?
Regarding that widget, maybe I prefer popup class that is premate to be used, but of course it should be able to use directly presenter and prepared dialog only if user explicitly want to show it.

imobachgs:
> in general I do not like much that so much nesting even for basic use case like in your example. What about Y2Issues namespace instead of Yast2::Issues ?

I considered these classes to be "utility classes", and for that reason, I decided to put them there. However, I am fine with moving them to `Y2Issues`.

> Regarding that widget, maybe I prefer popup class that is premate to be used, but of course it should be able to use directly presenter and prepared dialog only if user explicitly want to show it.

Yes, the idea is to offer something ready to be used and, if you want something different, you have the pieces to build it for your use case.

library/general/src/lib/y2issues.rb
```diff
@@ -0,0 +1,48 @@
...
+# * A class to collect errors ({Y2Issues::List}).
+# * A presenter to help when presenting the issues to the user ({Y2Issues::Presenter}).
+#
+# @example Registering an error
+#   list = Y2Issues::List.new
+#   list << Y2Issues::Issue.new("Could not read network configuration", severity: :fatal)
+module Y2Issues
+  # Reports the errors to the user
+  #
+  # @see Y2Issues::Reporter
```
jreidinger, line 35: I would here at least document params and report output.

library/general/src/lib/y2issues/issue.rb
```diff
@@ -0,0 +1,66 @@
...
+    attr_reader :location
+    # @return [String] Error message
+    attr_reader :message
+    # @return [Symbol] Error severity (:warn, :error)
+    attr_reader :severity
+
+    # @param message [String] User-oriented message describing the problem
+    # @param location [URI,String,nil] Where the error is located. Use a URI or
+    #   a string to represent the error location. Use 'nil' if it
+    #   does not exist an specific location.
```
jreidinger, line 51: I would write `Use 'nil' if there are no specific location.'

library/general/src/lib/y2issues/issue.rb
```diff
@@ -0,0 +1,66 @@
...
+    # @param severity [Symbol] warning (:warn) or error (:error)
```
jreidinger, line 52: To be honest I more like that fatal as it is clear that it is kind of issue that cannot be continue or ignored.

library/general/src/lib/y2issues/location.rb
```diff
@@ -0,0 +1,84 @@
...
+    #   location.path #=> "partitioning,1,partition,0"
+    #   location.id   #=> "filesystem_type"
+    #
+    # @example File reference
+    #   location = Location.parse("file:/etc/sysconfig/network/ifcfg-eth0:BOOTPROTO")
+    #   location.type #=> "file"
+    #   location.path #=> "/etc/sysconfig/network/ifcfg-eth0"
+    #   location.id   #=> "BOOTPROTO"
+    #
+    # @param str [String] Path specification
```
jreidinger, line 58: where is this path specification defined? I kind of expect also method that given location to that path specificion, but hard to know as I do not know what is use case of it.

library/general/src/lib/y2issues/presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+        all_issues << "#{path}:#{messages}"
+      end
+
+      Yast::HTML.List(all_issues)
+    end
+
+    # Return issues grouped by location where they were found
+    #
+    # @return [Hash<(#parent,#section_name),Installation::AutoinstIssues::Issue>]
+    #         Issues grouped by AutoYaST profile section
```
jreidinger, line 112: probably not only by autoyast one, not?

library/general/src/lib/y2issues/presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+    def issues_by_location(issues)
+      issues.each_with_object({}) do |issue, all|
+        group = if issue.location
+          "#{issue.location.type}:#{issue.location.path}"
```
jreidinger, line 116: why not ID? I expect that this kind of string output is part of location itself.

library/general/src/lib/y2issues/presenter.rb
imobachgs: Ouch, copy&paste problem (the presenter is pretty much the same as [AutoinstIssues::IssuePresenter](https://github.com/yast/yast-yast2/blob/4f0560a55f0b4ece41ed50063fd7baf00cd8eedc/library/general/src/lib/installation/autoinst_issues/issues_presenter.rb).

library/general/src/lib/y2issues/presenter.rb
imobachgs: Because we use `type + path` to group errors. For instance, let's consider that we have errors:

* Error 1: `file:/etc/sysconfig/network/ifcfg-eth0:BOOTPROTO`
* Error 2: `file:/etc/sysconfig/network/ifcfg-eth0:STARTMODE`
* Error 3: `file:/etc/sysconfig/network/ifcfg-eth1:BOOTPROTO` (not it is in a different file).

Then, the presenters will group *eth0* and *eth1* errors. See [this example](https://user-images.githubusercontent.com/15836/115024267-c23e5200-9eb7-11eb-9b97-d580dfe43359.png).

jreidinger: then I expect that issue has method `group` that can be called there. In general I do not like this string manipulation from other class.

jreidinger:
In general I like it, just few final notes :)

library/general/src/lib/y2issues/location.rb
```diff
@@ -0,0 +1,92 @@
...
+    #   location.id   #=> "filesystem_type"
+    #
+    # @example File reference
+    #   location = Location.parse("file:/etc/sysconfig/network/ifcfg-eth0:BOOTPROTO")
+    #   location.type #=> "file"
+    #   location.path #=> "/etc/sysconfig/network/ifcfg-eth0"
+    #   location.id   #=> "BOOTPROTO"
+    #
+    # @param str [String] String representing the path
+    # @return [Location]
```
jreidinger, line 59: I would add here cross reference to `#to_s`.

library/general/src/lib/y2issues/presenter.rb
imobachgs: Initially, I had a method `#group`, but how the things are grouped (or not) are decided by the presenter. So I decided to keep this logic in the presenter.

jreidinger (approved):

PR #1161, opened 2021-05-03
[SLE-15-SP3] Do not convert result to sym directly

dgdavid:
## Problem

A call to `Convert.to_symbol` was wrongly changed in https://github.com/yast/yast-yast2/pull/1143, which makes the installation crash with a `NoMethodError (undefined method ``to_sym' for false:FalseClass)` when the result from a client is `false`

* https://bugzilla.suse.com/show_bug.cgi?id=1185561
* https://trello.com/c/Kq990Qba/4698-tw-p3-1185561-ruby-error-in-installation-after-partitioning

## Solution

A kind of revert to previous code and keep using `result = Convert.to_symbol(result)` instead `result = result.to_sym`

## Additional note

The Bugzilla reference for version 4.3.59 has been fixed, switching from _bsc#1180**59**4_ to _bsc#1180**95**4_

### DIFF
--- a/library/control/src/modules/ProductControl.rb
+++ b/library/control/src/modules/ProductControl.rb
@@ -1348,7 +1348,7 @@ module Yast
           # This code will be triggered before the red pop window appears on the user's screen
           Hooks.run("installation_failure") if result == false
 
-          result = result.to_sym
+          result = Convert.to_symbol(result)
 
           Hooks.run("after_#{step_name}")
         else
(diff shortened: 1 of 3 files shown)


wfeldt (approved):
lgtm

wfeldt:
Although I still don't know what exactly triggers the bug...

dgdavid:
> Although I still don't know what exactly triggers the bug...

I see. We can wait to know more before merging it.

wfeldt:
The provided logs show that this was triggered by a failing install step (https://bugzilla.suse.com/show_bug.cgi?id=1185545, seemingly) that wasn't mentioned in the bug.

PR #1162, opened 2021-05-05
Do not convert result to sym directly

dgdavid:
The same than #1161, but for `master` branch.

It restores a conversion to symbol using `Convert.to_symbol` instead of Ruby `#to_sym` method.

Related to https://bugzilla.suse.com/show_bug.cgi?id=1185561


wfeldt (approved):
lgtm

PR #1163, opened 2021-05-11
Fix entries for ES and ca_ES

javierllorente:
- Remove Catalonia (duplicate for ES)
- Update ca_ES to Spain following other Spanish languages (gl_ES, eu_ES)


ancorgs (approved):
Despite any political claim someone could do, the suggested changes are technically correct.

1) The Catalonian entry in `country.ycp` was a duplicated key, which was wrong
2) The Catalonian entry in `country_long.ycp` was inconsistent with all the other xx_ES entries.

jreidinger:
@ancorgs thanks

PR #1165, opened 2021-05-25
Add Equatable mixin

joseivanlopez:
## Problem

Objects redefining `#eql?` method should redefine `#hash` method too. These two methods must be paired, otherwise some actions like subtracting arrays could generate unexpected results.

* Related to https://bugzilla.suse.com/show_bug.cgi?id=1186082.

## Solution

Add a `Equatable` mixin for keep `#eql?` and `#hash` paired.

### DIFF
--- a/library/general/src/lib/yast2/equatable.rb
+++ b/library/general/src/lib/yast2/equatable.rb
@@ -1,0 +1,104 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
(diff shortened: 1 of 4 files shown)


dgdavid (approved):
LGTM

teclator:
LGTM

library/general/src/lib/yast2/equatable.rb
```diff
@@ -0,0 +1,104 @@
...
+        @eql_attrs += names
+      end
+    end
+
+    # Hash key to identify objects
+    #
+    # Objects with the same values for their eql_attrs have the same hash
+    #
+    # @return [Integer]
+    def hash
```
joseivanlopez, line 90: A possible improvement is to have a `#eql_hash` method like this:

~~~
def eql_hash
  ([[:class, self.class]] + self.class.eql_attrs.map { |m| [m, send(m)] }).to_h
end

def hash
  eql_hash.hash
end
~~~

This would allow to classes that include the mixing to fine tune its hash (i.e, its comparison). For example, imagine we have a class that contains an array attribute, and we don't care about the elements order:

~~~
class MyClass
  include Yast2::Equatable

  attr_reader :array
  eql_attr :array

  def eql_hash
    h = super
    h[:array] = array.sort_by(&:hash) if h.keys.include?(:array)
    h
  end
~~~

PR #1166, opened 2021-06-01
CFA: recognize more fields from login.defs

ancorgs:
This is a pre-requisite for https://github.com/yast/yast-users/pull/284

This repository contains a CFA class to deal with login.defs. But it only handles a limited set of configuration parameters. This adds `umask` to that set, since it's needed in order to configure useradd (at some point, the `umask` configuration was part of `/etc/default/useradd` but now it's only present in login.defs).

Taking the opportunity, this PR also adds another parameters that influence the behavior of useradd.


joseivanlopez (approved):
LGTM

PR #1167, opened 2021-06-02
Make Yast2::Equatable#hash to be fine tuned easelly

teclator:
Added improvement already mentioned in:

https://github.com/yast/yast-yast2/pull/1165#discussion_r639537928

Which will be very useful for adopting it in yast2-netnetwork (see https://github.com/yast/yast-network/pull/1228)

### DIFF
--- a/library/general/src/lib/yast2/equatable.rb
+++ b/library/general/src/lib/yast2/equatable.rb
@@ -82,7 +82,7 @@ module Yast2
       end
     end
 
-    # Return a Hash containing all the attributes and values used for comparing
+    # Returns a Hash containing all the attributes and values used for comparing
     # the objects as well as a :class key with the object class as the value.
     #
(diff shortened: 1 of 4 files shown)


library/general/src/lib/yast2/equatable.rb
```diff
@@ -82,13 +82,21 @@ def eql_attr(*names)
       end
     end
 
+    # Return a Hash containing all the attributes and values used for comparing
```
joseivanlopez, line 85: NP: Returns

joseivanlopez (approved):
LGTM

library/general/src/lib/yast2/equatable.rb
teclator:
```suggestion
    # Returns a Hash containing all the attributes and values used for comparing
```

PR #1170, opened 2021-06-03
AutoYaST: Allow blank values in SectionWithAttributes

joseivanlopez:
## Problem

Sometimes, an AutoYaST profile might contain blank values for some properties, for example:

~~~xml
<users config:type="list">
  <user>
    <username>tux</username>
    <password_settings>
      <expire/>
      <mx>120</max>
      <inact>5</inact>
    </password_settings>
  </user>
</user>
~~~

In the example above, *expire* is empty, which means that the account has no expiration. By default, the *SectionWithAttributes* class converts the empty values to *nil*, so there is no way to distinguish when a properly was set as blank or it was missing in the profile. Such a differentiation is important in some cases like the *expire* option, because in case of blank, an empty string should be written in the /etc/shadow file for the *Expiration Date* field in order to make the account to never expire. And in case of nil (missing in the profile), nothing should be explicitly written in the file, giving the default value set by commands like *useradd*.

## Solution

Now on, the attributes definition in *SectionWithAttributes* allows to indicate for each attribute whether blank values are allowed or not.

### DIFF
--- a/library/general/src/lib/installation/autoinst_profile/section_with_attributes.rb
+++ b/library/general/src/lib/installation/autoinst_profile/section_with_attributes.rb
@@ -124,10 +124,13 @@ module Installation
         # Description of the attributes in the section.
         #
         # To be defined by each subclass. Each entry contains a hash with:
-        #   * :name        : the mandatory key
-        #   * :xml_name    : an optional alternative key
-        #   * :allow_blank : whether blank values (i.e, "" or []) are allowed. If set to false, the
-        #                    the value of the attribute will be nil when a blank is detected.
+        #   * :name        : (mandatory) name of the attribute.
+        #   * :xml_name    : (optional) name of the attribute in the xml profile. This is useful
+        #                    when the attribute is required to be called different to the its xml
+        #                    name.
+        #   * :allow_blank : (optional, false by default) whether blank values (i.e, "" or []) are
+        #                    allowed. If set to false, then the value of the attribute will be nil
+        #                    when a blank is detected.
         #
         # @return [Array<Hash>]
         def attributes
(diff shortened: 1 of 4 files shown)


library/general/src/lib/installation/autoinst_profile/section_with_attributes.rb
```diff
@@ -123,8 +123,11 @@ class SectionWithAttributes
       class << self
         # Description of the attributes in the section.
         #
-        # To be defined by each subclass. Each entry contains a hash with the
-        # mandatory key :name and an optional key :xml_name.
+        # To be defined by each subclass. Each entry contains a hash with:
+        #   * :name        : the mandatory key
+        #   * :xml_name    : an optional alternative key
```
imobachgs, line 128: np: Actually, it is not exactly an "alternative", but the name of the element in the XML profile.

library/general/src/lib/installation/autoinst_profile/section_with_attributes.rb
```diff
@@ -123,8 +123,11 @@ class SectionWithAttributes
...
+        #   * :allow_blank : whether blank values (i.e, "" or []) are allowed. If set to false, the
+        #                    the value of the attribute will be nil when a blank is detected.
```
imobachgs, line 130: Which is the default value?

library/general/test/installation/autoinst_profile/section_with_attributes_test.rb
```diff
@@ -105,6 +106,94 @@ def attributes
...
+          it "assigns the given value to the attribute" do
+            group = GroupSection.new_from_hashes(hash)
+
+            expect(group.name).to eq("a name")
+          end
+        end
+      end
+    end
+
+    context "when an attribute does not allows blank" do
```
imobachgs, line 153: "does not allow"

imobachgs (commented):
Just minor comments about the documentation.

imobachgs (approved):

PR #1171, opened 2021-06-08
Filter sysctl files

imobachgs:
Ignore sysctl files that do not use the `.conf` exception. The only exception right now are `/boot/sysctl.conf-*` files.

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1187018.

### DIFF
--- a/library/general/src/lib/cfa/sysctl_config.rb
+++ b/library/general/src/lib/cfa/sysctl_config.rb
@@ -157,7 +157,9 @@ module CFA
       if Yast::FileUtils.IsFile(path)
         [path]
       elsif Yast::FileUtils.IsDirectory(path)
-        Yast::SCR.Read(Yast::Path.new(".target.dir"), path).map { |file| File.join(path, file) }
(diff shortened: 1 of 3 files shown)


library/general/src/lib/cfa/sysctl_config.rb
```diff
@@ -157,7 +157,9 @@ def file_paths_in(path)
       if Yast::FileUtils.IsFile(path)
         [path]
       elsif Yast::FileUtils.IsDirectory(path)
-        Yast::SCR.Read(Yast::Path.new(".target.dir"), path).map { |file| File.join(path, file) }
+        Yast::SCR.Read(Yast::Path.new(".target.dir"), path)
+          .select { |f| f.end_with?(".conf") }
```
jreidinger, line 161: please write that note about this suffix also to this place, so we know in future why only .conf are selected.

shundhammer (approved):
If none of our OBS check script demands a bug ID in the change log, LGTM.

jreidinger (approved):

PR #1175, opened 2021-06-17
Add riscv64 architecture helper

dirkmueller:
This can be used for riscv64 specific handling elsewhere.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jun 17 06:15:02 UTC 2021 - Dirk Müller <dmueller@suse.com>
+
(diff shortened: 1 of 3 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Thu Jun 17 06:15:02 UTC 2021 - Dirk Müller <dmueller@suse.com>
+
+- add riscv64 architecture helper
```
jreidinger, line 4: in general change looks good. Just please add jira entry or bugzilla number here to clarify why change is needed and please also bump version, so it gets auto submitted.
Thanks

dirkmueller: done

jreidinger (approved):

PR #1176, opened 2021-06-21
Extend issues

jreidinger:
Reason: For Y2Issue::List it is needed sometimes to concat more sources of issues and for such case it is useful to be able to concat Lists. Size is common convenient method too check amount of issues.

Related to work on y2users.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Jun 21 20:51:57 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
+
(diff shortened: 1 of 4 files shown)


imobachgs (approved):

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Jun 21 20:51:57 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
+
+- Y2Issue::List: Add method size and concat (related to
```
imobachgs, line 4: np: methods

PR #1177, opened 2021-06-22
Small reorganization in Y2Issues

ancorgs:
## Problem

During the partial re-implementation of yast2-users, we found a couple of things to be improved in Y2Issues.

1) The Y2Issues severity levels were `:warn` and `:fatal`. But other parts of YaST use slightly different conventions for similar concepts[1]:
  - In the module `Report` there are functions line, `Message`,  `Warning` and `Error`.
  - In the `<report>` section of the AutoYaST profile, it's possible to define the behavior for the following categories: errors, warnings, messages, yesno_messages.

2) The current implementation of `Y2Issues::Reporter` was too focused in a single use case - showing errors found while processing the AutoYaST profile and offering the user the opportunity to abort installation.

## Solution

- Renamed `:fatal` to `:error`.
- Changed behavior of `Y2Issue::Reporter` to be more generic, just displaying the errors
- Introduced `Y2Issue::AutoinstReporter` to retain the old behavior

## Pending

- Adapt yast2-network to the new API
- Adapt yast2-users (the new y2users branch) to the new API
- Manual testing

## Notes

[1] The installation proposals follow a different convention, but I don't think it's expected to use Y2Issues in that context. Just for completeness, let's list here the warning levels accepted for a proposal:

  - `:notice`
  - `:warning` (default)
  - `:error`
  - `:blocker`
  - `:fatal`

A `:blocker` will prevent the user from continuing the installation. If any proposal contains a blocker warning, the Accept button in the proposal dialog will be disabled - the user needs to fix that blocker before continuing.

`:fatal` is like `:blocker` but also stops building the proposal.

An `:error` does not prevent continuing of the installation, but shows a popup that an user has to confirm to continue with the installation.

### DIFF
--- a/library/general/src/lib/y2issues.rb
+++ b/library/general/src/lib/y2issues.rb
@@ -28,7 +28,7 @@
 #
 # @example Registering an error
 #   list = Y2Issues::List.new
-#   list << Y2Issues::Issue.new("Could not read network configuration", severity: :fatal)
+#   list << Y2Issues::Issue.new("Could not read network configuration", severity: :error)
 module Y2Issues
   # Reports the errors to the user
   #
--- a/library/general/src/lib/y2issues.rb
+++ b/library/general/src/lib/y2issues.rb
@@ -46,6 +46,7 @@ require "y2issues/list"
 require "y2issues/presenter"
 require "y2issues/location"
 require "y2issues/reporter"
+require "y2issues/autoinst_reporter"
 
 # Issues types
 require "y2issues/issue"
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,17 +1,17 @@
 -------------------------------------------------------------------
 Tue Jun 22 10:00:23 UTC 2021 - Ancor Gonzalez Sosa <ancor@suse.com>
 
-- Y2Issue::Issue: renamed severity "fatal" to "error", to be more
+- Y2Issues::Issue: renamed severity "fatal" to "error", to be more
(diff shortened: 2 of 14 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Tue Jun 22 10:00:23 UTC 2021 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Y2Issue::Issue: renamed severity "fatal" to "error", to be more
+  consistent with other parts of (Auto)YaST
+- Changed behavior of Y2Issue::Reporter to be more generic (related
```
imobachgs, line 6: Y2Issues

imobachgs (approved):

imobachgs (approved):

library/general/src/lib/y2issues.rb
```diff
@@ -46,6 +46,7 @@ def self.report(issues)
 require "y2issues/presenter"
 require "y2issues/location"
 require "y2issues/reporter"
+require "y2issues/autoinst_reporter"
```
ancorgs, line 49: @imobachgs I was thinking... should we also add a `Y2Issues.autoinst_report` analogous to the `Y2Issues.report` defined a few lines above?

Or maybe add a parameter to it to specify an alternative reporter (not sure exactly in which form)?

ancorgs:
Closed in favor of #1178

PR #1178, opened 2021-06-23
Small reorganization in Y2Issues

ancorgs:
## Related pull requests

- This replaces #1177 
- Adapt yast2-users (Y2Users branch): https://github.com/yast/yast-users/pull/314
- Adapt yast2-network: https://github.com/yast/yast-network/pull/1240

## Problem

During the partial re-implementation of yast2-users, we found a couple of things to be improved in Y2Issues.

1) The Y2Issues severity levels were `:warn` and `:fatal`. But other parts of YaST use slightly different conventions for similar concepts[1]:
  - In the module `Report` there are functions line, `Message`,  `Warning` and `Error`.
  - In the `<report>` section of the AutoYaST profile, it's possible to define the behavior for the following categories: errors, warnings, messages, yesno_messages.

2) The current implementation of `Y2Issues.report` was too focused in a single use case - showing errors found while processing the AutoYaST profile and offering the user the opportunity to abort installation.

## Solution

- Renamed `:fatal` to `:error`.
- Extended `Y2Issue.report` with two new arguments that allow to specify what to do depending on the severity of the issues.

## Testing

- Adapted unit tests
- Manually tested with y2users with different usages of the reporter. See screenshots.

reporter-ask

reporter-continue

## Notes

[1] The installation proposals follow a different convention, but I don't think it's expected to use Y2Issues in that context. Just for completeness, let's list here the warning levels accepted for a proposal:

  - `:notice`
  - `:warning` (default)
  - `:error`
  - `:blocker`
  - `:fatal`

A `:blocker` will prevent the user from continuing the installation. If any proposal contains a blocker warning, the Accept button in the proposal dialog will be disabled - the user needs to fix that blocker before continuing.

`:fatal` is like `:blocker` but also stops building the proposal.

An `:error` does not prevent continuing of the installation, but shows a popup that an user has to confirm to continue with the installation.

### DIFF
--- a/library/general/src/lib/y2issues/reporter.rb
+++ b/library/general/src/lib/y2issues/reporter.rb
@@ -37,7 +37,7 @@ module Y2Issues
     def initialize(issues, report_settings: Yast::Report.Export)
       textdomain "base"
       @presenter = Presenter.new(issues)
-      @level = issues.fatal? ? :error : :warn
+      @level = issues.error? ? :error : :warn
       @log, @show, @timeout = find_settings(report_settings, @level)
     end
 
@@ -55,7 +55,8 @@ module Y2Issues
 
     # Displays a pop-up containing the issues
     #
-    # It can behave in two different ways depending if a fatal issue was found:
+    # It can behave in two different ways depending if the issues are only
+    # warnings or an error was found:
     #
     # * Ask the user if she/he wants to continue or abort the installation.
     # * Display a message and only offer an 'Abort' button.
--- a/library/general/src/lib/y2issues/reporter.rb
+++ b/library/general/src/lib/y2issues/reporter.rb
@@ -22,6 +22,7 @@ require "y2issues"
 
 Yast.import "Label"
 Yast.import "Report"
+Yast.import "HTML"
 
 module Y2Issues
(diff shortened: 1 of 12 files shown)


imobachgs (approved):

library/general/src/lib/y2issues/reporter.rb
```diff
@@ -37,45 +38,101 @@ class Reporter
     def initialize(issues, report_settings: Yast::Report.Export)
       textdomain "base"
       @presenter = Presenter.new(issues)
-      @level = issues.fatal? ? :error : :warn
+      @level = issues.error? ? :error : :warn
       @log, @show, @timeout = find_settings(report_settings, @level)
```
ancorgs, line 42: @imobachgs I was thinking about this. We are deciding which AutoYaST settings to use (`<warnings>` vs `<errors>`) based on the severity. But I was wondering whether it would make sense to use `<yesno_messages>` in the cases in which we ask the user whether to continue or not (when `ask` is used for the corresponding severity).

ancorgs: And talking about honoring the AutoYaST settings, what's the rationale to ignore the timeout and enforce it to zero in for the "error" level?

PR #1181, opened 2021-07-13
Prevent Crash if Replacing Content of Nonexistent Replace Point

shundhammer:
_**This is the fix for SLE-15-SP3. The merge to master will follow.**_

## Trello

https://trello.com/c/wC5C2kRW

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1187676

## Problem

Under some circumstances, updating progress in the _Progress_ module can lead to a crash with a UI error "No widget with ID ...".

## Probable Cause

Another pop-up dialog may be in the way, e.g. during some libzypp callbacks. That will then be the topmost dialog, and of course that dialog does not have the expected widget tree, i.e. not the ReplacePoint that the _Progress_ module wants to modify with `UI.ReplaceWidget()`.

## Fix

Be more defensive in that _Progress_ module: Check if the expected widget exists in the current dialog before trying to exchange its content.

## Test

There was no test environment for the _Progress_ module at all: The only way of testing it was with a live YaST module. But that makes it complicated to get into all cases; and, worse for this bug, there is no way to enforce a popup dialog being in the way.

So I added a new manual test client for this. It's not realistically possible to do that with auto-tests, so this needs some manual interaction and _watching_ what happens.

The new test client is started from the test directory:

```
cd yast-yast2/library/wizard/test/manual
yast2 ./progress_demo
```

or, for NCurses:

```
yast2 ./progress_demo --ncurses
```

It first asks which progress scenario to use:

progress-demo-01-ask-type

Then it opens one of either:

progress-demo-02-simple-main
_Simple progress_

progress-demo-04-complex-main
_Complex progress_

In either case, there are some buttons which directly correspond to the exported actions of the _Progress_ module, and you can open one or more pop-up dialogs that get in the way (which is the purpose of this test for this bug). That pop-up has the same buttons to test the _Progress_ module's behaviour.

progress-demo-03-simple-popup
_With popup dialog in the way_

Clicking around on those buttons might not update the progress output correctly, but it should at least not crash with a UI exception and error dialog.

### DIFF
--- a/library/wizard/src/modules/Progress.rb
+++ b/library/wizard/src/modules/Progress.rb
@@ -20,11 +20,6 @@
 #
 # ***************************************************************************
 # File:  modules/Progress.ycp
-# Module:  Progress
-# Summary:  Progress bar
-# Authors:  Petr Blahos <pblahos@suse.cz>
-#
-# $Id$
 #
 # Functions for progress bar.<br>
 # <pre>
@@ -88,6 +83,8 @@ require "yast"
 
 module Yast
   class ProgressClass < Module
+    include Yast::Logger
+
     def main
       Yast.import "UI"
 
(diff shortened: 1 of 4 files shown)


wfeldt (approved):
lgtm

shundhammer:
Notice that even though this maintenance request was _declined_, it was superseded by https://build.suse.de/request/show/245245 which was accepted and also includes these changes.

PR #1190, opened 2021-07-26
[SLE-15-SP3] Discourage the use of old URL modules

dgdavid:
As suggested during the review of https://trello.com/c/BnMnGsu9/ (internal link), point developers to the right tools for dealing with (libzypp)URLs.

It mentions not only the [URI](https://ruby-doc.org/stdlib-3.0.2/libdoc/uri/rdoc/URI.html) Ruby standard module, but also the _recently_ added [`Y2Packager::ZyppUrl`](https://github.com/yast/yast-yast2/blob/SLE-15-SP3/library/packages/src/lib/y2packager/zypp_url.rb) (see #1071), which allows dealing with libzypp URLs.

### DIFF
--- a/library/types/src/modules/URL.rb
+++ b/library/types/src/modules/URL.rb
@@ -30,6 +30,13 @@
 require "yast"
 
 module Yast
+  # A module for dealing with URLs
+  #
+  # @note This is legacy code ported from the
+  # {https://news.opensuse.org/2007/08/29/yast-independence-of-ycp/ YCP} era. Its use is highly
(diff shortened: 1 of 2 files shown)


library/types/src/modules/URL.rb
```diff
@@ -30,6 +30,13 @@
 require "yast"
 
 module Yast
+  # A module for dealing with URLs
+  #
+  # @note This is legacy code ported from the
+  # {https://news.opensuse.org/2007/08/29/yast-independence-of-ycp/  YCP} era. Its use is highly
```
dgdavid, line 36: If there is a better URL for giving context about YCP... it is welcome :wink:

ancorgs (approved):
LGTM

PR #1191, opened 2021-07-26
Discourages the use of old URL modules

dgdavid:
Same than https://github.com/yast/yast-yast2/pull/1190, but for `master` (future `SLE-15-SP4`) branch.


ancorgs (approved):
LGTM

PR #1195, opened 2021-09-15
Add Y2Issues::WithIssues mixin

joseivanlopez:
Add *WithIssues* mixin, which makes easier to work with a list of issues.

Needed for https://github.com/yast/yast-users/pull/326.

### DIFF
--- a/library/general/src/lib/y2issues/with_issues.rb
+++ b/library/general/src/lib/y2issues/with_issues.rb
@@ -1,0 +1,48 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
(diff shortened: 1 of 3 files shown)


library/general/src/lib/y2issues/with_issues.rb
```diff
@@ -0,0 +1,34 @@
...
+# find current contact information at www.suse.com.
+
+require "y2issues"
+
+module Y2Issues
+  # Mixin that provides a helper method to work with a list of issues
+  module WithIssues
+    # Executes the given block passing a list of issues
+    #
+    # @return [Y2Issues::List] list of issues filled in the block execution
```
jreidinger, line 27: I would love to see example here. That always helps to imagine how to call it.

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Sep 15 13:58:56 UTC 2021 - José Iván López González <jlopez@suse.com>
+
+- Add WithIssues mixin to make easier to work with a list of
```
jreidinger, line 4: and here I would use full name `Y2Issues::WithIssues`

library/general/src/lib/y2issues/with_issues.rb
```diff
@@ -0,0 +1,48 @@
...
+module Y2Issues
+  # Mixin that provides a helper method to work with a list of issues
+  #
+  # @example
+  #   class Example
+  #     include Y2Issues::WithIssues
+  #
+  #     def do_something
+  #       with_issues do |issues|
+  #         issues < Y2Issues::Issue.new("can do nothing")
```
jreidinger, line 31: NP: btw when seeing this, maybe you can define in block context also method issue? so it can be `issues < issue("can do nothing")

jreidinger (approved):

library/general/src/lib/y2issues/with_issues.rb
joseivanlopez: Well, I think we do not need it at this moment. I would prefer to extend the mixin in the future if we see that something like that can help.

PR #1196, opened 2021-09-15
Avoid jsc#PM-1998 and jsc#PM-2612 references

afaerber:
The changelog should reference jsc#SLE- ticket numbers.

PR #1197, opened 2021-09-15
Replace PM-2620 with the SLE-20563 reference

imobachgs:
Follow-up of #1196. Replaces PM-2620 reference with the epic one.


shundhammer (approved):
LGTM

PR #1198, opened 2021-09-17
Fixed losing the product selection (bsc#1190228)

lslezak:
## The Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1190228
- The products from addon modules are not selected to install, see the screenshot

sp4_addons_not_selected

## Details

- It turned out that the products actually *are* selected after adding the repositories
- But the selection is lost when reaching the installation summary
- Further debugging revealed the the selection is lost after calling `Pkg.SourceStartCache` which (re)loads the repositories
- That's actually not needed during installation, that code is mainly intended for installed system
- In the past this was OK, maybe some libzypp behavior has been changed...

## The Fix

- The code which loads the repositories was fixed to avoid reloads when at least one repository is defined
  - During installation, where at least one repository is present, it will skip the repository load 
  - In installed system it will load the repositories unless they are already loaded

## Testing

- Tested manually in the latest SP4 build, all products are correctly selected to install

sp4_addons_selected

### DIFF
--- a/library/packages/src/modules/PackageSystem.rb
+++ b/library/packages/src/modules/PackageSystem.rb
@@ -62,9 +62,6 @@ module Yast
       # Has Pkg::TargetInit run?
       @target_initialized = false
 
-      # Has Pkg::SourceStartCache run?
-      @source_initialized = false
-
       Yast.include self, "packages/common.rb"
 
       @_rpm_query_binary_initialized = false
@@ -94,7 +91,8 @@ module Yast
     def EnsureSourceInit
       PackageLock.Check
 
-      if @source_initialized
+      # no repository present (not even a disabled one)
+      if Pkg.SourceGetCurrent(false).empty?
         # this way, if somebody closed the cache outside of Package
         # (typically in installation), we will reinitialize
         # it's cheap if cache is already initialized
@@ -107,11 +105,8 @@ module Yast
         EnsureTargetInit()
       end
 
-      sources = Pkg.SourceStartCache(true)
-
-      if Ops.greater_than(Builtins.size(sources), 0)
-        @source_initialized = true
-      else
+      # at least one enabled repository?
+      if Pkg.SourceGetCurrent(true).empty?
         # all repositories are disabled or no repository defined
         Builtins.y2warning("No package repository available")
       end
@@ -310,7 +305,8 @@ module Yast
     def Available(package)
       EnsureSourceInit()
 
-      if !@source_initialized
+      # at least one enabled repository present?
+      if Pkg.SourceGetCurrent(true).empty?
         # error no source initialized
         return nil
       end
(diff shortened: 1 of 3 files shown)


library/packages/src/modules/PackageSystem.rb
```diff
@@ -107,11 +105,8 @@ def EnsureSourceInit
         EnsureTargetInit()
       end
 
-      sources = Pkg.SourceStartCache(true)
```
mchf, line 110: so if i got it correctly this is the fix. Doesn't just deleting it influence running system use case?

library/packages/src/modules/PackageSystem.rb
```diff
@@ -107,11 +105,8 @@ def EnsureSourceInit
...
-
-      if Ops.greater_than(Builtins.size(sources), 0)
-        @source_initialized = true
-      else
+      # at least one enabled repository?
+      if Pkg.SourceGetCurrent(true).empty?
```
mchf, line 109: ... or is the magic done by this true / false parameter in `SourceGetCurrent`? May be it would deserve better name then

library/packages/src/modules/PackageSystem.rb
lslezak: No, the running system is handled in the [code above](https://github.com/yast/yast-yast2/pull/1198/files#diff-29b7e878d71aa5590e4369adc14dfb84d31a05246bc2496734a6c5d84559c064R95-R99), there is another `Pkg.SourceStartCache` call.

library/packages/src/modules/PackageSystem.rb
lslezak: That boolean parameter just defines whether the result should include all repositories or just enabled ones.

Actually the `Pkg.SourceStartCache` call is the same as `Pkg.SourceStartManager + Pkg.SourceGetCurrent` (see [here](https://github.com/yast/yast-pkg-bindings/blob/987c4933a9fa8935664bd35d08d4c5ce5d050e30/src/Source_Load.cc#L545-L550)).

Yes, the naming and that boolean parameter look strange, but that's a legacy from even the pre-libzypp era...

library/packages/src/modules/PackageSystem.rb
lslezak: And I have tested that in an installed system to verify that it works fine.

library/packages/src/modules/PackageSystem.rb
mchf: hmm ... i was thinking of some constants ... but that might be an overkill.

mchf (approved):
lgtm

library/packages/src/modules/PackageSystem.rb
kobliha: A constant with some self-explanatory name or at least a comment above the line would definitely help in the future. We'll be having issues when refactoring the code. $0.02

PR #1199, opened 2021-10-18
Remove `ProductControlProduct` (replaced by `ControlProductSpec`)

imobachgs:
This PR includes these changes:

- Remove `ProductControlProduct`.
- Move the `.forced_base_product` method to `ProductSpec`.
- Move the calculation of the registration arch to the `Yast::Arch` module.


library/general/src/modules/Arch.rb
```diff
@@ -512,6 +512,22 @@ def is_wsl
...
+      "s390_64" => "s390x",
+      # ppc64le is the only supported PPC arch, we do not have to distinguish the BE/LE variants
+      "ppc64"   => "ppc64le"
+    }.freeze
+    private_constant :REG_ARCH
+
+    # Returns the architecture expected by SCC
+    #
+    # @return [String] Architecture
+    def registration_arch
```
jreidinger, line 527: I am just not sure about name. As it is not just registration, but also rpm architecture. so I would call it like rpm_arch as it can be used for other use cases not just for scc/registration.

PR #1201, opened 2021-11-03
[RFC] Software management API proposal

imobachgs:
This PR implements a tentative API for software management. It is still a WIP, but it serves as an indication of the direction we are following.

* The entry point to the API is the `SoftwareManager` class, which offers an API to initialize the software stack, query for "resolvables" (packages and products), etc.
* Potentially, it supports multiple backends. Support for libzypp is implemented as any other backend.
* Classes like `Package` or `Product` should be backend-agnostic (not implemented yet).
* It implements a rich query API (see `SoftwareSearch`), although it is far from being finished.

## An example

```ruby
require "y2packager/software_manager"
require "y2packager/libzypp_backend"

manager = Y2Packager::SoftwareManager.new
manager.probe # Initialize the software stack
query = manager.search.with(kind: :product)
query.each do |product|
  puts product.name
end

# you can extend the query later
query.with(name: "SLES")
query.first&.name #=> "SLES"
```

## Pending

* [ ] PackageSearch:
  * [ ] Load all the attributes, including custom ones. See the existing Y2Packager::Resolvable class.
  * [ ] Allow searching for multiple states.
  * [ ] Offer a better way to define conditions (?)
* [ ] Ad-hoc installation of packages/products
* [ ] Implement the `SoftwareProposal` class and the corresponding `SoftwareManager#commit` method.

### DIFF
--- a/library/packages/src/lib/y2packager/backend.rb
+++ b/library/packages/src/lib/y2packager/backend.rb
@@ -1,0 +1,48 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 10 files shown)


imobachgs:
We are following a more conservative approach.

PR #1202, opened 2021-11-11
simplified slide show

jreidinger:
This PR works together with https://github.com/yast/yast-installation/pull/993 and https://github.com/yast/yast-packager/pull/590

It is about simplifying slide show to show less to avoid confusion when more parallel approach will be implemented.
Discussion about topic together with some youtube vides can be found at https://lists.opensuse.org/archives/list/yast-devel@lists.opensuse.org/thread/UPBBRPTTSLMI74PEVJVV6BYYRFI5UTBM/

So how it looks?

Before:
old

and new one:
qt_proposal_centered
and ncurses
ncurses_proposal_centered


joseivanlopez (commented):
Code looks ok, but please, add some description to explain the problem and how it was simpllified. BTW, it would be nice to have some comparative screenshot, if possible.

joseivanlopez (approved):
LGTM

PR #1205, opened 2021-11-30
Handle release notes error

imobachgs:
Related to [bsc#1193148](https://bugzilla.suse.com/show_bug.cgi?id=1193148).

Do not crash when it is not possible to fetch the release notes package. It does not fix the real issue (why is the release notes package missing?), but at least the installer does not crash.

## Details

The initial fix was rather easy: just catching `PackageDownloader::FetchError` and `PackageExtractor::ExtractionFailed`. However, these exceptions look like an implementation detail from the outside: the caller needs to know that `PackageExtractor` and `PackageDownloader` are used by `Product#download`.

So after a short discussion with @ancorgs, he suggested following the same approach as `Y2Storage`: define a set of [domain-specific errors](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/exceptions.rb). So, now, you do not know who raise the exception, but the error itself.

### DIFF
--- a/library/control/src/modules/WorkflowManager.rb
+++ b/library/control/src/modules/WorkflowManager.rb
@@ -447,11 +447,11 @@ module Yast
 
       log.info("installation.xml path: #{path}")
       path
-    rescue ::Packages::PackageDownloader::FetchError
+    rescue Y2Packager::PackageFetchError
       # TRANSLATORS: an error message
       Report.Error(_("Downloading the installer extension package failed."))
       nil
-    rescue ::Packages::PackageExtractor::ExtractionFailed
+    rescue Y2Packager::PackageExtractionError
       # TRANSLATORS: an error message
       Report.Error(_("Extracting the installer extension failed."))
       nil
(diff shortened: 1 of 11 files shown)


ancorgs (approved):
LGTM. Much better now.

library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb
```diff
@@ -157,6 +157,8 @@ def release_notes_content(package, prefs)
           package.extract_to(tmpdir)
           file, lang = release_notes_file(tmpdir, prefs)
           file ? [File.read(file), lang] : nil
+        rescue PackageDownloadError, PackageExtractionError => e
+          log.warn "Could not download and extract the package: #{e.inspect}"
```
lslezak, line 161: Should we display some error message to the user? I'm not sure if failing silently is a good idea here... :thinking:

lslezak (approved):
In general LGTM

library/packages/src/lib/y2packager/release_notes_fetchers/rpm.rb
imobachgs: I am not sure. For instance, if the package is missing from the repo, I am not sure whether the user can do something about it.

imobachgs: Finally, I have implemented your suggestion, but in the [yast2-installation](https://github.com/yast/yast-installation/pull/997) module.

imobachgs:
I am not bumping the version because I plan to release another fix today.

lslezak (approved):

PR #1207, opened 2021-12-01
bsc#1193212: add Product#register_target

imobachgs:
yast2.rpm counterpart to https://github.com/yast/yast-update/pull/175.


library/packages/src/lib/y2packager/product.rb
```diff
@@ -152,6 +156,7 @@ def initialize(name: nil, short_name: nil, display_name: nil, version: nil, arch
       @vendor = vendor
       @order = order
       @installation_package = installation_package
+      @register_target = register_target || ""
```
jreidinger, line 159: this looks a bit strange, so why not have "" in initializer? Also `@param` doc does not mention nil as possible value.

jreidinger (approved):

PR #1208, opened 2021-12-02
Popup.YesNo(): Unify the internal [No] button ID (bsc#1193326)

lslezak:
- Unify the internal button IDs
- The new [Yast2::Popup]() class use the `:yes` and `:no` IDs while the old `Popup` module uses the `:yes` and `:no_button` IDs
- That's inconsistent and makes testing via libyui REST API more difficult as you need to know which implementation is used to display a particular popup
- So let's use the same IDs everywhere :+1: 
- Additionally the code has been refactored and simplified

### DIFF
--- a/library/general/src/modules/Popup.rb
+++ b/library/general/src/modules/Popup.rb
@@ -250,31 +250,19 @@ module Yast
     #
     # @return [Yast::Term] button box
     def AnyQuestionButtonBox(yes_button_message, no_button_message, focus)
-      yes_button = Empty()
-      no_button = Empty()
+      yes_opts = [:okButton]
+      no_opts = [:cancelButton]
 
       if focus == :focus_no
-        yes_button = PushButton(Id(:yes), Opt(:okButton), yes_button_message)
-        no_button = PushButton(
-          Id(:no_button),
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Popup.rb
```diff
@@ -250,31 +250,19 @@ def NoHeadline
...
-        )
+        yes_opts << :default
       end
 
-      button_box = ButtonBox(yes_button, no_button)
-      deep_copy(button_box)
+      yes_button = PushButton(Id(:yes), Opt(*yes_opts), yes_button_message)
+      no_button = PushButton(Id(:no), Opt(*no_opts), no_button_message)
+
+      ButtonBox(yes_button, no_button)
```
jreidinger, line 265: it a bit scary me to refactor it without test coverage in https://github.com/yast/yast-yast2/blob/master/library/general/test/popup_test.rb but if you test it enough manually then it has green from me

jreidinger (approved):

library/general/src/modules/Popup.rb
lslezak: Tested manually, works fine.

PR #1211, opened 2021-12-13
Backports fix for bsc#1180142 to SLE 12

imobachgs:
Backport fix for [bsc#1180142](https://bugzilla.suse.com/show_bug.cgi?id=1180142) to SLE 12 SP3. Please, check #1125 for details about the original fix.

Related to:

* https://github.com/yast/yast-installation/pull/1003
* https://github.com/yast/yast-update/pull/176

Manually tested upgrading from SLE 12 SP2 to SLE 12 SP4.

PR #1212, opened 2021-12-14
Add a missing entry in the changes file

imobachgs:
While merging #1211 into `SLE-12-SP4`, I found out that #893 was not mentioned in the changes file.


teclator (approved):
LGTM

PR #1215, opened 2021-12-14
Added RelURL class (jsc#SLE-22669)

lslezak:
## Description

- Related to https://jira.suse.com/browse/SLE-22669
- This PR adds support for evaluating the `relurl://` URLs
- It will be used by other YaST modules (self-update)

## Details

- It internally uses the [File.expand_path](https://ruby-doc.org/core-2.5.0/File.html#method-c-expand_path) Ruby function to convert the relative path to absolute.
- The advantage is that it also supports going up in the tree using the parent directory notation (`../`)
- The disadvantage is that some corner cases must be handled specifically, e.g. it replaces `~` with the home directory path on the local system which is not what you would expect in URLs.

## Testing

- Tested with unit tests with lots of corner cases.

### DIFF
--- a/library/general/src/lib/yast2/rel_url.rb
+++ b/library/general/src/lib/yast2/rel_url.rb
@@ -17,7 +17,10 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "yast"
 require "uri"
+
+Yast.import "InstURL"
 
 module Yast2
   # Class for working with relative URLs ("relurl://")
--- a/library/general/src/lib/yast2/rel_url.rb
+++ b/library/general/src/lib/yast2/rel_url.rb
@@ -25,7 +25,11 @@ Yast.import "InstURL"
 module Yast2
   # Class for working with relative URLs ("relurl://")
   class RelURL
-    attr_reader :base, :relative
+    # @return [URI] the input base URL
+    attr_reader :base
+
+    # @return [URI] the input relative URL
+    attr_reader :relative
 
     # Is the URL a relative URL?
     #
@@ -37,7 +41,8 @@ module Yast2
 
     # Create RelURL object with URL relative to the installation repository
     #
-    # @param rel_url [String, URI] the relative URL
+    # @param rel_url [String, URI] the relative URL, if non-relative URL is used
+    #  then the result is this URL
     # @return [RelURL]
     #
     # @note Works properly only during installation/upgrade, do not use
@@ -53,11 +58,14 @@ module Yast2
     # @param rel_url [String,URI] the relative URL, it should use the "relurl://"
     #  schema otherwise the base URL is ignored
     def initialize(base_url, rel_url)
-      @base = import_url(base_url)
-      @relative = import_url(rel_url)
+      @base = URI(base_url).dup
+      @relative = URI(rel_url).dup
+
+      preprocess_url(base)
+      preprocess_url(relative)
     end
 
-    # Build and absolute URL
+    # Build an absolute URL
     #
     # @param path [String,nil] optional URL subpath
     # @return [URI] the absolute URL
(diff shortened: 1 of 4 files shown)


library/general/src/lib/yast2/rel_url.rb
```diff
@@ -0,0 +1,130 @@
...
+    # Create RelURL object with URL relative to the installation repository
+    #
+    # @param rel_url [String, URI] the relative URL
+    # @return [RelURL]
+    #
+    # @note Works properly only during installation/upgrade, do not use
+    #   in an installed system.
+    def self.from_installation_repository(rel_url)
+      base_url = Yast::InstURL.installInf2Url("")
+      new(base_url, rel_url)
```
jreidinger, line 47: what will happen if nont relative url is called?

library/general/src/lib/yast2/rel_url.rb
```diff
@@ -0,0 +1,130 @@
...
+    #
+    # @param base_url [String,URI] the base URL
+    # @param rel_url [String,URI] the relative URL, it should use the "relurl://"
+    #  schema otherwise the base URL is ignored
+    def initialize(base_url, rel_url)
+      @base = import_url(base_url)
+      @relative = import_url(rel_url)
+    end
+
+    # Build and absolute URL
```
jreidinger, line 60: typo build *an*

library/general/src/lib/yast2/rel_url.rb
```diff
@@ -0,0 +1,130 @@
...
+      ret
+    end
+
+  private
+
+    # a helper method for importing an URL,
+    # it creates a copy of the input object and pre-processes the
+    # "file://" and "relurl://" URLs
+    def import_url(url)
+      ret = URI(url).dup
```
jreidinger, line 106: how it handles exception from URI?

library/general/src/lib/yast2/rel_url.rb
```diff
@@ -0,0 +1,130 @@
...
+        # URI requires absolute path
+        ret.path = File.join("/", ret.host, ret.path)
+        ret.host = nil
+      end
+
+      ret
+    end
+
+    # adjust the result if it is a "file://" URL
+    def export_url(url)
```
jreidinger, line 119: so export_url returns void? it is kind of unexpect for me. I expect URI object.

library/general/src/lib/yast2/rel_url.rb
```diff
@@ -0,0 +1,130 @@
...
+
+require "yast"
+require "uri"
+
+Yast.import "InstURL"
+
+module Yast2
+  # Class for working with relative URLs ("relurl://")
+  class RelURL
+    attr_reader :base, :relative
```
jreidinger, line 28: please document type of attributes

library/general/src/lib/yast2/rel_url.rb
```diff
@@ -0,0 +1,130 @@
...
+      export_url(ret)
+      ret
+    end
+
+  private
+
+    # a helper method for importing an URL,
+    # it creates a copy of the input object and pre-processes the
+    # "file://" and "relurl://" URLs
+    def import_url(url)
```
jreidinger, line 105: name is kind of strange, as it for me do more like read or parse and not import.

library/general/src/lib/yast2/rel_url.rb
lslezak: The result will be this non-relative URL. The caller is responsible to use a relative URL to make it work properly.

library/general/src/lib/yast2/rel_url.rb
lslezak: No handling, they will just pass through. https://github.com/yast/yast-yast2/pull/1215/files#diff-89f24a21ba7c6b66736e8891b6fc08a235e2ef35a6d11e39fc5ee3c076b1ab38R47-R55

jreidinger (approved):

joseivanlopez:
https://ci.suse.de/job/yast-yast-yast2-master/174/

PR #1216, opened 2021-12-14
Do not check Report internals in the commandline test

imobachgs:
The command line does some assertions on the WFM and Report internals. So when they change, the test breaks. Asserting on a private method is not the ideal solution, but it moves the test one level up and it does not break so easily.


teclator (approved):
LGTM

PR #1218, opened 2021-12-15
Fixed RelURL unit test randomly crashing

lslezak:
- Fixes a [Jenkins crash](https://ci.suse.de/job/yast-yast-yast2-master/173/console) ("component cannot import namespace 'InstURL'").
- The problem was that the `Yast::InstURL` module is part of the `yast2-packager` package, but we cannot build depend on it to avoid circular dependecies.
- That module is mocked at some else unit tests so it depended on the order of executing the tests, if the RelURL test was executed too early the `Yast::InstURL` module was not mocked yet and it crashed.
- Fixed by moving the mock to the top level test helper.

### DIFF
--- a/library/general/test/yast2/rel_url_test.rb
+++ b/library/general/test/yast2/rel_url_test.rb
@@ -1,6 +1,4 @@
-# this mocks the Yast::InstURL module
-require_relative "../../../packages/test/test_helper.rb"
-
+require_relative "../test_helper"
 require "yast2/rel_url"
 
 describe Yast2::RelURL do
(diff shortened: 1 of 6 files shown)


library/general/test/yast2/rel_url_test.rb
```diff
@@ -1,4 +1,6 @@
-require_relative "../test_helper"
+# this mocks the Yast::InstURL module
+require_relative "../../../packages/test/test_helper.rb"
```
jreidinger, line 2: this looks strange. I expect that mock is in top level test_helper, so you do not need to load mock from another sub module

jreidinger (approved):

PR #1219, opened 2021-12-15
Unify `Package`, `PackageSystem` and `PackageAI`

imobachgs:
This PR is part of the plan to unify `Package.rb`, `PackageSystem.rb`, and `PackageIA.rb` and, by the way, get rid of `PackagesCommonInclude`. See https://trello.com/c/GnnQQcDa for further details.

Once we merge this code, we should avoid calling `PackageSystem` methods and use `Package` instead. This module should know whether to forward the call to `PackageSystem` or `PackageAI`. 

The following changes are included:

* `CheckAndInstallPackages*` are now implemented in the `Package` module. For backward compatibility reasons, the `PackageSystem` implements those methods too (although it forwards those calls to the `Package` module).
* Add support for packages to uninstall (taboos) to the `PackagesProposal`, so `PackageAI` does not keep its list of packages anymore.
* Drop the `PackagesCommonInclude` module. Its methods are now implemented directly in `Package`.
* Improved test-unit coverage.

Bug report: [bsc#1194886](https://bugzilla.suse.com/show_bug.cgi?id=1194886).

### DIFF
--- a/library/general/src/modules/PackagesProposal.rb
+++ b/library/general/src/modules/PackagesProposal.rb
@@ -126,7 +126,7 @@ module Yast
 
       log.info("Adding #{log_label(optional)} #{resolvables} of type #{type} for #{unique_id}")
 
-      current_resolvables = data_for(unique_id, type, optional: optional)
+      current_resolvables = data_for(unique_id, type, data(optional))
       current_resolvables.concat(resolvables)
 
       true
@@ -153,7 +153,7 @@ module Yast
 
       log.info("Setting #{log_label(optional)} #{resolvables} of type #{type} for #{unique_id}")
 
-      current_resolvables = data_for(unique_id, type, optional: optional)
+      current_resolvables = data_for(unique_id, type, data(optional))
       current_resolvables.replace(resolvables)
 
       true
@@ -185,7 +185,7 @@ module Yast
 
       log.info("Removing #{log_label(optional)} #{resolvables} type #{type} for #{unique_id}")
 
-      current_resolvables = data_for(unique_id, type, optional: optional)
+      current_resolvables = data_for(unique_id, type, data(optional))
       current_resolvables.reject! { |r| resolvables.include?(r) }
 
       log.info("#{log_label(optional)} left: #{current_resolvables.inspect}")
@@ -195,8 +195,9 @@ module Yast
 
     # Adds a list of resolvables to not be installed
     #
-    # @param [String] unique_id the unique identificator
-    # @param [Array<String>] resolvables list of resolvables to taboo
+    # @param unique_id [String] the unique identificator
+    # @param type [Symbol] resolvable type
+    # @param resolvables [Array<String>] list of resolvables to taboo
     # @return [Boolean] whether successful
     def AddTaboos(unique_id, type, resolvables)
       return false if !CheckParams(unique_id, type)
@@ -208,55 +209,53 @@ module Yast
       # ... and from the optional list too
       RemoveResolvables(unique_id, type, resolvables, optional: true)
 
-      @taboos[unique_id] ||= {}
-      @taboos[unique_id][type] ||= []
-      @taboos[unique_id][type].concat(resolvables)
+      data_for(unique_id, type, @taboos).concat(resolvables)
+      log.info("Adding taboos #{resolvables} of type #{type} for #{unique_id}")
 
       true
     end
 
     # Sets the taboos list for a given unique_id
     #
-    # @param [String] unique_id the unique identificator
-    # @param [Array<String>] resolvables list of resolvables to taboo
+    # @param unique_id [String] the unique identificator
+    # @param type [Symbol] resolvable type
+    # @param resolvables [Array<String>] list of resolvables to taboo
     # @return [Boolean] whether successful
     def SetTaboos(unique_id, type, resolvables)
-      return false if !CheckParams(unique_id, :package)
+      return false if !CheckParams(unique_id, type)
 
-      @taboos[unique_id] ||= {}
-      @taboos[unique_id][type] = resolvables
+      current_taboos = data_for(unique_id, type, @taboos)
+      current_taboos.replace(resolvables)
 
       true
     end
 
     # Removes a resolvables from the list of taboos
     #
-    # @param [String] unique_id the unique identificator
-    # @param [Array<String>] resolvables list of resolvables do not taboo
+    # @param unique_id [String] the unique identificator
+    # @param type [Symbol] resolvables type
+    # @param resolvables [Array<String>] list of resolvables to taboo
     # @return [Boolean] whether successful
     def RemoveTaboos(unique_id, type, resolvables)
-      return false if !CheckParams(unique_id, :package)
+      return false if !CheckParams(unique_id, type)
 
-      log.info "Remove taboo #{resolvables} for #{unique_id}"
+      current_taboos = data_for(unique_id, type, @taboos)
+      current_taboos.reject! { |r| resolvables.include?(r) }
 
-      return unless @taboos.key?(unique_id) && @taboos[unique_id].key?(type)
-
-      taboo_resolvables = @taboos[unique_id][type]
-      @taboos[unique_id][type] = taboo_resolvables - resolvables
-      @taboos[unique_id].delete(type) if @taboos[unique_id][type].empty?
-      @taboos.delete(unique_id) if @taboos[unique_id].empty?
+      log.info("Removing taboos  #{resolvables} type #{type} for #{unique_id}")
 
       true
     end
 
     # Returns the list of taboos for a ID
     #
-    # @param [String] unique_id the unique identificator
+    # @param unique_id [String] the unique identificator
+    # @param type [Symbol] resolvable type
     # @return [Array<String>] List of taboos for the given ID
     def GetTaboos(unique_id, type)
       return [] unless @taboos.key?(unique_id) && @taboos[unique_id].key?(type)
 
-      @taboos[unique_id][type]
+      data_for(unique_id, type, @taboos)
     end
 
     # Returns the full list of taboos
@@ -273,7 +272,7 @@ module Yast
     # Returns all resolvables selected for installation.
     #
     # @param [String] unique_id the unique identificator
-    # @param [Symbol] type resolvable type
+    # @param [Symbol] type resolvables type
     # @param [Boolean] optional True for optional list, false (the default) for
     #   the required list
 
@@ -289,7 +288,7 @@ module Yast
 
     # Returns list of selected resolvables of a given type
     #
-    # @param [Symbol] type resolvable type
+    # @param [Symbol] type resolvables type
     # @param [Boolean] optional True for optional list, false (the default) for
     #   the required list
     # @return [Array<String>] list of resolvables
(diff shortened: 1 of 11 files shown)


library/general/src/modules/PackagesProposal.rb
```diff
@@ -190,6 +193,71 @@ def RemoveResolvables(unique_id, type, resolvables, optional: false)
...
+    # @param [Array<String>] resolvables list of resolvables to taboo
+    # @return [Boolean] whether successful
+    def AddTaboos(unique_id, resolvables)
+      return false if !CheckParams(unique_id, :package)
+
+      log.info("Taboo #{resolvables} for #{unique_id}}")
+
+      # Remove the resolvable from the list of resolvables to install
+      RemoveResolvables(unique_id, :package, resolvables)
+      RemoveResolvables(unique_id, :package, resolvables, optional: true)
```
jreidinger, line 208: I would like to see comment why it needs to be called twice with different params.

library/general/src/modules/PackagesProposal.rb
```diff
@@ -190,6 +193,71 @@ def RemoveResolvables(unique_id, type, resolvables, optional: false)
...
 
+    # Adds a list of resolvables to do not be installed
+    #
+    # @param [String] unique_id the unique identificator
+    # @param [Array<String>] resolvables list of resolvables to taboo
+    # @return [Boolean] whether successful
+    def AddTaboos(unique_id, resolvables)
+      return false if !CheckParams(unique_id, :package)
+
+      log.info("Taboo #{resolvables} for #{unique_id}}")
```
jreidinger, line 204: can you use `resolvables.inspect` for easier reading in logs?

library/general/src/modules/PackagesProposal.rb
```diff
@@ -190,6 +193,71 @@ def RemoveResolvables(unique_id, type, resolvables, optional: false)
       true
     end
 
+    # Adds a list of resolvables to do not be installed
+    #
+    # @param [String] unique_id the unique identificator
+    # @param [Array<String>] resolvables list of resolvables to taboo
+    # @return [Boolean] whether successful
+    def AddTaboos(unique_id, resolvables)
+      return false if !CheckParams(unique_id, :package)
```
jreidinger, line 202: To be honest, I do not get logic here. it should always succeed for :package...and it just check that unique id is not nil or empty. Maybe second parameter no longer makes sense?

library/general/src/modules/PackagesProposal.rb
```diff
@@ -190,6 +193,71 @@ def RemoveResolvables(unique_id, type, resolvables, optional: false)
...
+    #
+    # @param [String] unique_id the unique identificator
+    # @param [Array<String>] resolvables list of resolvables do not taboo
+    # @return [Boolean] whether successful
+    def RemoveTaboos(unique_id, resolvables)
+      return false if !CheckParams(unique_id, :package)
+
+      log.info "Remove taboo #{resolvables} for #{unique_id}"
+
+      taboo_resolvables = @taboos[unique_id] || []
```
jreidinger, line 239: Well, I would at least log if there are no taboos for given unique id.

library/general/src/modules/PackagesProposal.rb
```diff
@@ -190,6 +193,71 @@ def RemoveResolvables(unique_id, type, resolvables, optional: false)
...
+
+      true
+    end
+
+    # Returns the list of taboos for a ID
+    #
+    # @param [String] unique_id the unique identificator
+    # @return [Array<String>] List of taboos for the given ID
+    def GetTaboos(unique_id)
+      @taboos[unique_id] || []
```
jreidinger, line 251: same here. I would at least log that taboos is not defined for given id.

library/packages/src/modules/Package.rb
```diff
@@ -31,186 +31,356 @@
...
       packages = deep_copy(packages)
-      function = Convert.convert(
-        Ops.get(functions, "DoRemove"),
-        from: "any",
-        to:   "boolean (list <string>)"
-      )
-      function.call(packages)
+      PackageDialog(packages, true, message)
     end
+    # FIXME
```
jreidinger, line 341: FIXME what? I am surprised that rubocop does not complain

library/packages/src/modules/Package.rb
```diff
@@ -31,186 +31,356 @@
...
+
+    def Install(package)
+      InstallMsg(package, nil)
+    end
+
+    def InstallAll(packages)
+      packages = deep_copy(packages)
+      InstallAllMsg(packages, nil)
+    end
+    # FIXME
```
jreidinger, line 368: Same here. FIXME what?

library/packages/src/modules/PackageAI.rb
```diff
@@ -29,123 +29,101 @@
...
+    #
+    # @param toinstall [Array<String>] Name of the packages to install
+    # @param toremove [Array<String>] Name of the packages to remove
+    # @return [Boolean] true on success
+    #
+    # @see Yast::PackageClass#DoInstallAndRemove
+    def DoInstallAndRemove(toinst, torem)
+      if !toinst.empty?
+        Yast::PackagesProposal.AddResolvables("autoyast", :package, toinst)
+        Yast::PackagesProposal.RemoveTaboos("autoyast", toinst) # FIXME: should be done by PackagesProposal
```
jreidinger, line 54: Well, tricky question here. What will happen if different unique id taboo package?

imobachgs: Then it is tabooed.

library/general/src/modules/PackagesProposal.rb
```diff
@@ -193,23 +193,24 @@ def RemoveResolvables(unique_id, type, resolvables, optional: false)
...
 
-    # Adds a list of resolvables to do not be installed
+    # Adds a list of resolvables to not be installed
     #
     # @param [String] unique_id the unique identificator
     # @param [Array<String>] resolvables list of resolvables to taboo
     # @return [Boolean] whether successful
-    def AddTaboos(unique_id, resolvables)
-      return false if !CheckParams(unique_id, :package)
+    def AddTaboos(unique_id, type, resolvables)
```
jreidinger, line 202: please document new param

library/general/src/modules/PackagesProposal.rb
```diff
@@ -219,10 +220,11 @@ def AddTaboos(unique_id, resolvables)
     # @param [String] unique_id the unique identificator
     # @param [Array<String>] resolvables list of resolvables to taboo
     # @return [Boolean] whether successful
-    def SetTaboos(unique_id, resolvables)
+    def SetTaboos(unique_id, type, resolvables)
```
jreidinger, line 224: same here for missing type documentation. Also missing verify type in CheckParams....and last but not least. Is valid to pass nil resolvables to remove taboos?

library/general/src/modules/PackagesProposal.rb
```diff
@@ -232,13 +234,16 @@ def SetTaboos(unique_id, resolvables)
     # @param [String] unique_id the unique identificator
     # @param [Array<String>] resolvables list of resolvables do not taboo
     # @return [Boolean] whether successful
-    def RemoveTaboos(unique_id, resolvables)
+    def RemoveTaboos(unique_id, type, resolvables)
       return false if !CheckParams(unique_id, :package)
```
jreidinger: also here it is not checked in CheckParams

library/general/src/modules/PackagesProposal.rb
```diff
@@ -232,13 +234,16 @@ def SetTaboos(unique_id, resolvables)
...
 
       log.info "Remove taboo #{resolvables} for #{unique_id}"
 
-      taboo_resolvables = @taboos[unique_id] || []
-      @taboos[unique_id] = taboo_resolvables - resolvables
+      return unless @taboos.key?(unique_id) && @taboos[unique_id].key?(type)
```
jreidinger: easier can be to just call `return unless @taboos.dig(unique_id, type)`

jreidinger: or maybe even better 
```ruby
taboo_resolvables = @taboos.dig(unique_id, type)
return unless taboo_resolvables
```

library/general/src/modules/PackagesProposal.rb
```diff
@@ -248,15 +253,21 @@ def RemoveTaboos(unique_id, resolvables)
     #
     # @param [String] unique_id the unique identificator
     # @return [Array<String>] List of taboos for the given ID
-    def GetTaboos(unique_id)
-      @taboos[unique_id] || []
+    def GetTaboos(unique_id, type)
+      return [] unless @taboos.key?(unique_id) && @taboos[unique_id].key?(type)
```
jreidinger, line 256: same here with dig

library/general/src/modules/PackagesProposal.rb
```diff
@@ -248,15 +253,21 @@ def RemoveTaboos(unique_id, resolvables)
...
     end
 
     # Returns the full list of taboos
     #
     # @return [Array<String>] List of taboos
-    def GetAllTaboos
-      @taboos.values.flatten.uniq
+    def GetAllTaboos(type)
+      all_taboos = @taboos.values.each_with_object([]) do |tab, all|
+        all.concat(tab[type]) if tab.key?(type)
```
jreidinger, line 266: can this potentially also add nil? if so, then compact is missing.

library/general/src/modules/PackagesProposal.rb
```diff
@@ -248,15 +253,21 @@ def RemoveTaboos(unique_id, resolvables)
...
+
+      @taboos[unique_id][type]
     end
 
     # Returns the full list of taboos
     #
     # @return [Array<String>] List of taboos
-    def GetAllTaboos
-      @taboos.values.flatten.uniq
+    def GetAllTaboos(type)
```
jreidinger, line 264: missing type documentation

library/general/src/modules/PackagesProposal.rb
```diff
@@ -397,25 +396,26 @@ def log_label(optional)
...
-    # (required/optinal). If the list does not exit yet then a new empty list is created.
+    # (required/optional). If the list does not exit yet then a new empty list is created.
     #
     # @param [String] unique_id
     # @param [Symbol] type
+    # @param [Hash] Resolvables lists
     # @param [Boolean] optional True for optional list, false (the default) for
     #   the required list
     # @return [Array<String>] the stored resolvables list
-    def data_for(unique_id, type, optional: false)
```
jreidinger, line 325: removed optional param is still documented.

library/general/src/modules/PackagesProposal.rb
```diff
@@ -185,7 +185,7 @@ def RemoveResolvables(unique_id, type, resolvables, optional: false)
 
       log.info("Removing #{log_label(optional)} #{resolvables} type #{type} for #{unique_id}")
 
-      current_resolvables = data_for(unique_id, type, optional: optional)
+      current_resolvables = data_for(unique_id, type, data(optional))
```
jreidinger, line 188: I start feeling that `data` is no longer good name here.

library/general/src/modules/PackagesProposal.rb
imobachgs: I do not get how it could add nil. We are checking `tab.key?(type)` and, in that case, we concatenate `tab[type]` which is expected to be an array.

library/general/src/modules/PackagesProposal.rb
imobachgs: No, you should pass an empty array to remove all of them.

jreidinger (approved):

PR #1220, opened 2021-12-16
Fixed RelURL to work properly with the FTP URLs

lslezak:
## FTP URL Fix

It turned out that the `URI` class processes FTP URL paths a bit differently. Normally you have to use an absolute path otherwise you get an exception:

```
irb(main):001:0> u = URI("http://example.com")
=> #<URI::HTTP http://example.com>
irb(main):002:0> u.path = "path"
Traceback (most recent call last):
        4: from /usr/bin/irb:11:in `<main>'
        3: from (irb):2
        2: from /usr/lib64/ruby/2.5.0/uri/generic.rb:819:in `path='
        1: from /usr/lib64/ruby/2.5.0/uri/generic.rb:771:in `check_path'
URI::InvalidComponentError (bad component(expected absolute path component): path)
irb(main):003:0> u.path = "/path"
=> "/path"
```

But FTP URL is escaped when the path starts with `/`:

```
irb(main):001:0> u = URI("ftp://example.com")
=> #<URI::FTP ftp://example.com/>
irb(main):002:0> u.path = "/path"
=> "/path"
irb(main):003:0> u.to_s
=> "ftp://example.com/%2Fpath"
```

And it is allowed to use a relative path:

```
irb(main):004:0> u.path = "path"
=> "path"
irb(main):005:0> u.to_s
=> "ftp://example.com/path"
```

So there is an exception in the code for the FTP URLs to process them correctly.

*Note: that's because FTP really allows using absolute paths for files!*

## Improve Mocking Yast::InstURL Module

Originally the module was always mocked but then the problem is that the RSpec verifying doubles will not work as expected. There might be an inconsistency between the mocked `InstURL` and the real `InstURL`.

But when running the tests locally or in GitHub Actions the real module is actually present and we do no need to reimplement it completely. Then the verifying doubles would use the real module and provide more reliable testing.

### DIFF
--- a/library/general/test/yast2/rel_url_test.rb
+++ b/library/general/test/yast2/rel_url_test.rb
@@ -199,6 +199,11 @@ describe Yast2::RelURL do
       expect(relurl.absolute_url.to_s).to eq("file://foo/bar/test")
     end
 
+    it "works with file:/// base URL" do
+      relurl = Yast2::RelURL.new("file:///", "relurl://test")
+      expect(relurl.absolute_url.to_s).to eq("file://test")
+    end
+
     it "goes up with file:// base URL properly" do
       relurl = Yast2::RelURL.new("file://foo/bar", "relurl://../../test")
       expect(relurl.absolute_url.to_s).to eq("file://test")
--- a/library/general/test/yast2/rel_url_test.rb
+++ b/library/general/test/yast2/rel_url_test.rb
@@ -174,6 +174,26 @@ describe Yast2::RelURL do
       expect(relurl.absolute_url.to_s).to eq("http://user:password@example.com/test")
     end
 
+    it "works with ftp:// URL" do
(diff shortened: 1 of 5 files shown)


library/general/test/yast2/rel_url_test.rb
```diff
@@ -174,6 +174,26 @@
...
+    it "works with cifs:// URL" do
+      relurl = Yast2::RelURL.new("cifs://server/share/path", "relurl://test")
+      expect(relurl.absolute_url.to_s).to eq("cifs://server/share/path/test")
+    end
+
+    it "works with nfs:// URL" do
+      relurl = Yast2::RelURL.new("nfs://server/export/path", "relurl://test")
+      expect(relurl.absolute_url.to_s).to eq("nfs://server/export/path/test")
+    end
+
```
jreidinger, line 196: please add test also for file with triple slash. Which can also need special handling.

jreidinger (approved):

PR #1221, opened 2021-12-20
bsc#1193784: do not reinitialize repos during upgrade

imobachgs:
Reinitializing the repositories during upgrade to `/` makes the process fail. The problem is that, in that case, we need to use repositories of the system we are updating too (`/mnt`) and, after that call, they are gone.

See [bsc#1193784](https://bugzilla.suse.com/show_bug.cgi?id=1193784) and [bsc#1192437](https://bugzilla.suse.com/show_bug.cgi?id=1192437).

### DIFF
--- a/library/control/src/modules/WorkflowManager.rb
+++ b/library/control/src/modules/WorkflowManager.rb
@@ -60,6 +60,7 @@ module Yast
       Yast.import "String"
       Yast.import "XML"
       Yast.import "Report"
+      Yast.import "Mode"
 
       #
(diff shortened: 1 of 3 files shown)


imobachgs:
Reinitializing the repositories during upgrade to `/` makes the process fail. The problem is that, in that case, we need to repositories of the system we are updating too (`/mnt`).

See [bsc#1193784](https://bugzilla.suse.com/show_bug.cgi?id=1193784) and [bsc#1192437](https://bugzilla.suse.com/show_bug.cgi?id=1192437).

PR #1225, opened 2021-12-23
Ruby3 2

jreidinger:
more adaptations was needed.

PR #1228, opened 2022-01-07
Fixed test failure in Ruby 2.5

lslezak:
- The fix for Ruby 3.0 (#1225) caused a regression in Ruby 2.5
- Also added missing `Stage.normal` mock, for me the tests failed when running locally

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan  7 14:41:15 UTC 2022 - Ladislav Slezák <lslezak@suse.cz>
+
+- Fixed test failure in Ruby 2.5, caused by the fix for Ruby 3.0
+  (related to bsc#1193192)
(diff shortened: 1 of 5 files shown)


package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Jan  7 14:41:15 UTC 2022 - Ladislav Slezák <lslezak@suse.cz>
+
+- Fixed test failure in Ruby 2.5, caused by the fix for Ruby 3.0
+  (related to bsc#1193192)
+- 4.4.34
```
mvidner, line 6: spec file bump missing

mvidner:
OK, for the record I think it is a bug in Ruby 2.5 but it is not worth identifying.

jreidinger (approved):

teclator (approved):
LGTM

PR #1229, opened 2022-01-07
Remove useless "super"

lslezak:
- Fix up for #1228
- The super is actually not needed as the class does not inherit from any base class (besides implicit `Object`)
- Tested with `YAST_SUBMIT=sle_latest rake osc:build`, builds fine


teclator (approved):
LGTM

PR #1230, opened 2022-01-07
New rubocop

jreidinger:


library/control/src/modules/ProductControl.rb
```diff
@@ -653,15 +619,11 @@ def RunRequired(stage, mode)
...
-          # proposals
-        elsif !Ops.get_string(one_module, "proposal").nil? &&
-            Ops.get_string(one_module, "proposal", "") != ""
-          next true
-        end
+        name = one_module["name"]
+        proposal = one_module["proposal"]
+
+        next true if name && !name.empty?
+        next true if proposal && !proposal.empty?
```
dgdavid, line 626: Much better now :+1:

library/cwm/src/modules/TablePopup.rb
```diff
@@ -799,6 +791,7 @@ def TableHandle(descr, key, event_descr)
           else
             -1
           end
+          # rubocop:enable Lint/DuplicateBranch
```
dgdavid, line 794: Do you think is it a good opportunity for improving it a bit and stop using those legacies `Ops`?

jreidinger: Well, I think TablePopup from CWM is almost not used, so probably not worth it. Also my goal is to not introduce any bugs. So if we improve it, we need also test coverage, which is currently zero.

library/cwm/examples/date_time.rb
```diff
@@ -19,6 +21,8 @@ def label
 
 class EventDate < CWM::DateField
   def initialize
+    super
+
```
dgdavid, line 25: NP: Just curious about why in yast2-bootloader you kept `textdomain` before `super` and here is the other way around.

library/network/src/modules/PortRanges.rb
```diff
@@ -586,7 +586,7 @@ def FlattenServices(old_list, protocol)
               new_max = this_max_pr
               # the second one partly covers the first one (by its left side)
             elsif Ops.greater_or_equal(min_pr, this_min_pr) &&
-                Ops.less_or_equal(max_pr, this_max_pr)
+                Ops.greater_or_equal(max_pr, this_max_pr)
```
dgdavid, line 589: Nice catch.

library/cwm/examples/date_time.rb
jreidinger: well, it does not matter. It is order independent as textdomain is added to domain pool and not set exclusively for given object (it will be hard to do it in ruby).

dgdavid (approved):
LGTM

library/network/src/modules/PortRanges.rb
jreidinger: yeap, rubocop complain that condition is identical and then I spend 5 minutes thinking how it should be correctly.

library/cwm/src/modules/TablePopup.rb
dgdavid: Let's keep it as it is. Thanks.

PR #1232, opened 2022-01-10
[PoC] Run unit tests against both Ruby 2.5 and Ruby 3.0

lslezak:
- We currently run the GitHub Actions using the TumbleWeed based docker image
- That means it runs the unit tests against Ruby 3.0 (or Ruby 3.1 soon....)
- The problem is that we should run the unit tests also against Ruby 2.5 included in SLE15-SP4/Leap15.4 to make sure it also works fine in older Ruby
- This also fixes the problem with (old) Rubocop and Yard which do not work properly in Ruby 3.x, just run them in the Ruby 2.5 image
- We do not need to run *all* checks using both Docker images, e.g. the Perl syntax check or the POT check are enough to run just in one Docker image, these should not depend on the Ruby version included. (Um, which one should we prefer in this case? The TW image or the 15.4 image? :thinking: )

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,13 @@ on: [push, pull_request]
 jobs:
   Tests:
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        docker_repo: [ "containers", "containers_15.4" ]
+
     container:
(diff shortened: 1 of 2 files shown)


lslezak:
Closing, the change was applied globally by a script.

PR #1233, opened 2022-01-14
Adapt Report.yesno_popup to Ruby 3 (bsc#1193192)

ancorgs:
A hash was being passed to a method expecting the hash would be interpreted as a set of keyword params. That works in Ruby 2 as-is, but in Ruby 3 the `**` operator must be used.
 
The problem was detected by me when trying to open two instances of the Partitioner and also by the internal SUSE openQA instance during this test: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=Y.52.5&groupid=125


dgdavid (approved):
LGTM

jreidinger (approved):

PR #1235, opened 2022-01-21
Fix CWM dialog

joseivanlopez:
## Problem

Method `Dialog#run` delegates arguments to `Dialog.new`. Argument delegation is handled differently in ruby 2.6 and before, see [explanation here](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/):

In ruby 2.7 or later, an empty Hash argument is automatically converted and absorbed into **kwargs, and the delegation call removes the empty keyword hash, so no argument is passed to target:

~~~
def target(*args, **kwargs)
end

def foo(*args, **kwargs, &block)
  target(*args, **kwargs, &block)
end

foo(1)       #=> Ruby 2.7: calls target(1)
foo(1)       #=> Ruby 2.6: calls target(1, {})
~~~

https://bugzilla.suse.com/show_bug.cgi?id=1194984

## Solution

Do not pass empty `**kwargs` to the target method.

## Testing

* Added unit test, and checked in different ruby versions.

### DIFF
--- a/library/cwm/src/lib/cwm/dialog.rb
+++ b/library/cwm/src/lib/cwm/dialog.rb
@@ -31,7 +31,18 @@ module CWM
 
     # A shortcut for `.new(*args).run`
     def self.run(*args, **kws)
-      new(*args, **kws).run
+      # Argument delegation is handled differently since ruby 2.7:
+      #
+      # An empty Hash argument is automatically converted and absorbed into **kws, and the delegation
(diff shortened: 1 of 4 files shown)


jreidinger (approved):

dgdavid (approved):
LGTM

PR #1237, opened 2022-01-25
Use `Package` instead of `PackageSystem`

imobachgs:
Adapt to `Package` and `PackageSystem` changes. See https://github.com/yast/yast-yast2/pull/1219 for further information.


dgdavid (approved):

library/network/src/lib/network/susefirewall.rb
```diff
@@ -554,10 +554,10 @@ def SuSEFirewallIsInstalled
       return true if @needed_packages_installed
 
       if Mode.normal
-        @needed_packages_installed = PackageSystem.CheckAndInstallPackages([@FIREWALL_PACKAGE])
+        @needed_packages_installed = Package.CheckAndInstallPackages([@FIREWALL_PACKAGE])
         log.info "CheckAndInstallPackages -> #{@needed_packages_installed}"
       else
-        @needed_packages_installed = PackageSystem.Installed(@FIREWALL_PACKAGE)
+        @needed_packages_installed = Package.Installed(@FIREWALL_PACKAGE)
```
jreidinger, line 560: good example that we should finally clean susefirewall, as it is dead technology and we use only firewalld.

library/network/src/modules/DnsServerAPI.pm
```diff
@@ -657,15 +657,15 @@ sub UnquoteString {
 sub Init {
     if (Mode->test ()) {
       $package_installed = 1;
-      y2milestone("TestMode -> PackageSystem->Installed: ", $package_installed);
+      y2milestone("TestMode -> Package->Installed: ", $package_installed);
```
jreidinger, line 660: dead code. old testsuite is dead and so mode->test

library/packages/src/include/packages/common.rb
```diff
@@ -150,6 +150,7 @@ def RemoveAllMsg(packages, message)
     end
 
     def Install(package)
+      log.info "DEBUG: call Install(#{package.inspect})"
```
jreidinger, line 153: this looks like leftover. Either use really log.debug or remove that DEBUG prefix.

library/network/src/lib/network/susefirewall.rb
imobachgs: You are right, but I do not plan to do it in this PR.

jreidinger (approved):

PR #1239, opened 2022-01-26
Added Y2Packager::NewRepositorySetup class (bsc#1194453)

lslezak:
- Related to https://bugzilla.suse.com/show_bug.cgi?id=1194453
- Added Y2Packager::NewRepositorySetup class to track new repositories
- Similar to the already existing [Y2Packager::OriginalRepositorySetup](https://github.com/yast/yast-yast2/blob/4f0560a55f0b4ece41ed50063fd7baf00cd8eedc/library/packages/src/lib/y2packager/original_repository_setup.rb#L28) class 
- This class will used from other packages (y2-registration and y2-installation)
- 4.4.42

## Technical Details

During upgrade YaST removes the old repository services (e.g. SLE12) from the system. Unfortunately when upgrading via RMT it returns the same name for the new service (SLE15), YaST did not expected that and removed the "old" service from the system. But this accidentally removed the new service and that made troubles during upgrade.

## Testing

- Added unit tests
- Tested manually against the RMT server reported in the bug

### DIFF
--- a/library/packages/src/lib/y2packager/new_repository_setup.rb
+++ b/library/packages/src/lib/y2packager/new_repository_setup.rb
@@ -1,0 +1,52 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2022 SUSE LINUX GmbH, Nuremberg, Germany.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+require "yast"
+
+require "singleton"
+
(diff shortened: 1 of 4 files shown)


library/packages/src/lib/y2packager/new_repository_setup.rb
```diff
@@ -0,0 +1,52 @@
...
+
+require "yast"
+
+require "singleton"
+
+module Y2Packager
+  #
+  # This class stores the new repositories and services added during
+  # installation or upgrade. It can be used together with the
+  # OriginalRepositorySetup class to find the old and new repositories.
```
jreidinger, line 21: if that comparison will be made on multiple places, then maybe it makes sense to have class that find that diffs also in yast2?

jreidinger (approved):

library/packages/src/lib/y2packager/new_repository_setup.rb
lslezak: It is just a trivial "exclude", see https://github.com/yast/yast-installation/pull/1015/files#diff-d90a84d467a4dedd884dbed8f308018c56b095a7423a00109b558129a3ec8703R70

PR #1240, opened 2022-01-28
Bootloader timeout

jreidinger:
feature: support produce/role specific default timeout when booting.


dgdavid (approved):
I only miss a link to https://jira.suse.com/browse/SLE-22667 in the PR description :) But it LGTM :+1:

PR #1241, opened 2022-02-16
do not strip trailing white space in XML elements (bsc#1195910)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1195910
- https://trello.com/c/1uUwxpq6

Surrounding white space in some elements may be desired. Like in scripts or other file content. Be more careful when removing it.

## The catch

The AutoYaST profile is processed by YaST to construct the final profile. It's written, modified by scripts, then read again, for example. Ensure that during this processing the CDATA 'attribute' to elements is not lost.

## Solution

Strip leading white space when reading XML CDATA blocks. And create CDATA blocks when generating XML if an element contains trailing white space.

## Note

Why not also keeping leading white space? - It's common to have a leading newline in `script` sections. They must be removed.

### DIFF
--- a/library/xml/src/modules/XML.rb
+++ b/library/xml/src/modules/XML.rb
@@ -247,7 +247,11 @@ module Yast
       children = children.select(&:element?)
       # we need just direct text under node. Can be splitted with another elements
       # but remove whitespace only text
-      text = node.xpath("text()").text.strip
+      #
+      # Do NOT strip trailing white space in CDATA blocks. Maybe people put
+      # it intentionally there (bsc#1195910).
+      text_nodes = node.xpath("text()")
+      text = text_nodes.map { |n| n.cdata? ? n.content.lstrip : n.content.strip }.join
 
       type = fetch_type(text, children, node)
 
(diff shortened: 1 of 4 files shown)


library/xml/src/modules/XML.rb
```diff
@@ -247,7 +247,11 @@ def parse_node(node, result)
       children = children.select(&:element?)
       # we need just direct text under node. Can be splitted with another elements
       # but remove whitespace only text
-      text = node.xpath("text()").text.strip
+      #
+      # Do NOT strip trailing white space in CDATA blocks. Maybe people put
+      # it intentionally there (bsc#1195910).
+      text_nodes = node.xpath("text()")
+      text = text_nodes.map { |n| n.cdata? ? n.content.lstrip : n.content.strip }.join
```
jreidinger, line 254: is there reason for that lstrip in cdata? I expect it should be as it is

wfeldt: You want to strip leading newlines in scripts.

jreidinger (changes requested):
In general what I miss here is unit testing demonstrating issue.
And I am not sure if we should also check serializer, if to_xml parse_xml pair of operation does not strip newlines, which can be also issue.

PR #1243, opened 2022-02-16
do not strip trailing white space in XML elements (bsc#1195910)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1195910
- https://trello.com/c/1uUwxpq6

Surrounding white space in some elements may be desired. Like in scripts or other file content. Be more careful when removing it.

## The catch

The AutoYaST profile is processed by YaST to construct the final profile. It's written, modified by scripts, then read again, for example. Ensure that during this processing the CDATA 'attribute' to elements is not lost.

## Solution

Strip leading white space when reading XML CDATA blocks. And create CDATA blocks when generating XML if an element contains trailing white space.

## Note

Why not also keeping leading white space? - It's common to have a leading newline in `script` sections. They must be removed.
See https://doc.opensuse.org/projects/autoyast/#id-1.9.4.2.4.5 for an example.

## Note 2

On closer inspection, in SLE12 leading white space is preserved. And it causes no issues. Hm.

### DIFF
--- a/package/yast2.changes
+++ b/package/yast2.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Feb 16 15:41:54 UTC 2022 - Steffen Winterfeldt <snwint@suse.com>
+
+- do not strip trailing white space in XML elements (bsc#1195910)
+- 4.3.67
+
+-------------------------------------------------------------------
 Mon Dec 20 11:07:31 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Do not reinitialize the packaging system during offline
--- a/package/yast2.spec
+++ b/package/yast2.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2
-Version:        4.3.66
+Version:        4.3.67
 
 Release:        0
 Summary:        YaST2 Main Package


library/xml/src/modules/XML.rb
```diff
@@ -247,7 +247,13 @@ def parse_node(node, result)
...
       # we need just direct text under node. Can be splitted with another elements
       # but remove whitespace only text
-      text = node.xpath("text()").text.strip
+      #
+      # Do NOT strip trailing white space in CDATA blocks. Maybe people put
+      # it intentionally there (bsc#1195910).
+      #
+      # See also #add_element.
+      text_nodes = node.xpath("text()")
+      text = text_nodes.map { |n| n.cdata? ? n.content.lstrip : n.content.strip }.join
```
jreidinger, line 256: well, to be on safe side, I would do not stripping on cdata and below when writting checking spaces also at the beginning. This would make our to_xml/from_xml much more consistent as we never loose some characters that user can intentionally place.

wfeldt: As said before, this exaple will not work if we keep leading white space:
https://doc.opensuse.org/projects/autoyast/#id-1.9.4.2.4.5

library/xml/test/xml_test.rb
```diff
@@ -85,6 +85,30 @@
...
+          "  <lest>\nlest</lest>\n" \
+          "  <test> test</test>\n" \
+          "</test>\n"
+
+        expect(subject.YCPToXMLString("test", input)).to eq expected
+      end
+
+      it "creates CDATA element if string ends with spaces" do
+        input = { "test" => "test ", "lest" => "lest\n" }
+        expected = "<?xml version=\"1.0\"?>\n" \
```
lslezak, line 103: What about using heredoc to make it better readable?

```ruby
expected = <<XML
<?xml version=\"1.0\"?>
<!DOCTYPE test SYSTEM \"just_testing.dtd\">
<test>
  <lest><![CDATA[lest
]]></lest>
  <test><![CDATA[test ]]></test>
</test>
XML

```

(not tested)

shundhammer:
If it works, I am all for it.

jreidinger (approved):

PR #1246, opened 2022-03-02
New Doc: Invoking External Commands

shundhammer:
This is a more permanent place than the initial [Gist](https://gist.github.com/shundhammer/382faaae5371087e7afb13007510c6d5).

This documents how calling an external command in YaST works, how it should be used, and it covers security aspects.

This does not have a version bump since there is no code change, so it can wait to be submitted until there is a code change.

## Related

- https://github.com/yast/yast-iscsi-client/issues/111
- https://bugzilla.suse.com/show_bug.cgi?id=1196086
- https://trello.com/c/vySWWKUv/2727-3-security-research-iscsi-the-iscsi-commands-are-moving-from-sbin-to-usr-sbin
- https://github.com/yast/yast.github.io/issues/172

### DIFF
--- a/doc/yast-invoking-external-commands.md
+++ b/doc/yast-invoking-external-commands.md
@@ -1,6 +1,46 @@
 # Invoking External Commands in YaST
 
-## From the Ruby Code
+## Best Practice: Yast::Execute
+
+_This is the method to use since about 2018. Don't use SCR .target.bash in new code._
+
+```Ruby
+Yast::Execute.on_target!("ls", "-l", arg)
+```
+
+or
+
+```Ruby
+Yast::Execute.locally!("ls", "-l", arg)
+```
+
+This does **not** use a shell to invoke the command, it does a simple `fork()`
+/ `execvp()`. It does use `$PATH`, though. See below for security
+considerations.
+
+Since this does not use a shell, there is no wildcard file globbing, no I/O
+redirection, no pipelined commands, no `||` or `&&`. But all that should be
+handled in Ruby code anyway; don't use `| grep | awk` etc. pipelines in YaST
+code; Ruby can do all that better and safer.
+
+Under the hood, _Yast::Execute_ uses the [Cheetah Ruby Gem](https://github.com/openSUSE/cheetah).
+
+See also the [Yast::Execute reference documentation](https://www.rubydoc.info/github/yast/yast-yast2/master/Yast/Execute)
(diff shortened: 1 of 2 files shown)


shundhammer:
@msmeissn FYI

doc/yast-invoking-external-commands.md
```diff
@@ -0,0 +1,205 @@
...
+stderr = result["stderr"]
+```
+
+or
+
+```Ruby
+output = SCR.Execute(path(".target.bash_output"), command)["stdout"]
+```
+
+_(also available: .target.bash_background, .target.bash_input)_
```
mvidner, line 24: We're missing the current best practice, [Yast::Execute](https://github.com/yast/yast-yast2/blob/master/library/system/src/lib/yast2/execute.rb), based on [Cheetah](https://github.com/openSUSE/cheetah/), a standalone gem developed at SUSE by David Majda.

Its command-spawning part is [here](https://github.com/openSUSE/cheetah/blob/ac990d5afcff45153d2d9428059cabe01feb506f/lib/cheetah.rb#L557-L559)
It uses [Kernel#exec](https://ruby-doc.com/core/Kernel.html#method-i-exec)

... which probably uses PATH but I can't find it there :-/

mvidner (commented):
Uh, I have found, a... it's like a skeleton in the closet, but in reverse. A new kid on the block?

doc/yast-invoking-external-commands.md
shundhammer: AFAICS you can pass your custom environment as an option, otherwise it will keep the existing one. But in any case, afterwards it will restore the old one.

https://github.com/openSUSE/cheetah/blob/ac990d5afcff45153d2d9428059cabe01feb506f/lib/cheetah.rb#L557-L559

```Ruby
        with_env(options[:env]) do
          exec([command, command], *args)
        end
```

https://github.com/openSUSE/cheetah/blob/ac990d5afcff45153d2d9428059cabe01feb506f/lib/cheetah.rb#L437

```Ruby
    def with_env(env, &block)
      old_env = ENV.to_hash
      ENV.update(env)
      block.call
    ensure
      ENV.replace(old_env)
    end
```

shundhammer: https://www.rubydoc.info/github/openSUSE/cheetah/Cheetah.run

> :env (Hash) — default: {} — Allows to update ENV for the time of running the command. if key maps to nil value it is deleted from ENV.

mvidner (approved):
Thanks!

PR #1249, opened 2022-03-11
Force package backend

imobachgs:
In the recent past, we decided to merge `Package`, `PackageSystem` and `PackageAI`. The idea was to use `Package` as the entry point and rely on `PackageAI` and `PackageSystem`, depending on the situation. However, we found out later that things are not that easy. See [bsc#1196963](https://bugzilla.suse.com/show_bug.cgi?id=1196963).

There might a scenario where you want to force Package to work with the real packages. For instance, while reading the configuration during a `clone_system` operation: the mode is still `autoinst_config` but you are dealing with the underlying system. In those cases, you can force `Package` to work with `PackageSystem`. `Package` to work with `PackageSystem`.

As an escape hatch, we have decided to add an option to Package to override the decision about which module to use. This behavior is now documented in the Package module.

### DIFF
--- a/library/packages/src/modules/Package.rb
+++ b/library/packages/src/modules/Package.rb
@@ -53,7 +53,7 @@ module Yast
   # There might a scenario where you want to force Package to work with the real packages. For
   # instance, while reading the configuration during a `clone_system` operation: the mode is still
   # `autoinst_config` but you are dealing with the underlying system. In those cases, you can force
-  # {Package} to work with {PackageSystem}. {Package} to work with {PackageSystem}.
+  # {Package} to work with {PackageSystem}.
   #
   # @example Forcing to check for packages on the underlying system
   #   Yast::Package.Installed("firewalld", target: :system)
--- a/library/packages/src/modules/Package.rb
+++ b/library/packages/src/modules/Package.rb
@@ -43,10 +43,10 @@ module Yast
   # ## Prefer Package to PackageSystem
   #
   # Depending on the mode, this module decides if it should interact with PackageSystem (libzypp) or
-  # PackagesProposal. For instance, if you open a module in the AutoYaST UI, calling to
+  # PackageAI (AutoYaST). For instance, if you open a module in the AutoYaST UI, calling to
   # {CheckAndInstallPackages} does not install the package for real. Instead, it adds the package to
-  # the {PackagesProposal} so it gets exported in the profile. However, when running on other modes
-  # (normal, installation, etc.), it just installs the package.
+  # the list of packages to include in the profile. However, when running on other modes (normal,
+  # installation, etc.), it just installs the package.
   #
   # ## Overriding default behavior
   #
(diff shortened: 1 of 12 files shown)


library/packages/src/modules/Package.rb
```diff
@@ -38,31 +38,34 @@
...
+  # {CheckAndInstallPackages} does not install the package for real. Instead, it adds the package to
+  # the {PackagesProposal} so it gets exported in the profile. However, when running on other modes
+  # (normal, installation, etc.), it just installs the package.
+  #
+  # ## Overriding default behavior
+  #
+  # There might a scenario where you want to force Package to work with the real packages. For
+  # instance, while reading the configuration during a `clone_system` operation: the mode is still
+  # `autoinst_config` but you are dealing with the underlying system. In those cases, you can force
+  # {Package} to work with {PackageSystem}. {Package} to work with {PackageSystem}.
```
ancorgs, line 56: Repeated text. BTW you also copypasted the duplication to the description of the PR.

library/packages/src/modules/Package.rb
```diff
@@ -103,6 +106,32 @@ def main
       @removed_packages = []
     end
 
+    # Determines whether the package is provided or not
+    #
+    # This method checks whether any installed package provides the given "package".
+    #
+    # @param package [String] Package name
+    # @param target [Symbol,nil] :autoinst or :system. If it is nil,
```
ancorgs, line 114: I'm a bit confused by the possible values here. The description of the class it says this can work with either `PackageSystem` (libzypp) or `PackagesProposal`. So I was kind of expecting the values :system and :proposal (or something similar). But I see :autoinst instead that results in calls to `PackagesAI` (that is not exactly `PackagesProposal`).

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Mar 11 13:05:14 UTC 2022 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Extend the Package module to force using PackageSystem or
+  PackageAI without having the mode into account (bsc#1196963).
```
ancorgs, line 5: A bit technical and hard to connect with the real problems for a user checking whether this fixes a problem. Isn't it? This includes much more than extending an internal API. 

I would go for something closer to "AutoYaST clonning (export): fixed the way to check if packages are installed "

library/packages/src/modules/Package.rb
```diff
@@ -38,30 +38,32 @@
...
+  # {CheckAndInstallPackages} does not install the package for real. Instead, it adds the package to
+  # the list of packages to include in the profile. However, when running on other modes (normal,
+  # installation, etc.), it just installs the package.
+  #
+  # ## Overriding default behavior
+  #
+  # There might a scenario where you want to force Package to work with the real packages. For
+  # instance, while reading the configuration during a `clone_system` operation: the mode is still
+  # `autoinst_config` but you are dealing with the underlying system. In those cases, you can force
+  # {Package} to work with {PackageSystem}.
```
ancorgs, line 56: with {PackageAI}. Isn't it?

ancorgs (approved):
LGTM

ancorgs (approved):
LGTM again after the added note regarding YCP

PR #1253, opened 2022-04-28
Show what Product is Being Installed [SLE-15-SP4]

shundhammer:
## Trello

https://trello.com/c/CeYLzrD1/

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1196674

## Problem

Users installing many machines, especially with AutoYaST, might not be sure which product they are installing when almost everything is automated. They (or some automatism) might or might not have picked the correct AutoYaST profile, but they can't verify that.

## Fix

Show the name of the base product while packages are being installed in the new unified installation progress step, but only during system installation or upgrade (no matter if using AutoYaST or not); not in the installed system.

## Limitations

Only the _base_ product is shown, not any add-on products. This is in line with the reduction of this progress reporting to one single progress bar with only the overall progress, not individual packages.

Notice that many SLE-based products still show their own product name (e.g. "SUSE Linux Enterprise Server 15 Service Pack 4 for SAP"), not just the product they are derived from (just "SUSE Linux Enterprise Server 15 Service Pack 4") because they are genuine base products.

## Test

Manual tests:

- In an installed TW VM: Installing packages in the installed system
- Leap 15.4 installation
- SLE-15 SP4 installation

## Screenshots

Installing-product-Leap-15 4

_Leap 15.4 installation_

Change since this screenshot: No more bold font for the product name. If we want that, it should be done via QSS which is a lot more flexible.

Installing-product-SLE-HPC-15-SP4

_SLE-15 SP4 installation (Qt)_

Notice that it really displays the HPC product, not just SLE-15 SP4 which it is based on.

Installing-product-SLE-HPC-15-SP4-NCurses

_SLE-15 SP4 installation (NCurses)_

## Related PR

- PR for master / Factory: _TO DO_

### DIFF
--- a/library/packages/src/modules/SlideShow.rb
+++ b/library/packages/src/modules/SlideShow.rb
@@ -153,18 +153,21 @@ module Yast
 
     # Start the internal (global) timer.
     #
+    # FIXME: Obsolete published method
     def StartTimer
       nil
     end
 
     # Reset the internal (global) timer.
     #
+    # FIXME: Obsolete published method
     def ResetTimer
       nil
     end
 
     # Stop the internal (global) timer and account elapsed time.
     #
+    # FIXME: Obsolete published method
     def StopTimer
       nil
     end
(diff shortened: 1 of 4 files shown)


ancorgs (approved):
LGTM

PR #1255, opened 2022-05-04
Fix Unit Tests in AutoBuild

shundhammer:
## Problem

Autobuild for this package fails with unit tests:

```
Failed to load Module 'SlideShow' due to: cannot load such file -- y2packager/product_spec
```

https://ci.suse.de/job/yast-yast-yast2-SLE-15-SP4/4/console

## Cause

The .spec file of this package (yast2.spec) does not _require_ or _buildrequire_ yast2-packager, so AutoBuild does not install it in the build environment. But adding it would mean a cyclic dependency between packages _yast2_ (this package) and _yast2-packager_.

## Solution

This adds a workaround for the real-world problem, disregarding the academic problem that a system using the SlideShow does indeed not install _yast2-packager_:

This overrides Ruby `Kernel::require` with a custom version that skips loading some few predefined Ruby libraries; like the one that could not be loaded here.

## Does this Cause Problems?

Not in real life. That SlideShow.rb module that caused the error is only called by YaST clients that install packages, and those clients will all RPM-require _yast2-packager_ anyway.

## Version Bump?

No since it was `rake osc:sr` that failed here as the very last step in Jenkins. The last version never arrived in OBS. This PR is to make that possible.

### DIFF
--- a/library/packages/test/test_helper.rb
+++ b/library/packages/test/test_helper.rb
@@ -2,3 +2,29 @@ require_relative "../../../test/test_helper"
 require "pathname"
 
 PACKAGES_FIXTURES_PATH = Pathname.new(File.dirname(__FILE__)).join("data")
+
+LIBS_TO_SKIP = [
+  "y2packager/product_spec" # used in SlideShow.rb
+].freeze
+
+# Hack to avoid to require some files. Stolen from
+# https://github.com/yast/yast-storage-ng/blob/master/test/spec_helper.rb#L32-L50
+#
+# This is here to avoid a cyclic dependency with yast2-packager at build time.
+# yast2.spec does build-require yast2-packager, so the (Ruby) require for files
+# defined by that package must be avoided.
+#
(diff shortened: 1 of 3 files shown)


ancorgs (approved):
LGTM

library/packages/test/test_helper.rb
```diff
@@ -2,3 +2,29 @@
...
+#
+# Notice that the problem might be hidden for locally running the unit tests,
+# but not when calling them in an Autobuild environment (e.g. "rake osc:build"
+# or "rake osc:sr").
+module Kernel
+  alias_method :old_require, :require
+
+  def require(path)
+    old_require(path) unless LIBS_TO_SKIP.include?(path)
+  end
```
jreidinger, line 29: NP: maybe something for `yast/rspec` where we already have ability to skip loading yast module. So maybe another method to abstract mocking load of some ruby classes can be useful at multiple places.

PR #1256, opened 2022-05-06
Fix packager dep

jreidinger:
## Problem

pull request https://github.com/yast/yast-yast2/pull/1253 cause several build failures when module even transitively need SlideShow, but does not need packager to build itself.

## Solution

Load packager library only dynamic and handle if it is not available.

## Testing

- adapted unit tests to demonstrate that mocking of packager is not needed

## Screenshots

not affected

### DIFF
--- a/library/packages/src/modules/SlideShow.rb
+++ b/library/packages/src/modules/SlideShow.rb
@@ -427,7 +427,7 @@ module Yast
 
       product.display_name
     rescue LoadError
-      log.info "product spec not available. Probably packager missing"
+      log.info "product spec not available. Probably yast2-packager is missing"
 
       nil
     end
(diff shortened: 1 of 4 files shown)


library/packages/src/modules/SlideShow.rb
```diff
@@ -421,10 +420,16 @@ def product_name
...
       return nil if Mode.normal
 
+      # lazy load product spec as it lives in packager and maybe is not available
+      require "y2packager/product_spec"
       product = Y2Packager::ProductSpec.selected_base
       return nil if product.nil?
 
       product.display_name
+    rescue LoadError
+      log.info "product spec not available. Probably packager missing"
```
imobachgs, line 430: what about: "Probably yast2-packager is missing."

imobachgs (approved):

imobachgs (approved):

PR #1259, opened 2022-06-16
Package system container

jreidinger:
## Problem

When running dns-server in container it wants to install bind. But install it to container and not machine itself, so named cannot run.

## Solution

Init properly PackageSystem when root is changed.

## Testing

- *Tested manually*


lslezak (approved):
LGTM

PR #1260, opened 2022-06-24
save_y2logs - check both Packages and Packages.db (bsc#1200831)

lslezak:
## Problem

- `save_y2logs` might no save the list of installed packages when RPM stores the data in `Packages.db` file (instead of `Packages`). That depends on the RPM which is installed.

## Solution

- Check both file locations

## Testing

Asked @dgdavid for verifying the fix. :smile:


dgdavid (approved):
It works, thank you!

PR #1261, opened 2022-07-22
Fixed text mode control center in container (bsc#1199840)

lslezak:
## Details

- Always display the YaST modules installed in the system root
- When running in a container display the modules installed in the container (originally it displayed the modules from the host system which might not contain any YaST module)
- 4.5.7

## Notes

- The Qt control center is not affected, it always loads the files from the system root.
- The started YaST modules run in a new separate process so they use the chrooting as expected, they correctly manage the host system

#### The Original Behavior

menu_broken

#### With the Fix Applied

menu_fixed

### DIFF
--- a/library/desktop/src/clients/menu.rb
+++ b/library/desktop/src/clients/menu.rb
@@ -66,6 +66,13 @@ module Yast
         "Hidden"
       ]
 
+      # always load the desktop files from the system root
+      if WFM.scr_chrooted?
+        # revert back the chrooting
+        handle = Yast::WFM.SCROpen("chroot=/:scr", false)
+        Yast::WFM.SCRSetDefault(handle)
+      end
+
       Desktop.Read(@Values)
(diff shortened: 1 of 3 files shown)


library/desktop/src/clients/menu.rb
```diff
@@ -66,6 +66,13 @@ def main
         "Hidden"
       ]
 
+      # always load the desktop files from the system root
+      if WFM.scr_chrooted?
+        # revert back the chrooting
+        handle = Yast::WFM.SCROpen("chroot=/:scr", false)
+        Yast::WFM.SCRSetDefault(handle)
+      end
```
jreidinger, line 74: and will be the started modules chrooted or not?

lslezak: Yes, they will, it starts new subprocesses which have their own SCR instances.

jreidinger (approved):

PR #1262, opened 2022-07-27
Added method to find a package according to a pattern

mchf:
## Problem

The feature was not supported. Some packages do not have "default" placeholder but uses versions in name instead (php8), so simple patterns comes handy.

- https://bugzilla.suse.com/show_bug.cgi?id=1200016

## Solution

implemented simple wrapper which uses y2packager/resolvable

## Testing

- *Tested manually*

### DIFF
--- a/library/packages/src/modules/Package.rb
+++ b/library/packages/src/modules/Package.rb
@@ -224,11 +224,18 @@ module Yast
 
     # Tries to find a package according to the pattern
     #
-    # @param pattern [String] a pattern to match, no escaping done
+    # @param pattern [String] a regex pattern to match, no escaping done
     # @return list of matching package names
-    def by_pattern(pattern = "")
+    def by_pattern(pattern)
+      raise ArgumentError, "Missing search pattern" if pattern.nil? || pattern.empty?
+
       init_packager
-      Y2Packager::Resolvable.find(provides_regexp: "^#{pattern}$")
+
+      # NOTE:
+      # - Resolvable.find takes POSIX regexp, later select uses Ruby regexp
+      # - Resolvable.find supports regexps only for dependencies, so we need to
+      # filter result according to package name
+      Y2Packager::Resolvable.find({ provides_regexp: "^#{pattern}$" }, [:name])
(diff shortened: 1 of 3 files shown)


library/packages/src/modules/Package.rb
```diff
@@ -221,6 +222,18 @@ def reset
       @removed_packages.clear
     end
 
+    # Tries to find a package according to the pattern
+    #
+    # @param pattern [String] a pattern to match, no escaping done
+    # @return list of matching package names
+    def by_pattern(pattern = "")
+      init_packager
+      Y2Packager::Resolvable.find(provides_regexp: "^#{pattern}$")
```
mvidner, line 231: Maybe a comment noting that [Y2Packager::Resolvable.find](https://www.rubydoc.info/github/yast/yast-yast2/master/Y2Packager/Resolvable.find) takes POSIX EREs as opposed to Ruby Regexp below?

Also, its docs says to specify the used attrs for performance reasons
```suggestion
      Y2Packager::Resolvable.find(provides_regexp: "^#{pattern}$", [:name])
```

library/packages/src/modules/Package.rb
```diff
@@ -221,6 +222,18 @@ def reset
...
+        .select{|p| p.name =~ /\A#{pattern}\z/ }
```
mvidner, line 232: `find` has filtered the results, why do we need to do it again? Add a comment?

library/packages/src/modules/Package.rb
```diff
@@ -221,6 +222,18 @@ def reset
       @removed_packages.clear
     end
 
+    # Tries to find a package according to the pattern
+    #
+    # @param pattern [String] a pattern to match, no escaping done
+    # @return list of matching package names
+    def by_pattern(pattern = "")
```
mvidner, line 229: providing a default value for `by_pattern` IMHO only encourages unreadable call sites

library/packages/src/modules/Package.rb
mchf: thanks for the tip

mvidner (approved):

PR #1263, opened 2022-07-27
Fixed libzypp initialization (related to bsc#1199840)

lslezak:
## Problem

- YaSt when running in a container fails installing the needed packages
- The problem is that it uses the repositories configured in the container, not the repositories from the host

## Details

It turned out that the problem is in the `Pkg.GetSolverFlags` call which is called *before* initializing libzypp. In that case libzypp automatically initializes itself, but it uses by default the system root (`/`) which is the container. The libzypp initialization later done from YaST does not change anything as libzypp is already initialized.

Also it initialized the libzypp in a wrong order. The libzypp target always needs to be initialized *before* initializing the libzypp sources. The reason is that the target defines which GPG keys and config file (`zypp.conf`) are used. When initializing in wrong order you might get GPG signature errors.

## Solution

- Explicitly initialize libzypp before setting the solver flags
- Additionally I added some missing target initialization calls to ensure the target is always properly initialized.

## Testing

- Tested manually, now running `yast2-kdump` properly finds and installs the `kdump` package into the host system
- Also manually tested the `Available` and `PackageAvailable` calls in the debugger session

### DIFF
--- a/library/packages/src/modules/PackageSystem.rb
+++ b/library/packages/src/modules/PackageSystem.rb
@@ -154,8 +154,8 @@ module Yast
       Builtins.y2debug("toinstall: %1, toremove: %2", toinstall, toremove)
       return false if !PackageLock.Check
 
-      EnsureSourceInit()
       EnsureTargetInit()
+      EnsureSourceInit()
       ok = true
 
       Yast.import "Label"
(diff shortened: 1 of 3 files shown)


mvidner (changes requested):
Nice! Please copy the "Details" section of the PR description to the commit message

PR #1265, opened 2022-08-26
Transaction packages

jreidinger:
## Problem

YaST failed to install any additional package on transactional system.

## Solution

Inform user which packages are needed and let them install it.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

transactional_system

### DIFF
--- a/library/packages/src/modules/Package.rb
+++ b/library/packages/src/modules/Package.rb
@@ -460,11 +460,13 @@ module Yast
     # @return [Boolean]
     def IsTransactionalSystem
       return @transactional unless @transactional.nil?
+
       mounts = SCR.Read(path(".proc.mounts"))
       root = mounts.find { |m| m["file"] == WFM.scr_root }
       log.info "root in mounts #{root.inspect}"
 
       raise "Failed to find #{WFM.scr_root} at /proc/mounts" unless root
+
       # check if there are ro keyword in mount
       @transactional = /(?:^|,)ro(?:,|$)/.match?(root["mntops"])
     end
--- a/library/packages/src/modules/Package.rb
+++ b/library/packages/src/modules/Package.rb
@@ -34,9 +34,11 @@ require "yast"
 require "forwardable"
 require "y2packager/resolvable"
 
+Yast.import "CommandLine"
 Yast.import "Mode"
 Yast.import "PackageAI"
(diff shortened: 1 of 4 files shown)


library/packages/src/modules/Package.rb
```diff
@@ -452,30 +455,44 @@ def LastOperationCanceled
       @last_op_canceled
     end
 
-    publish function: :by_pattern, type: "list <string> (string)"
+    # Return if system is transactional and does not support direct package
+    # install
+    def TransactionalSystem
```
imobachgs, line 460: Why `TransactionalSystem` instead of `transactional_system?`? If it is just for keeping the consistency with the rest of public methods, I would call it `IsTransactionalSystem` (so it is clear it returns a boolean).

imobachgs (commented):
In general, it looks good (just a minor comment about a method name).

library/packages/src/modules/Package.rb
jreidinger: yeap, goal is consistence. So prefix with Is sounds good to me

package/yast2.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Aug 26 13:18:39 UTC 2022 - Josef Reidinger <jreidinger@suse.com>
+
+- Inform user when running on transactional system that it needs
```
imobachgs, line 4: What about "On transactional systems, inform the user that packages are required to be installed manually"?

imobachgs (approved):
Please, check the changes entry. Otherwise, LGTM.

library/packages/src/modules/Package.rb
```diff
@@ -512,6 +532,24 @@ def find_backend(target)
...
     end
+
+    # checks if working on transactional system
+    # if so, then it shows popup to user and abort yast
+    def check_transactional_system!(packages, mode = :install)
+      return unless IsTransactionalSystem()
+
+      msg = _("Transactional system detected. ")
+      case mode
+      when :install then msg += _("Following packages have to be installed manually:")
```
ancorgs, line 543: I know I'm late to the party but... https://dictionary.cambridge.org/grammar/british-grammar/following-or-the-following

Not to mention the usage of plural when there is only one package. :wink:

PR #1269, opened 2022-09-01
NetworkService enforce selected network service is enabled during installation (bsc#1202479)

teclator:
## Problem

During installation, if apparently the selected network service has not been modified (wicked is default backend) the service is not enabled explicitly in the target system which could end with it disabled. 

- https://bugzilla.suse.com/show_bug.cgi?id=1202479

## Solution

Added a parameter in order to force an enablement of the selected network service even when if has not been modified.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/library/network/src/modules/NetworkService.rb
+++ b/library/network/src/modules/NetworkService.rb
@@ -238,10 +238,11 @@ module Yast
     end
 
     # Helper to apply a change of the network service
-    def EnableDisableNow
-      return if !Modified()
+    # @param force [Boolean] whether the service should forced to be enabled or not
+    def EnableDisableNow(force: false)
+      return if !force && !Modified()
 
-      if current_name
+      if current_name && Modified()
         stop_service(current_name)
         disable_service(current_name)
       end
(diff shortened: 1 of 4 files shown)


library/network/src/modules/NetworkService.rb
```diff
@@ -238,10 +238,11 @@ def reset!
     end
 
     # Helper to apply a change of the network service
-    def EnableDisableNow
-      return if !Modified()
```
imobachgs, line 242: I do not know if it is easier to read "return unless force || Modified()"...

package/yast2.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Sep  1 08:08:30 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
+
+- Allow to force an enablement of the selected network service even
```
imobachgs, line 4: I find the description rather confusing for a user. I would try to think on a higher level.

imobachgs (changes requested):

library/network/src/modules/NetworkService.rb
mchf: I personally prefer the way how it is know ... i dislike ```unless``` ;-)

mchf (approved):
lgtm

imobachgs (approved):

PR #1270, opened 2022-09-02
NetworkService enforce selected network service is enabled during installation (bsc#1202479) (master)

teclator:
It merges #1269 into master

### DIFF
--- a/library/network/src/modules/NetworkService.rb
+++ b/library/network/src/modules/NetworkService.rb
@@ -237,8 +237,14 @@ module Yast
       @cached_name = nil
     end
 
-    # Helper to apply a change of the network service
-    # @param force [Boolean] whether the service should forced to be enabled or not
+    # Helper to apply a change of the network service enabling the service selected and taking care
+    # of disabling the current one in case of modified.
+    #
+    # TODO: this method could take care of disabling all the services except the one selected in
+    #   order to prevent that multiple backends are enabled at the same time.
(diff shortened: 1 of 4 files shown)


library/network/src/modules/NetworkService.rb
```diff
@@ -238,10 +238,11 @@ def reset!
...
 
     # Helper to apply a change of the network service
-    def EnableDisableNow
-      return if !Modified()
+    # @param force [Boolean] whether the service should forced to be enabled or not
+    def EnableDisableNow(force: false)
+      return unless force || Modified()
 
-      if current_name
+      if current_name && Modified()
```
jreidinger, line 245: so will this work in the case that we think it is not modified, but it was modified by some rpm during installation? Is this `Modified` needed?

teclator: this is about the service, for example when current network service is NM and you select wicked it will stop NM and disable it. But if it is not modified like for example in case of wicked which is the default during installation then there is nothing that should be stopped or disabled.

The current problem was that the cockpit pattern added NM and the service will be enabled by the pattern or other rpm or any other third party thingy, but that is not responsibility of NetworkService, it will just stop and disable the current service and will enable the selected one in case of forced, before the fix we basically return without enable wicked explicitly.... not sure if I clarified your question

jreidinger (approved):

PR #1271, opened 2022-09-09
Replaced egrep with grep -E

mchf:
## Problem

See https://github.com/yast/yast.github.io/pull/336 for details

### DIFF
--- a/library/general/src/modules/Arch.rb
+++ b/library/general/src/modules/Arch.rb
@@ -459,7 +459,7 @@ module Yast
     def is_zkvm
       if @_is_zkvm.nil?
         # using different check than on x86 as recommended by IBM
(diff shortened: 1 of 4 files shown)


library/general/src/modules/Arch.rb
```diff
@@ -459,7 +459,7 @@ def is_kvm
     def is_zkvm
       if @_is_zkvm.nil?
         # using different check than on x86 as recommended by IBM
-        @_is_zkvm = s390 && Yast::WFM.Execute(path(".local.bash"), "/usr/bin/egrep 'Control Program: KVM' /proc/sysinfo") == 0
+        @_is_zkvm = s390 && Yast::WFM.Execute(path(".local.bash"), "/usr/bin/grep -E 'Control Program: KVM' /proc/sysinfo") == 0
```
mvidner, line 462: NP: plain grep would work as well here

mvidner (approved):

library/general/src/modules/Arch.rb
mchf: done

PR #1272, opened 2022-09-14
Backport from master

mchf:
## Problem

leap_latest test failes in https://github.com/yast/yast-http-server/pull/58
As we work in perl we have no power to mock method calls. Also Mode->test doesn't work anymore, so backporting https://github.com/yast/yast-yast2/pull/1262 seems to be the only way how to make leap_lates tests pass

### DIFF
--- a/library/packages/src/modules/Package.rb
+++ b/library/packages/src/modules/Package.rb
@@ -224,11 +224,18 @@ module Yast
 
     # Tries to find a package according to the pattern
     #
-    # @param pattern [String] a pattern to match, no escaping done
+    # @param pattern [String] a regex pattern to match, no escaping done
     # @return list of matching package names
-    def by_pattern(pattern = "")
+    def by_pattern(pattern)
+      raise ArgumentError, "Missing search pattern" if pattern.nil? || pattern.empty?
+
       init_packager
-      Y2Packager::Resolvable.find(provides_regexp: "^#{pattern}$")
+
+      # NOTE:
+      # - Resolvable.find takes POSIX regexp, later select uses Ruby regexp
+      # - Resolvable.find supports regexps only for dependencies, so we need to
+      # filter result according to package name
+      Y2Packager::Resolvable.find({ provides_regexp: "^#{pattern}$" }, [:name])
         .select{|p| p.name =~ /\A#{pattern}\z/ }
         .map(&:name)
         .uniq
--- a/library/packages/src/modules/Package.rb
+++ b/library/packages/src/modules/Package.rb
@@ -231,8 +231,7 @@ module Yast
 
       init_packager
 
-      # NOTE:
-      # - Resolvable.find takes POSIX regexp, later select uses Ruby regexp
+      # NOTE: - Resolvable.find takes POSIX regexp, later select uses Ruby regexp
       # - Resolvable.find supports regexps only for dependencies, so we need to
       # filter result according to package name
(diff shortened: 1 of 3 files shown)


library/packages/src/modules/Package.rb
```diff
@@ -221,6 +222,24 @@ def reset
...
     end
 
+    # Tries to find a package according to the pattern
+    #
+    # @param pattern [String] a regex pattern to match, no escaping done
+    # @return list of matching package names
+    def by_pattern(pattern)
+      raise ArgumentError, "Missing search pattern" if pattern.nil? || pattern.empty?
+
+      init_packager
```
jreidinger, line 232: is it safe to call it always?

lslezak: It should be responsibility of the caller to ensure that the package manager is properly initialized. It might have potentially some side effects during offline upgrade or some other less used scenarios.

The other methods here also do not run initialization so please remove it.

And for initialization you should use [PackageSystem.EnsureTargetInit](https://github.com/yast/yast-yast2/blob/1ce6c5d2f2188f757b73550e738a188bdd8c1d1c/library/packages/src/modules/PackageSystem.rb#L74) and [PackageSystem.EnsureSourceInit](https://github.com/yast/yast-yast2/blob/1ce6c5d2f2188f757b73550e738a188bdd8c1d1c/library/packages/src/modules/PackageSystem.rb#L92) calls in the caller code. These calls avoid initializing the already initialized package manager.

library/packages/src/modules/Package.rb
```diff
@@ -221,6 +222,22 @@ def reset
...
+    #
+    # @param pattern [String] a regex pattern to match, no escaping done
+    # @return list of matching package names
+    def by_pattern(pattern)
+      raise ArgumentError, "Missing search pattern" if pattern.nil? || pattern.empty?
+
+      # NOTE: - Resolvable.find takes POSIX regexp, later select uses Ruby regexp
+      # - Resolvable.find supports regexps only for dependencies, so we need to
+      # filter result according to package name
+      Y2Packager::Resolvable.find({ provides_regexp: "^#{pattern}$" }, [:name])
```
lslezak, line 235: Two problems here:

- It does not specify the kind of the resolvable, this might return any matching objects including patterns or products (i.e. not only packages).
- The `:name` attribute is always loaded, you do not need to specify that (minor issue) 

```suggestion
      Y2Packager::Resolvable.find({ kind: :package, provides_regexp: "^#{pattern}$" })
```


However, the function seems to be very specific, it searches Provides **and** the name using the same regexp. This looks like a very special case which very likely won't be used by anyone else. I'm not sure if it makes sense to share it here, I'd move it to the code which needs this.

Finally, the method name looks a bit confusing, "pattern" could be easily confused with libzypp patterns. A better name would be `by_provides_regexp` or something like that...

library/packages/src/modules/Package.rb
```diff
@@ -221,6 +222,22 @@ def reset
...
+        .select { |p| p.name =~ /\A#{pattern}\z/ }
```
lslezak, line 236: I'm not sure if using the same regexp as POSIX regexp and then as a Ruby regexp is a good idea. In many cases this will work, but there are still some corner cases (like that `\A`) where it would break. In some specific case it might be OK, but I'd avoid this POSIX regexp == Ruby regexp assumption in a generic code.

lslezak (commented):
In the end, maybe we should actually move the code to yast2-http-server (or whatever needs this)... :thinking:

PR #1276, opened 2022-10-05
remove #postprocess_url from RelURL class

wfeldt:
## Problem

- https://trello.com/c/CAytlL3W
- https://jira.suse.com/browse/SLE-22578
- https://jira.suse.com/browse/SLE-24584

`Yast2::RelURL` does a nice handling of relative URLs. But `#postprocess_url` messes up the results and always returns URLs starting with `//` (exactly two slashes). This is wrong.

Ruby's URI class works just fine. Just leave it at that. If YaST really relies on something like `file://foo/bar` this should be fixed in YaST (but I don't think this is the case).

## See also

- https://github.com/openSUSE/linuxrc/pull/299
- https://github.com/yast/yast-installation/pull/1063
- https://github.com/yast/yast-autoinstallation/pull/852

### DIFF
--- a/library/general/src/lib/yast2/rel_url.rb
+++ b/library/general/src/lib/yast2/rel_url.rb
@@ -104,7 +104,6 @@ module Yast2
         ret.path = absolute_path
       end
 
-      postprocess_url(ret)
       ret
     end
 
@@ -118,18 +117,6 @@ module Yast2
       # URI requires absolute path
       url.path = File.join("/", url.host, url.path)
       url.host = nil
-    end
-
-    # a reverse method to "preprocess_url", fix the URL if it is a "file://" URL
-    def postprocess_url(url)
-      return if url.scheme != "file" && !url.host.nil? && !url.host.empty?
-
-      path = url.path.sub(/\A\/+/, "").split("/")
-      url.host = path.shift
-
-      rest = File.join(path)
-      rest.prepend("/") unless rest.empty?
-      url.path = rest
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lslezak:
The problem with Ruby URL is that it does not handle the `file://` scheme well. It still considers the first path element as a host name:

```
# ruby -r uri -e "u=URI('file://foo/bar'); puts u.host; puts u.path;u.host='';puts u"
foo
/bar
file:///bar
```

IIRC the pre/post processing was added to workaround these problems.

wfeldt:
Yes, but this is correct; `file://foo/bar` (with 2 slashes) is unusual. Preprocessing is ok to make file with 2 slashes valid as we document it in some places. But essentially there should be one or three slashes. And I think `file:///foo/bar` does work in YaST already.

lslezak:
OK, I'll do some more tests...

wfeldt:
I have looked at it again and IMHO `preprocess_url` does the right thing - that is, fixing the hostname part (i.e. converting 2-slash variants to the 1/3-slash variant).

But I'm not convinced there should be a `postprocess_url` that undoes it. Not even for `file` and `relurl`. `file` should never have a 'host' component and IMHO also `relurl` should not. I'm aware that `relurl` is documented to use 2 slashes but iirc we internally 'fix' that aleady and allow also the 1/3 slash variants.

FWIW, the existing logic in `postprocess_url` is not correct and turns every url into a 2-slash url (the initial `return if` condition is wrong). I was tempted to 'fix' `postprocess_url` to return the exact number of slashes as the original base url uses but couldn't really convince myself that this is a sane thing. The whole url processing is a bit too weird as it is already.

lslezak (approved):
LGTM, thanks!

PR #1279, opened 2022-10-31
Fix hash vs keyword arguments in RSpec expectations (bsc#1204871)

mvidner:
## Problem

- https://bugzilla.opensuse.org/show_bug.cgi?id=1204871 "- [Staging] rubygem-rspec-* 3.12 breaks various yast modules build"
- this description is copied from https://github.com/yast/yast-registration/pull/584 as this is a base library. Further PRs will just link here

New rspec-mocks distinguish between hash and keyword arguments passed to `with`.

The tested code does use hashes but we've used keywords in tests because
it's shorter to write. Fixed.

rspec-mocks 3.11.2 / 2022-10-25:

> Bug Fixes: Support keyword argument semantics when constraining argument
> expectations using with on Ruby 3.0+ with instance_double
> https://github.com/rspec/rspec-mocks/pull/1473

## Solution

Use hashes in tests instead of keywords

## Testing

- [x] Red-green tested with manually installed rspec 3.12.0

## Screenshots

Not affected

### DIFF
--- a/library/cwm/test/cwm_test.rb
+++ b/library/cwm/test/cwm_test.rb
@@ -48,6 +48,7 @@ describe Yast::CWMClass do
 
   let(:test_stringterm) { VBox(HBox("w1")) }
   let(:widget_names) { ["w1", "w2"] }
+  let(:event) { { "ID" => :my_event_id } }
   let(:test_widgets) do
     {
       "w1" => {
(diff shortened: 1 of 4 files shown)


library/cwm/test/cwm_test.rb
```diff
@@ -158,65 +159,65 @@ def generic_save(_key, _event)
...
-      expect(subject.validateWidgets(run_widgets, "ID" => :event)).to eq(true)
+      expect(self).to receive(:w1_validate).with("w1", event).and_return(true)
+      expect(self).to receive(:w2_validate).with("w2", event).and_return(true)
+      expect(subject.validateWidgets(run_widgets, event)).to eq(true)
     end
 
     context "if a handler returns false" do
       before do
-        expect(self).to receive(:w1_validate).with("w1", "ID" => :event).and_return(false)
+        expect(self).to receive(:w1_validate).with("w1", event).and_return(false)
```
mvidner, line 215: BTW the last 3 changes in this file don't have a `with` call so they are just for consistency

joseivanlopez (approved):
LGTM

PR #1280, opened 2022-11-30
add arch filter class to unify architectures specification

jreidinger:
## Problem

There are multiple places where we need to specify in text form condition to apply only to certain hardware architectures and it is not unified so far.

Related to https://trello.com/c/C3apZtHv/3231-d-installer-support-for-dynamic-configuration

## Solution

Unified solution is based on specification of  subvolume handling https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/subvol_specification.rb#L119

There are some improvements to it:

- allowing white spaces
- case insensitivity
- support for 'all' method
- raise exception when invalid element of string  is used

## Testing

- new unit tests are added

## Open Questions

There is one more place which define architecture filtering - in workflows https://github.com/yast/yast-yast2/blob/master/library/control/src/modules/WorkflowManager.rb#L209 Question is if we want to use filter also here? If so, then we need to support `all` keyword to match everything and of course adapt that code.

### DIFF
--- a/library/general/src/lib/yast2/arch_filter.rb
+++ b/library/general/src/lib/yast2/arch_filter.rb
@@ -36,7 +36,8 @@ module Yast2
 
     # @return [Array<Hash>] list of specifications where each entry is hash with key `:method` and
     #   `:negate`, where method is Yast::Arch method and negate specify if
-    #   method have to return false
+    #   method have to return false. There is one specific method `:all` that is not in Yast::Arch,
+    #   but can be used to always return true.
     attr_reader :specifications
 
     # creates new architecture filter from passed list of individual specifications
@@ -48,7 +49,7 @@ module Yast2
         method = spec.downcase
         negate = spec.start_with?("!")
         method = spec[1..-1] if negate
-        raise Invalid, spec unless Yast::Arch.respond_to?(method)
+        raise Invalid, spec unless valid_method?(method)
 
         @specifications << { method: method.to_sym, negate: negate }
       end
@@ -67,6 +68,7 @@ module Yast2
     #   "ppc, !board_powernv" # spaces are allowed
     #   "!ppc64,!aarch64" # always returns false as there is none positive method
     #   "s390, !is_zKVM" # return true on s390 when not running in zKVM hypervisor
+    #   "all,!s390" # return true on all archs except s390
     #   "invalid" # raises ArchFilter::Invalid exception
     def self.from_string(value)
       new(value.split(",").map(&:strip))
@@ -76,9 +78,23 @@ module Yast2
     # @return [Boolean]
     def match?
       negative, positive = @specifications.partition { |s| s[:negate] }
-      return false if negative.any? { |s| Yast::Arch.public_send(s[:method]) }
+      return false if negative.any? { |s| invoke_method(s[:method]) }
 
-      positive.any? { |s| Yast::Arch.public_send(s[:method]) }
+      positive.any? { |s| invoke_method(s[:method]) }
+    end
+
+  private
+
+    def invoke_method(m)
+      return true if m == :all
+
+      Yast::Arch.public_send(m)
+    end
+
+    def valid_method?(m)
+      return true if m.to_s == "all"
+
+      Yast::Arch.respond_to?(m)
     end
   end
 end
--- a/library/general/src/lib/yast2/arch_filter.rb
+++ b/library/general/src/lib/yast2/arch_filter.rb
@@ -60,15 +60,17 @@ module Yast2
     # When "!" is used it is called negative method and without it is called positive.
     # List of possible values are supported with comma separator. In list
     # at least one positive specified method have to return true and all methods
-    # with `!` have to be false to return true as result. Whitespaces are allowed. Only `!` and method
+    # with `!` have to be false to return true as result. If there are no positive methods
+    # then only negatives are evaluated. Whitespaces are allowed. Only `!` and method
     # has to be without space. Also it is case insensitive, so acronyms can be in upper case.
     # @example various format and how it behave in given situations
     #   "x86_64,ppc64" # returns true on either x86_64 or ppc64
     #   "ppc,!board_powernv" # returns false on powernv_board or non-ppc
     #   "ppc, !board_powernv" # spaces are allowed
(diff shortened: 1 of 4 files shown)


ancorgs:
It generally looks good. Just two notes:

- Documentation is hard to read. Fortunately examples help. Maybe we could invoke @mvidner here for some help on generating understandable (English) texts.
- I agree having some way to specify all architectures would make sense. Not only to support the workflow manager but also for cases in which you only want to exclude a particular case. As documented ""!ppc64,!aarch64" always returns false so there is no way to say "always except for ppc64 and aarch64". That would be possible with something like this "*,!ppc64,!aarch64" or "any,!ppc64,!aarch64".

mvidner:
The main question is: WHO will write the architecture specifications/filters, and where.

I'm asking because currently we delegate the elementary matchers to the implementation of `Yast::Arch`. That's fine if `ArchFilter` is internal. But as soon as we tell release managers or even customers to use it, we will have to list the individual recognized architectures authoritatively, and preferably in one place.

jreidinger:
@mvidner well, I expect it will be as usual us who will write it, but it is also used in control.xml and also d-installer.yml can be modified by users for having their own modified installer. But for list it is always hard to keep it in sync :( Even now in control.xml documentation we do not write all possible values - https://github.com/yast/yast-installation/blob/master/doc/control-file.md#the-volumes-subsection

library/general/src/lib/yast2/arch_filter.rb
```diff
@@ -0,0 +1,100 @@
...
+    # with `!` have to be false to return true as result. Whitespaces are allowed. Only `!` and method
+    # has to be without space. Also it is case insensitive, so acronyms can be in upper case.
+    # @example various format and how it behave in given situations
+    #   "x86_64,ppc64" # returns true on either x86_64 or ppc64
+    #   "ppc,!board_powernv" # returns false on powernv_board or non-ppc
+    #   "ppc, !board_powernv" # spaces are allowed
+    #   "!ppc64,!aarch64" # always returns false as there is none positive method
+    #   "s390, !is_zKVM" # return true on s390 when not running in zKVM hypervisor
+    #   "all,!s390" # return true on all archs except s390
+    #   "invalid" # raises ArchFilter::Invalid exception
```
joseivanlopez, line 72: I am not sure about `all`. I would expect something like:

* `!ppc`: true if arch != ppc
* `!ppc,!x86`: true if arch != ppc and arch != x86

jreidinger: Well, it was how I think it works originally. But found one use case where it won't work and also original code does not work in that way - https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/subvol_specification.rb#L132 ( hint: it use match even when negate is true ).

The case where it won't work is `ppc,!board_powernv` because with that change to `!` it means ppc or non powernv...so true also for x86_64 or s390.

jreidinger: and btw you write test case that verify this behavior at https://github.com/yast/yast-storage-ng/blob/master/test/y2storage/subvol_specification_test.rb#L132

joseivanlopez: umm, what I would do is to always use AND for negations and OR for possitives (as the code does now): possitives.any? AND negatives.none?. So, `ppc,!board_powernv` means `ppc` AND `!board_powernv`.

My only complaint is about using only negations. In that case, possitives.any? should eval true, so `!ppc,!x86` means all except `x86` and `ppc`.

mvidner:
> Even now in control.xml documentation we do not write all possible values - https://github.com/yast/yast-installation/blob/master/doc/control-file.md#the-volumes-subsection

OK. Now that I've read that file, I'm actually worried about the meaning of the comma `,` operator:

> Normally, architectures are combined with logical OR, i.e.
>
>     <archs>i386,x86_64</archs>
> means "if architecture i386 or x86_64". If the current architecture is an architecture that was excluded with "!", that subvolume is not used no matter what other architectures are specified that might also apply.

mvidner:
OK, let's borrow naming from logic:
- an **atom** is a call to some boolean method in Arch, (documenting the supported list and semantics is WIP)
- a **literal** is an atom or its negation via `!`
- a **conjunction** (logical **and**) is a list of literals joined with `&`
- a **disjunction** (logical **or**) is a list of literals joined with `|`
- a **guessjunction** :rofl: is a list of literals joined with `,`: if it contains a negation `!` it is a conjunction, otherwise a disjunction

IMHO this captures the existing usages of `ppc,!board_powernv` (where, importantly, `board_powernv` is a subset of `ppc`) and `i386,x86_64`

TODO: does this work with `all`?

library/general/src/lib/yast2/arch_filter.rb
```diff
@@ -0,0 +1,105 @@
...
+      def initialize(spec)
+        super("Invalid part of architecture specification '#{spec}'")
+      end
+    end
+
+    # @return [Array<Hash>] list of specifications where each entry is hash with key `:method` and
+    #   `:negate`, where method is Yast::Arch method and negate specify if
+    #   method have to return false. There is one specific method `:all` that is not in Yast::Arch,
+    #   but can be used to always return true.
+    attr_reader :specifications
```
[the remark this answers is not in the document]
jreidinger: well, I use it mainly for testing purpose and kind of make sense for me to have it here, but I am not strongly against moving it to private.

library/general/src/lib/yast2/arch_filter.rb
```diff
@@ -0,0 +1,105 @@
...
+    #   "!ppc64,!aarch64" # returns false on ppc64 and arm and true for others
+    #   "s390, !is_zKVM" # return true on s390 when not running in zKVM hypervisor
+    #   "" # always return true
+    #   "all" # always return true
+    #   "invalid" # raises ArchFilter::Invalid exception
+    def self.from_string(value)
+      new(value.split(",").map(&:strip))
+    end
+
+    # checks if filter match current hardware
```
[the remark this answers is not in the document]
jreidinger: yeap, I even consider allowing optional argument here to overwrite some of current values, but failed to find usecase beside unit testing.

mvidner (approved):

PR #1282, opened 2023-01-04
RSpec keywords vs hash, second round

mvidner:
## Problem

- https://trello.com/c/jRe5bWVE
- https://bugzilla.suse.com/show_bug.cgi?id=1206419

## Solution

- ~~fix more expectations to correctly use a hash where the method really accepts a hash~~ No longer needed after the upstream fix
- it is an upstream bug, look at https://github.com/rspec/rspec-mocks/pull/1514
    - and here's my OBS submission to d:l:r:e https://build.opensuse.org/request/show/1057124

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

*If the fix affects the UI attach some screenshots here.*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,9 @@
+/.bundle/
+/vendor/
+/Gemfile
+/Gemfile.lock
+/.ruby-version*
+
 *.pot
 *.log
 *.trs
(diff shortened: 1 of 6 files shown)


mvidner:
With rspec-mocks-3.12.2, this PR is no longer necessary.
Submitting that to OBS: 
https://build.opensuse.org/request/show/1057124

PR #1283, opened 2023-01-10
Fix help button in YaST Control Center

joseivanlopez:
## Problem

With ncurses UI, the YaST Control Center shows a help button which opens an empty popup (help texts provided by the client are missing).

Note: this problem only affects to the ncurses UI because the help button is not offered in the Qt UI.

* https://bugzilla.suse.com/show_bug.cgi?id=1206919

## Solution

The help button's option `:helpButton` was used in a wrong way: if it is used, the click/activate event of that button is handled purely internal in libyui. In that case, libyui traverses up the widget hierarchy to find the next widget that has a property `:helpText` set and uses that text for a (UI internal) help window's content.

In this case, the button has that option, but the application wants to handle the event itself (there is no widget with`:helpText` property in the hierarchy).

Removing the `:helpButton` option on the help button fixes the problem.

## Testing

* Tested manually

## Screenshots

Screenshot from 2023-01-10 11-58-00

### DIFF
--- a/library/desktop/src/clients/menu.rb
+++ b/library/desktop/src/clients/menu.rb
@@ -264,7 +264,7 @@ module Yast
           VSpacing(1.0),
           HBox(
             HSpacing(1),
-            PushButton(Id(:help), Opt(:key_F1, :helpButton), Label.HelpButton),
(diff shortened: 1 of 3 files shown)


shundhammer (approved):
LGTM

PR #1285, opened 2023-01-26
Allow dir:///foo to equal dir:/foo (bsc#1207239)

mvidner:
## Problem

Since Ruby 3.2, URI("dir:///foo").to_s no longer returns "dir:/foo" and it makes tests fail in yast2-packager:

If we route the 3-slash form via C++ zypp::Url, it gets normalized to a single slash, see also
https://github.com/openSUSE/libzypp/issues/441

- https://trello.com/c/jRe5bWVE
- https://bugzilla.suse.com/show_bug.cgi?id=1207239

## Solution

In the class that wraps zypp::Url, consider those equal

It fixes a comparison in PkgFinishClient#added_by_installer?

## Testing

- Added a new unit test

## Screenshots

N/A

### DIFF
--- a/library/packages/src/lib/y2packager/zypp_url.rb
+++ b/library/packages/src/lib/y2packager/zypp_url.rb
@@ -136,7 +136,7 @@ module Y2Packager
     # @return [String]
     def inspect
       # Prevent SimpleDelegator from forwarding this to the wrapped URI object
-      "#<#{self.class}:#{object_id}} @uri=#{uri.inspect}>"
+      "#<#{self.class}:#{object_id} @uri=#{uri.inspect}>"
     end
 
     # Compares two URLs
@@ -146,7 +146,10 @@ module Y2Packager
     # object is introduced to replace an existing URI one.
     def ==(other)
       if other.is_a?(URI::Generic)
-        uri == other
+        uri == other ||
+          # zypp will normalize dir:///foo to dir:/foo
+          # https://github.com/openSUSE/libzypp/issues/441
+          uri == URI(other.to_s.sub(":///", ":/"))
       elsif other.class == self.class
         uri == other.uri
       else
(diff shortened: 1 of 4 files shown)


library/packages/src/lib/y2packager/zypp_url.rb
```diff
@@ -136,7 +136,7 @@ def expanded
     # @return [String]
     def inspect
       # Prevent SimpleDelegator from forwarding this to the wrapped URI object
-      "#<#{self.class}:#{object_id}} @uri=#{uri.inspect}>"
+      "#<#{self.class}:#{object_id} @uri=#{uri.inspect}>"
```
jreidinger, line 139: good catch :+1:

library/packages/src/lib/y2packager/zypp_url.rb
```diff
@@ -146,7 +146,10 @@ def inspect
     # object is introduced to replace an existing URI one.
     def ==(other)
       if other.is_a?(URI::Generic)
-        uri == other
+        uri == other ||
+          # zypp will normalize dir:///foo to dir:/foo
+          # https://github.com/openSUSE/libzypp/issues/441
+          uri == URI(other.to_s.sub(":///", ":/"))
```
jreidinger, line 152: can we use it in generaal ofr all URI types? or only for dir?

mvidner: Short answer: yes

`URI::Generic` is a base class for all the Ruby URIs. The `:///` means there is an empty *authority* part, where *authority* is [user[:password]@]host[:port], and `:/` means the authority is undefined. Ruby 3.2 normalizes it to 3 slashes, zypp normalizes it to 1 slash.

But I may be misunderstanding your question. How can I improve the code or the comment?

jreidinger: it zypp do it for *all* triple slash, then it is fine for me. I just want to be sure.

PR #1294, opened 2023-10-25
Fix zones reader (bsc#1216534)

teclator:
## Problem

A recent change in firewall-cmd --list-all-zones output is breaking the zones reader as the default zone is not read properly and therefore it is considered to be deleted.

- https://bugzilla.suse.com/show_bug.cgi?id=1216534

## Solution

Adapted the firewalld zones reader ignoring anything after the zone name.

## Testing

- *Adapted current unit test*
- *Tested manually*


shundhammer (approved):
LGTM

PR #1299, opened 2024-01-23
Adapt kernel

jreidinger:
## Problem

In agama there are no /etc/install.inf, but kernel parsing depends on it. Result is that kernel parameters are not properly passed to installed system.

## Solution

Read /proc/cmdline when /etc/install.inf is not available.

## Testing

- *Tested manually*

### DIFF
--- a/library/system/src/modules/Kernel.rb
+++ b/library/system/src/modules/Kernel.rb
@@ -169,13 +169,13 @@ module Yast
       @cmdline_parsed = true
       return if !(Stage.initial || Stage.cont)
 
-      # live installation does not create /etc/install.inf (bsc#793065)
-      tmp = if Mode.live_installation
+      # Check if /etc/install.inf exists
+      tmp = if SCR.Dir(path(".etc.install_inf")).empty?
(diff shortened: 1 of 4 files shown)


library/system/src/modules/Kernel.rb
```diff
@@ -169,13 +169,14 @@ def ParseInstallationKernelCmdline
       @cmdline_parsed = true
       return if !(Stage.initial || Stage.cont)
 
-      # live installation does not create /etc/install.inf (bsc#793065)
-      tmp = if Mode.live_installation
+      # Check if /etc/install.inf exists
+      tmp = if SCR.Dir(path(".etc.install_inf")).empty?
```
teclator, line 173: I guess /etc/install.inf should be already copied in the target system when this check is done as well as the read is performed but we have to be careful about it.

teclator (approved):
LGTM

library/system/src/modules/Kernel.rb
jreidinger: well, I keep it working as before...so previously it calls SCR.Execute..so I use also SCR for Dir to check existence.

Issue #978, opened 2019-11-14
save_y2logs stores log as rw only for root


coogor:
This has the lovely consequence, once you want to attach a saved log file to bugzilla you get an 'internal server error' , as no other user can read the log file! (and of course, you are not completely working as root, so this is a daily issue)
Maybe a chmod 666 saved_file.tat.xz as last step would fix this

jreidinger:
@coogor sadly I worry this won't be accepted by security guys as it contain many info that can be abused by potential attacker, like configuration of services, firewall configuration, etc. We filter out passwords, but there are more sensitive informations like content of journal to which common user does not have access and many others.

coogor:
I see....but maybe issue a sentence that the file is rw only for root, and you may not be able to upload it if working under a different user?

dgdavid:
Hi @coogor!

Sorry for the delay, too much stuff in the TODO queue :)

Actually, that is already mentioned in https://en.opensuse.org/openSUSE:Report_a_YaST_bug#I_attached_.2Fvar.2Flog.2FYaST2.2Fy2log_to_a_YaST2_bug.2C_and_still_I_am_asked_to_attach_y2logs._Why.3F

I'm not sure if it worth adding such information in the command output. Let's wait for more feedback and see what we finally decide.

Thanks.

mvidner:
Well, we do have a contradiction:
1) we ask the user to upload the logs, potentially to a publicly visible bug
2) we explicitly make the file unreadable for non-root (for "security" reasons, see [bsc#673990](https://bugzilla.suse.com/show_bug.cgi?id=673990) for the introduction of this feature 9 years ago)

wfeldt:
I believe the link you had in mind is

https://en.opensuse.org/openSUSE:Report_a_YaST_bug#Firefox_fails_to_attach_the_.2Ftmp.2Fy2logs-.2A.tar.gz_that_I_saved_with_save_y2logs._Why.3F

We simply cannot use more liberal permissions. That you get an 'internal server error' would indicate more an issue with the web page.

A short notice about the need to adjust the permissions when you are going to upload the file might be a solution. I would nobody expect to find that cited paragraph.

mvidner:
FWIW, I find it strange to get a 500 Server error when trying to upload an unreadable file. I would expect a friendlier error from Firefox.
![bugzilla500](https://user-images.githubusercontent.com/102056/73744881-067d8080-4752-11ea-82ee-276ace3625fe.png)

[...]

Issue #970, opened 2019-10-10
save_y2logs should save journalctl, not only journalctl --dmesg


ggardet:
`save_y2logs` should save `journalctl -b`, not only `journalctl --dmesg`:
https://github.com/yast/yast-yast2/blob/master/scripts/save_y2logs#L83

It would help when systemd service fails (e.g. rmt)

ancorgs:
I created a corresponding card to track this in the private trello board used by the YaST Team at SUSE to coordinate and prioritize: https://trello.com/c/uqSvbtGh/4599-improvements-in-savey2logs

PR #1304, opened 2024-02-21
Sle 15 sp6

mchf:
## Problem

https://github.com/yast/yast-yast2/issues/978 - Permissions/ownership of the resulting file. Leaded to this [yast-devel thread](https://lists.opensuse.org/yast-devel/2020-02/msg00005.html)

https://github.com/yast/yast-yast2/issues/970 Missing journalctl -b.

## Solution

added journalctl -b instead of journalctl --dmesg and a note about resulting archive permissions

## Testing

- *Tested manually*


shundhammer (approved):
LGTM

PR #1309, opened 2024-05-06
Properly close popups for nested progress callbacks (bsc#1223281)

lslezak:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1223281
- After deleting the old repositories a progress popup stays open and blocks the following UI calls

no_widget_error

## Details

- The problem happens only when the self-update is enabled, with disabled self-update it works fine
-  It turned out that with updated libzypp the start and end callbacks are called twice:
  ```
  modules/PackageCallbacks.rb(ProgressStart):2338 ProgressStart: 992
  modules/PackageCallbacks.rb(ProgressStart):2338 ProgressStart: 992
  modules/PackageCallbacks.rb(ProgressEnd):2391 ProgressFinish: 992
  modules/PackageCallbacks.rb(ProgressEnd):2391 ProgressFinish: 992
  ```
- The code is actually prepared for this situation, it counts the nesting level for the start and end calls
- But unfortunately there was a bug when closing the nested callback, it closed only one (the last) popup

## Solution

- Properly close the open popup when there are nested callbacks (multiple open popups)

## Testing

- Tested manually
- All popups are closed, the workflow properly continues to the next step

### DIFF
--- a/library/packages/src/modules/PackageCallbacks.rb
+++ b/library/packages/src/modules/PackageCallbacks.rb
@@ -2397,7 +2397,7 @@ module Yast
       )
 
       if !Mode.commandline && IsProgressPopup()
-        UI.CloseDialog if Builtins.size(@progress_stack) == 0
+        UI.CloseDialog if Builtins.size(@progress_stack) >= 0
       elsif full_screen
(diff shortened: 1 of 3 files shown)


joseivanlopez (approved):
LGTM

PR #1310, opened 2024-05-07
[SLE15-SP6] Properly close popups for nested progress callbacks (bsc#1223281)

lslezak:
- Merging #1309 from SP5 to SP6


joseivanlopez (approved):
LGTM

PR #1311, opened 2024-05-07
[master] Properly close popups for nested progress callbacks (bsc#1223281)

lslezak:
- Merging #1310 from SP6 to master


lslezak:
@mchf is it OK to backport of the `journalctl` fix from SP6 to `master`?

mchf (approved):
I believe, it is ok to backport it to the master. The fix was requested way before SP6 was with us anyway.

Thanks

PR #1316, opened 2025-02-20
Respect Agama kernel parameters

jreidinger:
## Problem

Agama do not use linuxrc for kernel parameters filtering. Result is that even installation parameters are proposed to target system.

bsc: 
- [1234678](https://bugzilla.suse.com/show_bug.cgi?id=1234678)
- https://bugzilla.suse.com/show_bug.cgi?id=1237390

## Solution

Read kernel parameters from agama file if it is available.

## Testing

- *Tested manually*


joseivanlopez (approved):
LGTM

PR #1318, opened 2025-05-12
WIP: Drop YaST Command Line Interface Support

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1243018

## Problem / Motivation

YaST is only used as a library for the Agama backend in SLES-16 / Leap 16.

The functionality of YaST goes much further, but  we cannot guarantee that all the details and all the behavior in YaST are compatible in every detail with SLES-16 / Leap 16. This will create pitfalls for users who may find a way to still use YaST functionality that will then increasingly be outdated.

In long run, this will be true for Factory / Tumbleweed as well.

So we are starting to prevent pitfalls with this: The YaST command line  interface (CLI) should not be accessible anymore. This is even more important because this is an aspect of YaST that is less commonly used and thus less thoroughly tested.

## Fix

This PR makes the YaST  CLI unaccessible: When trying to use it, it will simply exit with an error message in the log and on stderr.

```console
sh@balrog:~> sudo yast2 lan list
FATAL: The YaST command line interface is not supported anymore. Exiting.
sh@balrog:~>
```

```console
sh@balrog:~> sudo yast2 lan
```
(-> The YaST lan module starts normally)

### DIFF
--- a/library/commandline/src/modules/CommandLine.rb
+++ b/library/commandline/src/modules/CommandLine.rb
@@ -1209,6 +1209,7 @@ module Yast
         return true
       else
         Mode.SetUI("commandline")
+        fatal_no_more_cli_support # This does not return
       end
 
       if !cmdline_supported
@@ -1722,6 +1723,15 @@ module Yast
       list[0..-2].map { |l| "'#{l}'" }.join(", ") + " " +
         Builtins.sformat(_("or '%1'"), list[-1])
     end
+  end
+
+  # Write a message that yast-cli is not supported anymore to stderr and
+  # to the log and exit with an error code.
+  def fatal_no_more_cli_support
+    msg = "FATAL: The YaST command line interface is not supported anymore. Exiting."
+    log.error(msg)
+    warn(msg)
+    exit(1)
   end
 
   CommandLine = CommandLineClass.new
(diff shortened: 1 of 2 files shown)


mvidner:
Ruby docs about the various exiting methods: https://ruby-doc.org/3.4.1/Kernel.html#module-Kernel-label-Exiting

mvidner:
@shundhammer IMHO it is acceptable to replace `exit` with an unhandled exception:
- it is well known how to work with them in RSpec
- the user behavior is also well defined in YaST
- it may look a bit ugly (backtrace...) but we can polish it later
- the main thing is that we declare early that this feature is dropped

Use a regular `RuntimeError`.

shundhammer:
Hm.. I tried with this:

```Ruby
  # Write a message that yast-cli is not supported anymore to stderr and
  # to the log and exit with an error code.
  def fatal_no_more_cli_support
    msg = "FATAL: The YaST command line interface is not supported anymore. Exiting.".freeze
    log.error(msg)
    raise RuntimeError, msg
  end
```

And now of course it opens a _window_ with an internal error and offers me to start the _byebug_ debugger; which is clearly not a good thing when you are doing command line operations, and suddenly it blocks with a window.

Is there any way to get a simple low-level Ruby exception instead?

shundhammer:
Superseded by https://github.com/yast/yast-ruby-bindings/pull/295

PR #1330, opened 2026-05-12
Drop TPM check

joseivanlopez:
## Problem

The check for TPM2 is misplaced in yast2.

## Solution

Drop logic for checking TPM2 availability. This check is now implemented by yast-storage-ng, see https://github.com/yast/yast-storage-ng/pull/1424.


ancorgs (approved):
LGTM... after merging and increasing the version of yasts-storage-ng, of course.

If we merge now with the version increment only in this package we'll break things.

